PR #5, opened 2017-03-29
Stitching!

astblake:
**Thursday 30th March**

Have fixed the changes I made to MultiPandoraApi - things are ready again!

**Wednesday 29th March (middle of night!)**

Update: Oops, I'm afraid to say that the code isn't ready just yet! While testing on MicroBooNE, I realised I'd broken the two-pass reconstruction. Both passes need to access the MultiPandoraApi simultaneously without interfering with each other - I've broken that part!

**Wednesday 29th March**

Here's my first attempt at a pull request! There are two strands of work:

1. Lots of new code in the LArStitching directory! Cosmic-ray tracks that point at each other across a drift volume boundary are first shifted in X and then stitched together.

2. Modification of track-building algorithm so that it will run inside the stitching instance. Essentially, the calculation of dQ/dL has been migrated from LArContent to LArSoft (I had to do this, otherwise the stitching instance would have to know about all the 2D transformations - ugh).

3. Hopefully I've fixed more tab characters than I created!

See what you think. I've made an associated pull request for LArReco.


astblake:
Test - can I add a new comment?

astblake:
Yes!

PandoraPFA:
Hi Andy,

This is a lot of code to review in one go, so there are likely issues missing from the list below. The changes will require a lot of testing, both in the Pandora environment and in the proper larsoft environment. Comments:

*LArContent*

-I don't think the StitchingCRMergingTool will be reproducible, and certainly its operations will vary from job to job. This is now firmly deprecated. There should no longer be any loops over unordered containers. If you do want to loop over such a container, you need to dump the keys into a vector (or list), sort this vector, loop over the vector and look-up the relevant entry in the unordered container. Examples can be found in e.g. https://github.com/astblake/LArContent/blob/5470b482a68797b9bf899a24288457c71a356e73/larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc#L164

Some relevant areas in this tool include those starting near lines 337 and 568.

-In the StitchingCRMergingTool, you cache all the PointingClusters in a map, then (presumably accidentally?) make a new copy after retrieval from the map on lines 221,222

-The StitchingCRMergingTool and Stitching alg still use NULL, which should be updated to the c++11 nullptr (which has a type, rather than just being a preprocessor paste of 0)

-The old StitchingPfoMergingTool can presumably now be deleted, or edited, renamed and used as some kind of stitching visualization tool.

-Is there a potential to divide by zero in the StitchingHelper, lines 261, 264? I guess the StitchingHelper should move to the LArHelpers directory, for consistency.

-The AreVolumesAdjacent implementation looks like it assumes that volumes will never be stacked in the y or z dimensions? Is this OK? (I assume so.)

-I see the same hard-coded 30cm appearing in multiple locations, but it's difficult to know how best to handle this. I suppose it could be fed-in via the custom LAr geometry, like the wire pitch, but it is definitely not a per-drift volume quantity; it's a per global geometry property. We should have a think about this.

-The LArPointingClusterHelper is now really looking quite complicated. Changes to this may also impact the existing reconstruction, but some quite patrec tests should hopefully show that everything is still OK. In the long term, I don't really like the way that algs follow either a SlidingLinearFit or a PointingCluster approach, when it's almost the very same thing happening underneath. It would be good to unify the algorithm interface. I'd suggest cleaning up and clarifying the SlidingLinearFit interface and adding all PointingCluster functionality to this interface too. (A long term idea, though).

-Would it be possible to add a display of the detector (and maybe drift-volume) edges to the event display sometime, now that all the information is available? This would need care in PandoraMonitoring, which currently tries to remain truly quite generic, or in the existing Pandora detector geometry information.


*In-LArSoft stuff*

-What's the role of the PFParticleStitcher module in DUNEPandora now? Shouldn't it just be removed?

-Should the hit widths now be used for DUNE?

-Can the PCA-based shower characterisation now be enabled for DUNE, via editing of the PandoraSettings file? -> LArPCAShowerBuilding

-Is a feature branch needed for uboonecode too, ensuring that default settings are updated and correct? with e.g. enable readout(?) gaps set to true, truncate readout set to false, etc.


*LArReco*

-This now needs to be able to use a single Pandora instance, or multiple instances according to the user's wish. Can you e.g. modify LArPandoraGeometry to write the master/global drift volume details to the xml file (and do only this if there is only a single volume), then LArReco can sewt things up using this information unless configured to use multiple volumes, when it can continue as it does now.


Thanks a lot - this is a big step forwards and a lot of work!

Cheers,

John

astblake:
Hi John,

As I haven't yet been sent my exam scripts for marking, I have a free weekend! I've updated LArContent according to your comments. I've written some replies below.

I'll do some more testing tomorrow!

Best Wishes,
Andy


LArContent
==========

* Reproducibility issues - FIXED! (I hope?). Most of the operations were supposed to be independent of the ordering of Pfos (for example when every Pfo is compared with every other Pfo).

* LArPointingCluster's - FIXED! (I hope?). These should now be retrieved by reference everywhere.

* NULL vs nullptr - FIXED!

* StitchingPfoMergingTool - DELETED!

* LArStitchingHelper - MOVED! I've moved this helper into the LArHelpers directory. I suspect that MultiPandoraAPI also needs a new home.

* Potential divide-by-zero in LArStitchingHelper? - Good point, I've added an extra check on those variables. There's also a previous check requiring |pX|<1 that should hopefully do the job of preventing pYZ=0 here as well.

* Question about AreVolumesAdjacent method - We only stitch drift volumes that are adjacent in X and at the same level in Y and Z. We also assume that APAs stacked in Y and Z have the same wire angles (which is currently the case in every single-phase TPC).

* Hard-coded 30cm cuts - I added these because I wanted to have some breathing space in my calculations. I'm not certain that adjacent drift volumes will always be exactly level in Y and Z, and exactly touching in X. But I hope these assumptions will always work to within 30cm. 

* LArPointingCluster vs. LArSlidingLinearFit - Yes, that's a longer-term thing, but a good point. The current reason for having LArPointingCluster is that SlidingLinearFit doesn't have a clear forward/backward convention, which makes it harder to check whether two clusters are pointing at each other. LArPointingCluster builds on LArSlidingLinearFit by adding a convention.


LArSoft stuff
=============

* PFParticleStitcher - Yes, this module is now obsolete, but I found that it's still embedded in *so* many 35t scripts (big sigh). I want to wait until later before trying to unravel all the 35t scripts.

* Using hit widths - It would be nice to switch on hit widths in DUNE, but I worry that this creates problems when using cheated hits.

* LArPCAShowerBuilding - This looks okay for cosmic-ray events, but we ought to test this on some beam events before switching it on by default. 

* feature branch of uboonecode? - (good point ... The default settings are fine for MicroBooNE ... Don't need a feature branch unless we want one).


LArReco
=======

* geometry xml file - LArPointingGeometry is modified to write out the appropriate xml file. One big change is that there are now *three* possible versions of these files for multi-drift detectors: (1) multiple drift volumes; (2) multiple drift volumes with U and V views switched; (3) single global drift volume. We'll have to be careful which xml files are cached in LArReco.

PandoraPFA:
Closed via LArSoft coordination meeting, merged by LArSoft team.

PR #7, opened 2017-04-18
Added support vector machine interface.

jaw86:


PandoraPFA:
Merged, with minor edits, via the command line.

PR #17, opened 2017-08-07
Fixing fitted track directions

astblake:
Hi John,

Here's my fix. I've made two changes:

(1) Track directions now always point away from vertex, with no assumptions about z-ordering of MinLayer and MaxLayer in 3D sliding linear fits.

(2) Have move the 3D track trajectory fit from TrackParticleBuildingAlgorithm to PfoHelper. I've also added a second generic fit which takes a vector of 3D coordinates as its input rather than a Pfo. I think this second method might serve a couple of purposes:
    (a) Can call it from a stand-alone LArSoft track-fitting module
    (b) Can apply it to the "cosmic" hypothesis in the cosmic/neutrino separation

Let me know what I need to change !!

Best Wishes,
Andy

### DIFF
--- a/larpandoracontent/LArCustomParticles/TrackParticleBuildingAlgorithm.cc
+++ b/larpandoracontent/LArCustomParticles/TrackParticleBuildingAlgorithm.cc
@@ -13,7 +13,6 @@
 #include "larpandoracontent/LArHelpers/LArPfoHelper.h"
 
 #include "larpandoracontent/LArObjects/LArTrackPfo.h"
-#include "larpandoracontent/LArObjects/LArThreeDSlidingFitResult.h"
 
 #include "larpandoracontent/LArCustomParticles/TrackParticleBuildingAlgorithm.h"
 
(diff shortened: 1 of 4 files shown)


PandoraPFA:
The changes essentially look fine, but there is one issue that is a concern: the bulk repetition of code between the two LArPfoHelper::GetSlidingFitTrajectory functions should be avoided.

It it was not for the fact that we need to store the 2D CaloHits for later access in LArSoft (something that will not be needed for so much longer, hopefully) then you could just have the CaloHit function dump a list of CartesianVectors and then call the CartesianVector version.

With some clever refactoring, we should be able to avoid the repetition. There's probably a multitude of different implementations. For instance, the function could be templated, with some accompanying TypeAdapter functions (with specialisations) to provide the information you need at the point of querying either a CaloHit or a CartesianVector (i.e. position and CaloHit address, nullptr for input CartesianVectors [not so nice, in itself, but hey]).

As another, less elegant approach, you could instead (as a hidden implementation detail) intercept and fill a vector of pairs, of position and CaloHit addresses (again just nullptr for the non-CaloHit version), as the final input to the function that does the work.

Also, it would be good if you could move to using range-based for loops as a matter of course now, to clean up some of the code:

e.g. 

for (CartesianPointVector::const_iterator pIter = pointVector.begin(), pIterEnd =  pointVector.end(); pIter != pIterEnd; ++pIter)
{
    const CartesianVector nextPoint = *pIter;
    ...

should currently actually be:

for (CartesianPointVector::const_iterator pIter = pointVector.begin(), pIterEnd =  pointVector.end(); pIter != pIterEnd; ++pIter)
{
    const CartesianVector &nextPoint = *pIter;
    ...

but would be even better as:

for (const CartesianVector &nextPoint : pointVector)
{
    ...

PandoraPFA:
Can you examine and test feature/TrackTrajectory ? - Thanks!

https://github.com/PandoraPFA/LArContent/blob/feature/TrackTrajectory/larpandoracontent/LArHelpers/LArPfoHelper.cc#L522

https://github.com/PandoraPFA/LArContent/blob/feature/TrackTrajectory/larpandoracontent/LArHelpers/LArPfoHelper.h#L300

PandoraPFA:
Merged, now using template approach described above.

PR #18, opened 2017-08-14
Added the possibility of using... 

loressa:
...SVM probability instead of binary classification


PandoraPFA:
All looks good, but you should: i) check the indentation, and, ii), set the probability cut to a configurable parameter (even though it's almost always going to remain as 0.5f here, it's always best to avoid having a situation where you can't modify cut values without needing to release new code).

loressa:
Aargh, sorry about the indentation. The configurable parameter for the probability cut was a good idea.

PandoraPFA:
Whitespace change, and floating point operator<=

Will fix up and merge via command line.

PandoraPFA:
Merged. Should there be an associated pull request to MachineLearningData, with an updated SVM file?

PR #19, opened 2017-08-15
Strengthen tie-breaking in shower and track overlap result operators

loressa:


PandoraPFA:
As discussed, will test change that leads to a sensible/conceptual operator< i.e. placing "weaker" overlap results behind (return true in operator<) the stronger overlap results in lists accessed by tools.

PandoraPFA:
Merged.

PR #20, opened 2017-08-20
Fixing "GetSlidingFitTrajectory"; extracting shower reconstruction; adding BoxGaps

astblake:
Hi John,

Here's is a pull request based on my work from the past week:

(1) Have added one-line fix to "GetSlidingFitTrajectory" method - the code now runs in LArSoft.

(2) Our 2D and 3D sliding fits will now be forward/backward symmetric: 
       (i) MinLayer and MaxLayer positions are now required to equidistant from the cluster centroid. (Previously, MinLayer was assigned using an extremal hit. However, I found that the layer contributions and fit results were different if the primary axis was reversed; also the extremal hit would often end up in the -1'th layer due to rounding errors. That's all fixed now).
       (ii) 3D cones are now equally spaced along 3D sliding fits, and equidistant from the ends.

(3) Have removed most of the functionality from PcaShowerBuildingAlgorithm:
       (i) Created a LArShowerPCA object in LArObjects/PfoObjects to hold PCA results.
       (ii) Created a "GetPrincipalComponents" method in LArPfoHelper to run the PCA analysis, correctly assign the axis directions, and output a LArShowerPCA object.
       (iii) We should now be able to create stand-alone track-building and shower-building modules in LArSoft using the tools in LArPfoHelper. We can also remove our separate LArTrackPfo and LArShowerPfo objects.
       (iv) Note: we'll need to keep versions of the track-building (and shower-building?) algorithms in LArContent. For one thing, the track-building algorithm tunes the position of the track vertex.
       (v) Further note: I'm sure it's possible to do better than a PCA analysis in Liquid Argon!

(4) Attempted the following pieces of house-keeping:
       (i) Shortened the iterative loops in LArPfoHelper.cc [but do I need to input pPfo->GetClusterList by reference in all the loops?]
       (ii) Added a try/catch block to fix "STATUS_CODE_NOT_FOUND" exceptions thrown from ParticleRecoveryAlgorithm due to invalid sliding linear fits (i.e. fits with a zero x-component).

Let me know what you think.

Best Wishes,
Andy


astblake:
Hi John,

I think that I've also finished my work on BoxGaps this afternoon. Here's a quick summary:

 (a) Added some parallel methods for BoxGaps in LArGeometryHelper (alongside the existing treatment of LineGaps).
 (b) Modified CrossGapsExtensionAlgorithm so that it can make use of BoxGaps - previously it just considered the start/end z positions of gaps and clusters.
 (c) CrossGapsAssociationAlgorithm requires no change.

The BoxGaps themselves will need to be created in larpandora - we'll need to add a few lines of code in LArPandoraInput at some point.

One spanner-in-the-works is that DUNE accelerator events seem to have non-zero T0 values, so the hits appear slightly displaced from the gaps.

Best Wishes,
Andy

larpandoracontent/LArObjects/LArPfoObjects.cc
```diff
@@ -0,0 +1,128 @@
...
+        return m_pCaloHit;
+
+    throw StatusCodeException(STATUS_CODE_NOT_INITIALIZED);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+LArShowerPCA::LArShowerPCA(const CartesianVector &centroid, const CartesianVector &primaryAxis, const CartesianVector &secondaryAxis,
+    const CartesianVector &tertiaryAxis, const CartesianVector &eigenValues) :
```
PandoraPFA: Check indentation of constructor arguments  vs. initializer list.

larpandoracontent/LArObjects/LArThreeDSlidingConeFitResult.cc
```diff
@@ -117,16 +117,24 @@ void ThreeDSlidingConeFitResult::GetSimpleConeList(const unsigned int nLayersFor
...
-    unsigned int nConeSamplingSteps(0);
-    const unsigned int coneInterval((nCones > 1) ? (nLayers - nLayersForConeFit) / (nCones - 1) : std::numeric_limits<unsigned int>::max());
+    // Calculate intervals and offsets such that cones are evenly spaced along sliding fit and are equidistant from the extremal layers
+    const unsigned int coneInterval((nCones > 1) ? (nLayers - nLayersForConeFit) / (nCones - 1) : 1);
+    const ThreeDSlidingFitResult &slidingFitResult(this->GetSlidingFitResult());
+    const CartesianVector coneDisplacement(slidingFitResult.GetGlobalMaxLayerPosition() - slidingFitResult.GetGlobalMinLayerPosition());
+    const bool isForward(coneDisplacement.GetZ() > std::numeric_limits<float>::epsilon());
+    const unsigned int coneOffset1((nLayers - nLayersForConeFit - (nCones - 1) * coneInterval) / 2);
+    const unsigned int coneOffset2((1 == nLayers % 2) && isForward ? 1 : 0);
+    const unsigned int coneOffset(coneOffset1 + coneOffset2);
```
PandoraPFA: This logic is pretty tortuous, but seems sound - will have to read through it again later to double check!

larpandoracontent/LArObjects/LArThreeDSlidingFitResult.cc
```diff
@@ -224,12 +224,20 @@ TrackState ThreeDSlidingFitResult::GetPrimaryAxis(const CartesianPointVector *co
...
+    // However, downstream algorithms should be independent of this convention.
+    if (fitDirection.GetZ() < 0.f)
+        fitDirection *= -1.f;
 
     for (const CartesianVector &coordinate : *pPointVector)
         minProjection = std::min(minProjection, fitDirection.GetDotProduct(coordinate - centroid));
 
-    return TrackState(centroid + (fitDirection * minProjection), fitDirection);
+    // Define layers based on centroid rather than individual extremal hits
+    const float fitProjection(layerPitch * std::floor(minProjection / layerPitch));
```
PandoraPFA: Is there a way that minProjection might still be std::numeric_limits<float>::max() at this point?

Yes, but only if all the fitDirection.GetDotProduct(coordinate - centroid) results on line 235 also return float max. Technically possible, but vanishingly unlikely in a real use case.

larpandoracontent/LArObjects/LArTwoDSlidingFitResult.cc
```diff
@@ -572,7 +572,10 @@ void TwoDSlidingFitResult::CalculateAxes(const CartesianPointVector &coordinateV
     for (const CartesianVector &coordinate : coordinateVector)
         minProjection = std::min(minProjection, fitDirection.GetDotProduct(coordinate - centroid));
 
-    m_axisIntercept = centroid + (fitDirection * minProjection);
+    // Define layers based on centroid rather than individual extremal hits
+    const float fitProjection(layerPitch * std::floor(minProjection / layerPitch));
```
PandoraPFA: Same comment as above, regarding minProjection still holding a float max value and then (potentially) being divided by a number less than one.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -394,11 +394,26 @@ CartesianVector LArGeometryHelper::GetWireAxis(const Pandora &pandora, const Hit
...
+        }
+
+        // Check box gaps (need to input 3D hit type)
+        const BoxGap *const pBoxGap = dynamic_cast<const BoxGap*>(pDetectorGap);
+
+        if (pBoxGap)
+        {
+            if (pBoxGap->IsInGap(testPoint2D, TPC_3D, gapTolerance))
+                return true;
+        }
```
PandoraPFA: I don't think we can proceed with this as a valid solution - it is working hard to subvert the original, intended use of detector gaps, with a IsInGap function in a purely abstract base class.

I think it would all be much simpler if we modify the LineGap class in the SDK to hold information about whether the gap is in the drift direction or the wire direction.

All of these (slow, often deprecated) dynamic casts should disappear as would the separate treatments for LineGaps and BoxGaps below.

BoxGaps could indeed be used for work that is truly in a 3D coordinate system.

The LineGap modification would not take long at all, and there would need to be an associated tweak to the display of LineGaps in PandoraMonitoring.

What do you think?

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -394,11 +394,26 @@ CartesianVector LArGeometryHelper::GetWireAxis(const Pandora &pandora, const Hit
 
 bool LArGeometryHelper::IsInGap(const Pandora &pandora, const CartesianVector &testPoint2D, const HitType hitType, const float gapTolerance)
 {
-    // ATTN: input test point MUST be a 2D position vector 
+    // ATTN: input test point MUST be a 2D position vector
     for (const DetectorGap *const pDetectorGap : pandora.GetGeometry()->GetDetectorGapList())
     {
-        if (pDetectorGap->IsInGap(testPoint2D, hitType, gapTolerance))
-            return true;
+        // Check line gaps (need to input 2D hit type)
```
PandoraPFA: We need a way of enforcing this!

PandoraPFA:
Merged in part, but now requires review of remaining aspect (use of drift-type line gaps in reconstruction) and resubmission against current master branches.

PR #21, opened 2017-11-07
Svm pfo characterisation using 3d information 

loressa:
These are the changes to LArContent for the updated svm pfo characterisation using 3D information. While cleaning the code prior to this PR, I modified some classes wrt my local changes. It has been checked that it preserves the same results and hopefully it's better structured and have less repetitions than my local versions. In any case, we can iterate on this, as I said I made these modifications recently, so maybe better options are to appear after another look (and/or pair of eyes). Some of these changes to be highlighted:
    - Edited SvmPfoCharacterisationBaseAlgorithm to work with either 2D information at the cluster level (and then a minimum number of views to separate tracks and showers), as used previously, or 3D information at the PFO level. For this reason I modified the algorithms inheriting from this algorithm (i.e. cut-based and svm-based) to use a local version of IsClearTrack(pfo), which now in the base algorithm is replaced by IsTrackLike(pfo). Option m_useThreeDInformation added.
    - Removed the terrible use of featureVector[0] assumed as length always to produce the ratio of variables, now calculated inside the appropriate TrackShowerIDFeatureTools. 
    - Should work as it is at the moment. To replace the 2D -> 3D SVM characterisation, changes to MachineLearningData and settings files are to be included as well (PRs to follow)

Hopefully indentation is all good this time and no tabs remained after my checks.


larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.h
```diff
@@ -26,21 +26,28 @@ class SvmPfoCharacterisationAlgorithm : public PfoCharacterisationBaseAlgorithm
...
+    virtual bool IsClearTrack(const pandora::ParticleFlowObject *const pPfo) const;
     virtual bool IsClearTrack(const pandora::Cluster *const pCluster) const;
     pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
 
-    ClusterCharacterisationFeatureTool::FeatureToolVector   m_featureToolVector;    ///< The feature tool map
-    SupportVectorMachine    m_supportVectorMachine;         ///< The support vector machine
+    ClusterCharacterisationFeatureTool::FeatureToolVector   m_featureToolVector;         ///< The feature tool map
+    PfoCharacterisationFeatureTool::FeatureToolVector       m_featureToolVectorThreeD;   ///< The feature tool map for 3D info
+    PfoCharacterisationFeatureTool::FeatureToolVector       m_featureToolVectorMissingW; ///< The feature tool map for missing W view
+
```
PandoraPFA: Going back to earlier comments, we should ask whether we still want to run the 3x2D version here, or whether we could simplify everything (including the base class interface!) by just now running with Pfos and in 3D. Are there any planned usages of the 3x2D approach still?

Removing all the 3x2D stuff, including the IsClearTrack(const Cluster*) callback would definitely tidy things up. An intermediate case might be to add two different algorithms, rather than one algorithm that has a single ShouldUseThreeDInfo setting.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -38,7 +42,26 @@ void ShowerFitFeatureTool::Run(SupportVectorMachine::DoubleVector &featureVector
...
         std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
 
-    featureVector.push_back(CutClusterCharacterisationAlgorithm::GetShowerFitWidth(pAlgorithm, pCluster, m_slidingShowerFitWindow));
+    if (m_ratioVariables)
+    {
+        float ratio(-1.f);
+        try
+        {
+            const TwoDSlidingFitResult slidingFitResultLarge(pCluster, m_slidingLinearFitWindow, LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+            float straightLineLength = (slidingFitResultLarge.GetGlobalMaxLayerPosition() - slidingFitResultLarge.GetGlobalMinLayerPosition()).GetMagnitude();
```
PandoraPFA: const float

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -38,7 +42,26 @@ void ShowerFitFeatureTool::Run(SupportVectorMachine::DoubleVector &featureVector
...
+        }
+        catch (const StatusCodeException &)
+        {
+            ratio = -1.f;
+        }
+        featureVector.push_back(ratio);
+    }
+    else
+    {
+        featureVector.push_back(CutClusterCharacterisationAlgorithm::GetShowerFitWidth(pAlgorithm, pCluster, m_slidingShowerFitWindow));
```
PandoraPFA: Some shared aspects here between the with/without ratioVariables code. There is good exception handling for the ratio variable case, but nothing for the other case. Would suggest calculate shower fit width in same way (with exception handling) regardless of ratioVariables option. Then, for ratio option only, obtain straight line length (with exception handling) and divide.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -95,7 +126,17 @@ const pandora::Cluster * const pCluster)
...
+    if (m_ratioVariables && (straightLineLengthLarge > std::numeric_limits<double>::epsilon()))
+    {
+        diffWithStraightLineMean  /= straightLineLengthLarge;
+        diffWithStraightLineSigma /= straightLineLengthLarge;
+        dTdLWidth                 /= straightLineLengthLarge;
+        maxFitGapLength           /= straightLineLengthLarge;
+        rmsSlidingLinearFit       /= straightLineLengthLarge;
+    }
+
+    if(m_addStraightLineLength)
```
PandoraPFA: Prefer "if (...)" - i.e. space between "f" and "("

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+        {
+            ratio = -1.f;
+        }
+        featureVector.push_back(ratio);
+    }
+    else
+    {
+        featureVector.push_back((CutClusterCharacterisationAlgorithm::GetVertexDistance(pAlgorithm, pCluster)));
+    }
+}
```
PandoraPFA: See above comments. An important question (similar to discussion above) is the expectation for use with and without ratio variables. Is it worth supporting ratio variables on/off if this will only ever be used in one configuration? The ratioVarables option certainly has a big footprint in the implementation!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+
+        diffWithStraightLineMean   += diffWithStraightLineMeanCluster;
+        diffWithStraightLineSigma  += diffWithStraightLineSigmaCluster;
+        dTdLWidth                  += dTdLWidthCluster;
+        maxFitGapLength            += maxFitGapLengthCluster;
+        rmsSlidingLinearFit        += rmsSlidingLinearFitCluster;
+    }
+
+    const int nClusters(clusterList.size());
+    diffWithStraightLineMean   /= nClusters;
```
PandoraPFA: Care here, as dividing a floating point value by an int. This involves an implicit cast, which would be better as const float nClusters(static_cast<float>(clusterList.size()); Do also need to do something in case nClusters is zero.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    diffWithStraightLineSigma  /= nClusters;
+    dTdLWidth                  /= nClusters;
+    maxFitGapLength            /= nClusters;
+    rmsSlidingLinearFit        /= nClusters;
+
+    if(m_addStraightLineLength)
```
PandoraPFA: if (

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+
+        FloatVector diffWithStraightLineVector;
+        const HitType hitType(LArClusterHelper::GetClusterHitType(pCluster));
+        CartesianVector previousFitPosition(slidingFitResult.GetGlobalMinLayerPosition());
+        float dTdLMin(+std::numeric_limits<float>::max()), dTdLMax(-std::numeric_limits<float>::max());
+
+        for (const auto &mapEntry : slidingFitResult.GetLayerFitResultMap())
+        {
+            rmsSlidingLinearFit += mapEntry.second.GetRms();
+
```
PandoraPFA: Might want to cache mapEntry.second as a const reference to a LayerFitResult here, just to improve readability.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    (void) PandoraContentApi::GetCurrentList(*pAlgorithm, pVertexList);
+
+    if (pVertexList->empty() || (pVertexList->size() != 1) || (VERTEX_3D != pVertexList->front()->GetVertexType()))
+    {
+        throw StatusCodeException(STATUS_CODE_FAILURE);
+    }
+
+    const Vertex *const nuVertex(pVertexList->front());
+    //find the particle vertex
+    const Vertex *const pVertex = LArPfoHelper::GetVertex(pInputPfo);
```
PandoraPFA: Presumably there's no way that the input pfo is a neutrino here (with the neutrino vertex)?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    const CartesianVector nuPosition(nuVertex->GetPosition()), pPosition(pVertex->GetPosition());
```
PandoraPFA: Prefer particlePosition, otherwise looks like a pointer variable, pVariable.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+
+    float vertexDistance((nuPosition - pPosition).GetMagnitudeSquared());
```
PandoraPFA: Potentially important: this is vertex distance *squared*. Are your sure  this is what you intended?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+            LArPfoHelper::GetThreeDClusterList(pInputPfo, clusterList);
+            if (1 == clusterList.size())
+            {
+                const Cluster *const pCluster(clusterList.front());
+                const ThreeDSlidingFitResult sliding3DFitResult(pCluster, m_slidingLinearFitWindow, LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+                const TwoDSlidingFitResult slidingFitResult(sliding3DFitResult.GetFirstFitResult());
+                float straightLineLength = (slidingFitResult.GetGlobalMaxLayerPosition() - slidingFitResult.GetGlobalMinLayerPosition()).GetMagnitude();
+                if (straightLineLength > std::numeric_limits<double>::epsilon())
+                    ratio = vertexDistance / straightLineLength;
+            }
```
PandoraPFA: The ratio variables option really does have a large code footprint. If we're using ratio variables, we should think of a way of tidying this; otherwise we can enjoy removing a lot of this stuff!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+        std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    // Need the 3D clusters and hits to calculate PCA components
+    ClusterList threeDClusterList;
+    LArPfoHelper::GetThreeDClusterList(pInputPfo, threeDClusterList);
+
+    CaloHitList threeDCaloHitList; // Might be able to get through LArPfoHelper instead
+    const Cluster *const pThreeDCluster(threeDClusterList.front());
+    pThreeDCluster->GetOrderedCaloHitList().FillCaloHitList(threeDCaloHitList);
```
PandoraPFA: Should now be able to get all 3D hits directly from the Pfo, using the LArPfoHelper.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+void OpeningAngleFeatureTool::Divide3DCaloHitList(const Algorithm *const pAlgorithm, CaloHitList threeDCaloHitList, CartesianPointVector &pointVectorStart, CartesianPointVector &pointVectorEnd)
+{
+    const VertexList *pVertexList = nullptr;
+    (void) PandoraContentApi::GetCurrentList(*pAlgorithm, pVertexList);
+
+    if (!pVertexList || (pVertexList->size() != 1) || (VERTEX_3D != pVertexList->front()->GetVertexType()))
+        throw StatusCodeException(STATUS_CODE_FAILURE);
+
+    const Vertex *const pVertex(pVertexList->front());
+    const CartesianVector nuVertex(pVertex->GetPosition().GetX(),pVertex->GetPosition().GetY(),pVertex->GetPosition().GetZ());
```
PandoraPFA: Could just have ```const CartesianVector &nuVertex(pVertexList->front()->GetPosition());```

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    CaloHitVector threeDCaloHitVector(threeDCaloHitList.begin(), threeDCaloHitList.end());
+
+    //order by distance to vertex, so first ones are closer to nuvertex
+    std::sort(threeDCaloHitVector.begin(), threeDCaloHitVector.end(), std::bind(ChargeFeatureTool::SortByDistanceToVertex, std::placeholders::_1, std::placeholders::_2, nuVertex));
+    CaloHitList orderedCaloHitList(threeDCaloHitVector.begin(),threeDCaloHitVector.end());
+    const int nhits(orderedCaloHitList.size());
```
PandoraPFA: unsigned int
Can you make the SortByDistanceToVertex function static (and stateless!) and avoid the use of std::bind? This might not be possible because of the vertex argument... happy to test/discuss!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+
+    for (const CaloHit *const pCaloHit : orderedCaloHitList)
+    {
+        if (pointVectorStart.size()<(nhits/2))//half in start, half in end
```
PandoraPFA: Could shorten number of lines of code, by e.g. ```CartesianPointVector &targetVector((pointVectorStart.size() < (nhits / 2)) ? pointVectorStart : pointVectorEnd); targetVector.push_back(pCaloHit->GetPositionVector());```

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    {
+        std::cout << "PcaShowerParticleBuildingAlgorithm::OpeningAngle - principal eigenvalue less than or equal to 0." << std::endl;
+        throw StatusCodeException( STATUS_CODE_INVALID_PARAMETER );
+    }
+    else if (std::fabs(eigenValues.GetY()) < std::numeric_limits<float>::epsilon())
+    {
+        return 0.f;
+    }
+
+    return std::atan(std::sqrt(eigenValues.GetY()) * sinTheta / std::sqrt(eigenValues.GetX()));
```
PandoraPFA: atan vs. atan2?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+        LArPfoHelper::GetThreeDClusterList(pInputPfo, threeDClusterList);
+
+        if (threeDClusterList.empty())
+            return;
+
+        CaloHitList threeDCaloHitList; // Might be able to get through LArPfoHelper instead
+        const Cluster *const pThreeDCluster(threeDClusterList.front());
+        pThreeDCluster->GetOrderedCaloHitList().FillCaloHitList(threeDCaloHitList);
+
+        if (threeDCaloHitList.empty())
```
PandoraPFA: See earlier comment about getting 3D hits from Pfos.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+            float pca1(-1.f), pca2(-1.f);
+            if(principalEigenvalue > std::numeric_limits<float>::epsilon())
+            {
+                pca1 = secondaryEigenvalue/principalEigenvalue;
+                pca2 = tertiaryEigenvalue/principalEigenvalue;
+            }
+            featureVector.push_back(pca1);
+            featureVector.push_back(pca2);
+        }
+        else
```
PandoraPFA: Different numbers of features with, without ratio variables?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.h
```diff
@@ -198,6 +208,212 @@ class VertexDistanceFeatureTool : public ClusterCharacterisationFeatureTool
...
+    OpeningAngleFeatureTool();
+
+    void Run(SupportVectorMachine::DoubleVector &featureVector, const pandora::Algorithm *const pAlgorithm, const pandora::ParticleFlowObject *const pInputPfo);
+
+private:
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+    void Divide3DCaloHitList(const pandora::Algorithm *const pAlgorithm, pandora::CaloHitList threeDCaloHitList,
+        pandora::CartesianPointVector &pointVectorStart, pandora::CartesianPointVector &pointVectorEnd);
+
+    //static bool SortByDistanceToVertex(const pandora::CaloHit *const left, const pandora::CaloHit *const right, const pandora::CartesianVector &nuVertex);
```
PandoraPFA: As discussed above (might not work)!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+            {
+                startCharge += pCaloHitCharge;
+            }
+            if (hitCounter >= std::floor(totalHits*(1.f-endChargeFraction)))
+            {
+                endCharge += pCaloHitCharge;
+            }
+
+            if(hitCounter>1)
+                incrementCharge = (pCaloHitCharge-previousCharge)*(pCaloHitCharge-previousCharge);
```
PandoraPFA: Why squared? Some kind of RMS-like variable?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    m_addStartCharge(false),
+    m_addEndCharge(true),
+    m_endChargeFraction(0.1f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ChargeFeatureTool::CalculateChargeVariables(const Algorithm *const pAlgorithm, const pandora::Cluster *const pCluster, float &totalCharge, float &chargeSigma, 
+    float &chargeMean, float &startCharge, float &endCharge, float &incrementCharge, float endChargeFraction)
```
PandoraPFA: Might be better to create a little class to hold all these variables, with a default constructor setting them all to their initial values as appropriate. Then just have a single argument, by non-const reference, to pass in and receive populated ChargeResults.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+    CaloHitList clusterCaloHitList;
+    pCluster->GetOrderedCaloHitList().FillCaloHitList(clusterCaloHitList);
+    CaloHitVector clusterCaloHitVector(clusterCaloHitList.begin(), clusterCaloHitList.end());
+
+    //TODO: might give problems if vertex in the middle of the cluster ?
+    std::sort(clusterCaloHitVector.begin(), clusterCaloHitVector.end(), std::bind(SortByDistanceToVertex, std::placeholders::_1, std::placeholders::_2, vertexPosition2D));
+    CaloHitList orderedCaloHitList(clusterCaloHitVector.begin(),clusterCaloHitVector.end());
+
+    return orderedCaloHitList;
+}
```
PandoraPFA: Maybe provide orderedCaloHitList as a non-const reference argument; can then fill and sort the container directly, rather than returning a copy.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -326,19 +373,781 @@ StatusCode MipEnergyFeatureTool::ReadSettings(const TiXmlHandle xmlHandle)
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool ChargeFeatureTool::SortByDistanceToVertex(const CaloHit *const left, const CaloHit *const right, const CartesianVector &nuVertex2D)
+{
+    float distanceL((left->GetPositionVector()-nuVertex2D).GetMagnitude());
+    float distanceR((right->GetPositionVector()-nuVertex2D).GetMagnitude());
```
PandoraPFA: Now can avoid taking square root and just sort based upon MagnitudeSquared (square root rather slow).

loressa:
Hi John,

Thanks for the comments, I will look into them in detail. 

The vertex squared distance is interesting. I don’t think it was consciously intended, I probably just forgot that was squared. It doesn’t matter in terms of results as long as it is used as the model was created, but I agree it’s better not to have it squared without a reason. I’ll have a look, otherwise if it needs to change it means I need to create another model, test, etc. So my reply is going to take several days. 

About the other two ones, I was thinking the contrary, that we probably still want to have the 3x2D approach, so it’s still possible to use the previous model we provided, although this could still be done by checking out a previous version of LArContent; but that ratioVariables, although it’s a potentially interesting thing to change, results have show us that it’s better to divide by the length and in practice we don’t use ratioVariables = false, and the models provided are all with ratioVariables = true. So maybe the best for simplicity is to drop both things?

I need to think about some things. For example, related to the class for the variables for the charge feature tool, I realise that, although I added several variables there, which were interesting to test and play with, in the end I am only using two of them (i.e. some of the addX are meant to be always false), so maybe it’s better for simplicity to only have the code of the things that will be used in the end? 

Best,
Lorena

PandoraPFA:
Hi Lorena,

Don't worry about making all the changes instantly; as you know, there are already developments queued up to get into LArSoft before this can go in.

Your comments about the 3x2D approach and useRatioVariables aspects are interesting.

It's nice to be able to offer fallback (to 3x2D) functionality, to allow use of an old approach just by changing the xml config and no implementation code. However, we need to ask ourselves whether we will ever truly do this... If the answer is "no", we should clean up the PfoCharacterisation, by removing the IsClearTrack(const Cluster*) callback. If the answer is "it's not very likely", we could separate out the two approaches into two algorithms, one of which (3x2D) can be trivially deleted in time. We probably don't want the two approaches in one algorithm, so that it becomes rather involved maintenance to remove the 3x2D approach in future, and this removal then never happens.

It sounds like useRatioVariables is here to stay, switched on, then. In this case, it does remove some of the problems with there being no exception handling if useRatioVariables is false (this code disappears!). One thing we should do is refactor the straightLineLength calculation into a single, reusable function accessible to all the different tools (static, stateless function?), as this should reduce the number of lines of implementation in each tool considerably (and this function can do all the required exception handling for this tricky variable in one place).

It will be interesting to see if using the raw vertex distance, rather than distance squared, makes any difference. Not squaring makes the variable a little better motivated, but, as you say, the SVM might just have brought everything under control again anyhow.

Thanks!

PandoraPFA:
A code snippet I found that should indicate how we can avoid the use of the ugly std::bind when trying to sort hits/clusters/etc by distance to a specified vertex object:

```
class cmp {
    int param;
public:
    cmp(int p) : param(p) {}

    bool operator()(int i, int j) {
        // logic uses param
    }
};

int main() {
    std::vector<int> v(100);
    // initialize v with some random values

    int paramA = 4;

    std::sort(v.begin(), v.end(), cmp(paramA));
}
```

Compared to this example, we would have a vector of CaloHits, and paramA could be the address of the vertex, or it's position as a CartesianVector.

The call would then look something like:
```std::sort(hitVector.begin(), hitVector.end(), vertexComparator(pNeutrinoVertex));```

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 24 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok, didn't know std::sqrt was that slow

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 23 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 21 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: it is, but probably not worth keeping it, see previos message

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 22 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: could be done that way, but following a comment I made later on, maybe it's better to remove the features I didn't use in the end? In that way we would only have two variables filled and would simplify this a lot. I added (almost) all the different features I tested, but maybe it's not really interesting other than for myself in development and I could just remove some stuff for simplicity, thoughtS?

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 19 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: different indeed, but should be irrelevant if the ratioVariables off option is dropped

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 17 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: have only tried atan

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 16 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: oh, I see, yes.

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 15 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: saw your alternative for the std::bind later on, which looks good to me. I'll have a look to remove the std::bind (didn't know about its ugliness)

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 14 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: It's funny how I overcomplicate things sometimes XD

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 13 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok, I'll have a look.

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 12 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: I would/will enjoy removing a lot of stuff

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 11 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: to be checked.

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 10 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 8 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 7 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: damn

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 6 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: got it

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 5 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: under discussion as well

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 4 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: its the only one missing!

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 3 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok, I'll have a look

File: larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc (hunk 2 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc]: ok

File: larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.h (hunk 1 above)
Comment (loressa) [reply larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.h]: under discussion

File: larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc [hunk 25]
```diff
@@ -112,7 +158,7 @@ StatusCode SvmPfoCharacterisationAlgorithm::ReadSettings(const TiXmlHandle xmlHa
             return STATUS_CODE_INVALID_PARAMETER;
         }
 
-        std::string fullSvmFileName(m_svmFileName);
+        std::string fullSvmFileName(m_svmFileName), fullSvmFileNameMissingW(m_svmFileNameMissingW);;
 #ifdef CETLIB_AVAILABLE
```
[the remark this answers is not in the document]
loressa: I've noticed. that was updated after I did this pull request, and I saw the reason of "merges needed" issues right now. I'll update.

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -38,41 +40,76 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+            trueMCPdg = pMCParticle->GetParticleId();
         }
         catch (const StatusCodeException &) {}
 
-        LArSvmHelper::ProduceTrainingExample(m_trainingOutputFile, isTrueTrack, featureVector);
+        if (trueMCPdg!=0)
+        {
+	  std::string outputFile;
+            outputFile.append(m_trainingOutputFile);
+            const char* end=((wClusterList.empty()) ? "MissingW.txt" : ".txt");
```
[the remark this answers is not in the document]
loressa: ok

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -38,41 +40,76 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+
     if (m_trainingSetMode)
     {
         bool isTrueTrack(false);
+        int trueMCPdg(0);
 
         try
         {
-            const MCParticle *const pMCParticle(MCParticleHelper::GetMainMCParticle(pCluster));
+            const MCParticle *const pMCParticle(LArMCParticleHelper::GetMainMCParticle(pPfo));
```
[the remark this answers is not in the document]
loressa: that would be something useful, in fact I realized of the 20% of cases in which it will fail (as discussed, also with Jack, in a group meeting). However in my case here I am not interested in having something else, I'd rather not consider those cases with PFOs mixed of different MC for the training examples (the purpose of using this), that's why I added a check so that if this fails I won't include that PFO in the training file

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -38,41 +40,76 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+    LArPfoHelper::GetClusters(pPfo, TPC_VIEW_W, wClusterList);
+
+    PfoCharacterisationFeatureTool::FeatureToolVector featureToolVector(wClusterList.empty() ? m_featureToolVectorMissingW : m_featureToolVectorThreeD);
+    SupportVectorMachine::DoubleVector featureVector(LArSvmHelper::CalculateFeatures(featureToolVector, this, pPfo));
+    SupportVectorMachine supportVectorMachine(wClusterList.empty() ? m_supportVectorMachineMissingW : m_supportVectorMachine);
+
     if (m_trainingSetMode)
     {
         bool isTrueTrack(false);
+        int trueMCPdg(0);
```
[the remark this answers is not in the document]
loressa: sounds good

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -38,41 +40,76 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowObject *const pPfo) const
+{
+    //charge related features are only calculated using hits in W view
+    ClusterList wClusterList;
+    LArPfoHelper::GetClusters(pPfo, TPC_VIEW_W, wClusterList);
+
+    PfoCharacterisationFeatureTool::FeatureToolVector featureToolVector(wClusterList.empty() ? m_featureToolVectorMissingW : m_featureToolVectorThreeD);
+    SupportVectorMachine::DoubleVector featureVector(LArSvmHelper::CalculateFeatures(featureToolVector, this, pPfo));
+    SupportVectorMachine supportVectorMachine(wClusterList.empty() ? m_supportVectorMachineMissingW : m_supportVectorMachine);
+
```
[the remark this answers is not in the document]
loressa: sounds good, what about noChargeInfo?

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -38,41 +40,76 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowObject *const pPfo) const
+{
+    //charge related features are only calculated using hits in W view
+    ClusterList wClusterList;
+    LArPfoHelper::GetClusters(pPfo, TPC_VIEW_W, wClusterList);
+
+    PfoCharacterisationFeatureTool::FeatureToolVector featureToolVector(wClusterList.empty() ? m_featureToolVectorMissingW : m_featureToolVectorThreeD);
+    SupportVectorMachine::DoubleVector featureVector(LArSvmHelper::CalculateFeatures(featureToolVector, this, pPfo));
+    SupportVectorMachine supportVectorMachine(wClusterList.empty() ? m_supportVectorMachineMissingW : m_supportVectorMachine);
```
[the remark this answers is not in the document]
loressa: ok

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -38,41 +40,76 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowObject *const pPfo) const
+{
+    //charge related features are only calculated using hits in W view
+    ClusterList wClusterList;
+    LArPfoHelper::GetClusters(pPfo, TPC_VIEW_W, wClusterList);
+
+    PfoCharacterisationFeatureTool::FeatureToolVector featureToolVector(wClusterList.empty() ? m_featureToolVectorMissingW : m_featureToolVectorThreeD);
+    SupportVectorMachine::DoubleVector featureVector(LArSvmHelper::CalculateFeatures(featureToolVector, this, pPfo));
```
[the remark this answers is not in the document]
loressa: yup

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -22,8 +24,8 @@ namespace lar_content
 
 SvmPfoCharacterisationAlgorithm::SvmPfoCharacterisationAlgorithm() :
     m_trainingSetMode(false),
-    m_ratioVariables(true),
     m_enableProbability(false),
+    m_useThreeDInformation(false),
```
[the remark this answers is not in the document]
loressa: good point. Moving to the new 3D set as default, then yes. I'll change it.

larpandoracontent/LArTrackShowerId/PfoCharacterisationBaseAlgorithm.cc
```diff
@@ -56,7 +57,7 @@ StatusCode PfoCharacterisationBaseAlgorithm::Run()
         {
             PandoraContentApi::ParticleFlowObject::Metadata pfoMetadata;
 
-            if (this->IsClearTrack(pPfo))
+            if (this->IsTrackLike(pPfo))
```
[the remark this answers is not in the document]
loressa: could be done differently. I opted for this way so the IsClearTrack remains as it is in CutPfo for instance

larpandoracontent/LArContent.cc
```diff
@@ -302,9 +302,16 @@
...
     d("LArNNearbyClustersFeatureTool",          NNearbyClustersFeatureTool)                                                     \
     d("LArMipEnergyFeatureTool",                MipEnergyFeatureTool)                                                           \
-    d("LArVertexDistanceFeatureTool",           VertexDistanceFeatureTool)
+    d("LArVertexDistanceFeatureTool",           VertexDistanceFeatureTool)                                                      \
+    d("LArThreeDVertexDistanceFeatureTool",     ThreeDVertexDistanceFeatureTool)                                                \
+    d("LArChargeFeatureTool",                   ChargeFeatureTool)                                                              \
+    d("LArThreeDChargeFeatureTool",             ThreeDChargeFeatureTool)                                                        \
+    d("LArPCAFeatureTool",                      PCAFeatureTool)                                                                 \
+    d("LArOpeningAngleFeatureTool",             OpeningAngleFeatureTool)                                                        
+
```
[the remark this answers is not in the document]
loressa: hmmm - depends if we are going to maintain both 2D and 3D implementations. If we do, what about LArTwoDXFeatureTool and LArThreeDXFeatureTool? shorter than ClusterCharacterisation while understanding the same

larpandoracontent/LArTrackShowerId/CutPfoCharacterisationAlgorithm.cc
```diff
@@ -78,6 +78,13 @@ bool CutPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+bool CutPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowObject *const /*pPfo*/) const
+{
+	throw StatusCodeException(STATUS_CODE_FAILURE);
+}
+
```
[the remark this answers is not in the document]
loressa: somehow. Same thing is present in the CheatingPfoCharacterisationAlgorithm, which I followed for this: https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArCheating/CheatingPfoCharacterisationAlgorithm.cc

loressa:
Just did a commit, need to fix indentation and double check, then follow with commits in LArReco and MachineLearningData, so will follow up on it.

larpandoracontent/LArTrackShowerId/SvmClusterCharacterisationAlgorithm.h
```diff
@@ -1,5 +1,5 @@
 /**
- *  @file   larpandoracontent/LArTrackShowerId/SvmClusterCharacterisationAlgorithm.h
+ *  @file   larpandoracontent/LArTrackShowerId/SvmSvmClusterCharacterisationAlgorithm.h
```
PandoraPFA: Should remove this `SvmSvm`

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -5,10 +5,15 @@
  * 
  *  $Log: $
  */
+#ifdef CETLIB_AVAILABLE
```
PandoraPFA: Don't need this any more; won't compile if `CETLIB_AVAILABLE` is actually defined.

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -35,42 +40,79 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
     }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowObject *const pPfo) const
+{
+
+    if (!LArPfoHelper::IsThreeD(pPfo))
+        return (pPfo->GetParticleId() == MU_MINUS);
```
PandoraPFA: So everything is a track if it doesn't have any 3D hits? Are we sure this is the right thing to do? Maybe redirect to 3x2D (although there might only be e.g. 1x2D or 2x2D available in this case!)?

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -35,42 +40,79 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
+
+    //charge related features are only calculated using hits in W view
+    ClusterList wClusterList;
+    LArPfoHelper::GetClusters(pPfo, TPC_VIEW_W, wClusterList);
+
+    PfoCharacterisationFeatureTool::FeatureToolVector featureToolVector(wClusterList.empty() ? m_featureToolVectorNoChargeInfo : m_featureToolVectorThreeD);
```
PandoraPFA: Should be a const reference, to avoid copying the feature tool vector. This wouldn't work if it wasn't for the fact that there's only pointers to feature tools in the vector!

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -35,42 +40,79 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
 
     if (m_trainingSetMode)
     {
         bool isTrueTrack(false);
+        bool isMainMCParticleSet(false);
 
         try
         {
-            const MCParticle *const pMCParticle(MCParticleHelper::GetMainMCParticle(pCluster));
+            const MCParticle *const pMCParticle(LArMCParticleHelper::GetMainMCParticle(pPfo));
```
PandoraPFA: I've now deleted this function in an upcoming feature branch! Will need to work out what to do here...

Maybe just leave this one with me; I might leave the function in now that the rest of `LArMCParticleHelper` will be so much tidier, anyway.

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -35,42 +40,79 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const Cluster *const pCluster
...
             isTrueTrack = ((PHOTON != pMCParticle->GetParticleId()) && (E_MINUS != std::abs(pMCParticle->GetParticleId())));
+            isMainMCParticleSet = pMCParticle->GetParticleId();
```
PandoraPFA: implicit `int` to `bool` conversion

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.h
```diff
@@ -94,110 +81,207 @@ class LinearFitFeatureTool : public ClusterCharacterisationFeatureTool
...
+class ThreeDChargeFeatureTool : public PfoCharacterisationFeatureTool
+{
+public:
     /**
-     *  @brief  Calculation of vertex distance
+     *  @brief  Default constructor
+     */
+    ThreeDChargeFeatureTool();
+
+	class VertexComparator
```
PandoraPFA: Indentation; here and line 243, below. Missing class doxygen.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -25,68 +27,61 @@ using namespace pandora;
...
     if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
         std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
 
-    featureVector.push_back(CutClusterCharacterisationAlgorithm::GetShowerFitWidth(pAlgorithm, pCluster, m_slidingShowerFitWindow));
+    float ratio(-1.f);
+    try
+    {
+        const TwoDSlidingFitResult slidingFitResultLarge(pCluster, m_slidingLinearFitWindow, LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+        const float straightLineLength = (slidingFitResultLarge.GetGlobalMaxLayerPosition() - slidingFitResultLarge.GetGlobalMinLayerPosition()).GetMagnitude();
+        if (straightLineLength > std::numeric_limits<double>::epsilon())
```
PandoraPFA: `straightLineLength` is a single float, so probably should compare to `std::numeric_limits<float>::epsilon()`

Most of the other comments below are highlighting this same point :)

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -95,27 +90,26 @@ const pandora::Cluster * const pCluster)
...
-
-    if (m_addDiffWithStraightLineSigma)
-        featureVector.push_back(diffWithStraightLineSigma);
-
-    if (m_addDTDLWidth)
-        featureVector.push_back(dTdLWidth);
-
-    if (m_addMaxFitGapLength)
-        featureVector.push_back(maxFitGapLength);
+    if (straightLineLengthLarge > std::numeric_limits<double>::epsilon())
```
PandoraPFA: `double` -> `float`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -136,17 +130,18 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
 
         for (const auto &mapEntry : slidingFitResult.GetLayerFitResultMap())
         {
-            rmsSlidingLinearFit += mapEntry.second.GetRms();
+            const LayerFitResult layerFitResult(mapEntry.second);
```
PandoraPFA: `const LayerFitResult &` to take a reference, rather than copy

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+{
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+        std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    float straightLineLength(-1.f), ratio(-1.f);
+    try
+    {
+        const TwoDSlidingFitResult slidingFitResultLarge(pCluster, m_slidingLinearFitWindow, LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+        straightLineLength = (slidingFitResultLarge.GetGlobalMaxLayerPosition() - slidingFitResultLarge.GetGlobalMinLayerPosition()).GetMagnitude();
+        if (straightLineLength > std::numeric_limits<double>::epsilon())
```
PandoraPFA: `double` -> `float`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
 
-        for (const Cluster *const pCandidateCluster : *pClusterList)
-        {
-            if ((LArClusterHelper::GetClusterHitType(pCluster)) != (LArClusterHelper::GetClusterHitType(pCandidateCluster)))
-                continue;
+        this->CalculateVariablesSlidingLinearFit(pCluster, straightLineLengthLargeCluster, diffWithStraightLineMeanCluster, maxFitGapLengthCluster, rmsSlidingLinearFitCluster);
 
-            if ((pCandidateCluster != pCluster) && (pCandidateCluster->GetNCaloHits() >= m_minClusterCaloHits))
-                candidateClusters.push_back(pCandidateCluster);
+        if (straightLineLengthLargeCluster > std::numeric_limits<double>::epsilon())
```
PandoraPFA: `double` -> `float`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
 
-    int nNearbyClusters(0);
+        diffWithStraightLineMean   += diffWithStraightLineMeanCluster;
+        maxFitGapLength            += maxFitGapLengthCluster;
+        rmsSlidingLinearFit        += rmsSlidingLinearFitCluster;
+    }
 
-    for (const Cluster *const pCandidateCluster : candidateClusters)
+    const float nClusters(static_cast<float>(clusterList.size()));
+    if (nClusters > std::numeric_limits<double>::epsilon())
```
PandoraPFA: Might be best to check that `clusterList` isn't empty (a concrete check). Inside `if` statement, can then cast `clusterList.size()` to a float and can then divide safely.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+        rmsSlidingLinearFit = 0.f;
+
+        FloatVector diffWithStraightLineVector;
+        const HitType hitType(LArClusterHelper::GetClusterHitType(pCluster));
+        CartesianVector previousFitPosition(slidingFitResult.GetGlobalMinLayerPosition());
+        float dTdLMin(+std::numeric_limits<float>::max()), dTdLMax(-std::numeric_limits<float>::max());
+
+        for (const auto &mapEntry : slidingFitResult.GetLayerFitResultMap())
+        {
+            const LayerFitResult layerFitResult(mapEntry.second);
```
PandoraPFA: Take a reference, as above

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+    const Vertex *const nuVertex(pVertexList->front());
+    //find the particle vertex
+   float vertexDistance(-1.f);
+    try
     {
-        for (const CaloHit *const pCaloHit : *layerIter.second)
-            mipEnergy += pCaloHit->GetMipEquivalentEnergy();
+        const Vertex *const pVertex = LArPfoHelper::GetVertex(pInputPfo);
+        const CartesianVector nuPosition(nuVertex->GetPosition()), pfoPosition(pVertex->GetPosition());
+        vertexDistance = std::sqrt((nuPosition - pfoPosition).GetMagnitudeSquared());
```
PandoraPFA: Can just call `GetMagnitude()` to avoid extra sqrt - might look cleaner/more standard.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+    LArPcaHelper::EigenVectors eigenVecsStart, eigenVecsEnd;
+    LArPcaHelper::EigenValues eigenValuesStart(0.f, 0.f, 0.f), eigenValuesEnd(0.f, 0.f, 0.f);
+
+    LArPcaHelper::RunPca(pointVectorStart, centroidStart, eigenValuesStart, eigenVecsStart);
+    LArPcaHelper::RunPca(pointVectorEnd, centroidEnd, eigenValuesEnd, eigenVecsEnd);
+
+    float openingAngle(this->OpeningAngle(eigenVecsStart.at(0), eigenVecsStart.at(1), eigenValuesStart));
+    float closingAngle(this->OpeningAngle(eigenVecsEnd.at(0), eigenVecsEnd.at(1), eigenValuesEnd));
+
+    featureVector.push_back(std::abs(openingAngle-closingAngle));
```
PandoraPFA: We use `std::fabs(...)` by default, although I think the old abs vs. fabs issues are no more.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+    this->Divide3DCaloHitList(pAlgorithm, threeDCaloHitList, pointVectorStart, pointVectorEnd);
+    // Run the PCA analysis twice
+    CartesianVector centroidStart(0.f, 0.f, 0.f), centroidEnd(0.f, 0.f, 0.f);
+    LArPcaHelper::EigenVectors eigenVecsStart, eigenVecsEnd;
+    LArPcaHelper::EigenValues eigenValuesStart(0.f, 0.f, 0.f), eigenValuesEnd(0.f, 0.f, 0.f);
+
+    LArPcaHelper::RunPca(pointVectorStart, centroidStart, eigenValuesStart, eigenVecsStart);
+    LArPcaHelper::RunPca(pointVectorEnd, centroidEnd, eigenValuesEnd, eigenVecsEnd);
+
+    float openingAngle(this->OpeningAngle(eigenVecsStart.at(0), eigenVecsStart.at(1), eigenValuesStart));
```
PandoraPFA: Can both be const floats

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+    float closingAngle(this->OpeningAngle(eigenVecsEnd.at(0), eigenVecsEnd.at(1), eigenValuesEnd));
+
+    featureVector.push_back(std::abs(openingAngle-closingAngle));
+
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ThreeDOpeningAngleFeatureTool::Divide3DCaloHitList(const Algorithm *const pAlgorithm, CaloHitList threeDCaloHitList, CartesianPointVector &pointVectorStart, CartesianPointVector &pointVectorEnd)
```
PandoraPFA: It looks like threeDCaloHitList should definitely be by const reference, rather than a potentially expensive copy.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------
+ThreeDChargeFeatureTool::ThreeDChargeFeatureTool() :
+    m_endChargeFraction(0.1f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+/*bool ThreeDChargeFeatureTool::SortByDistanceToVertex(const CaloHit *const left, const CaloHit *const right, const CartesianVector &nuVertex2D)
```
PandoraPFA: Can probably remove this now?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+    float totalCharge(-1.f), chargeSigma(-1.f), chargeMean(-1.f), endCharge(-1.f);
+    ClusterList pClusterList;
+    LArPfoHelper::GetClusters(pInputPfo, TPC_VIEW_W, pClusterList);
+    if ((!pClusterList.empty()) && (pClusterList.size() == 1))
+    {
+        const Cluster *const pCluster(pClusterList.front());
+        this->CalculateChargeVariables(pAlgorithm, pCluster, totalCharge, chargeSigma, chargeMean, endCharge, m_endChargeFraction);
+    }
+
+    if (chargeMean > std::numeric_limits<double>::epsilon())
```
PandoraPFA: `double` -> `float`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+        chargeSigma /= chargeMean;
+
+    if (totalCharge > std::numeric_limits<double>::epsilon())
```
PandoraPFA: `double` -> `float`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -199,147 +194,567 @@ void LinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandora::Clu
...
+
+    const CartesianVector vertexPosition2D(LArGeometryHelper::ProjectPosition(pAlgorithm->GetPandora(), pVertex->GetPosition(), hitType));
+
+    CaloHitList clusterCaloHitList;
+    pCluster->GetOrderedCaloHitList().FillCaloHitList(clusterCaloHitList);
+    CaloHitVector clusterCaloHitVector(clusterCaloHitList.begin(), clusterCaloHitList.end());
+
+    //TODO: might give problems if vertex in the middle of the cluster ?
+    std::sort(clusterCaloHitVector.begin(), clusterCaloHitVector.end(), VertexComparator(vertexPosition2D));
+    CaloHitList orderedCaloHitList(clusterCaloHitVector.begin(),clusterCaloHitVector.end());
```
PandoraPFA: `orderedCaloHitList` is an unnecessary intermediate.

larpandoracontent/LArTrackShowerId/PfoCharacterisationBaseAlgorithm.h
```diff
@@ -32,14 +32,23 @@ class PfoCharacterisationBaseAlgorithm : public pandora::Algorithm
 protected:
     pandora::StatusCode Run();
 
+    /**
+     *  @brief  Whether pfo is identified as a clear track using its three clsuters
```
PandoraPFA: `clsusters`

PandoraPFA:
Merged - please do as much testing as you can with the tip of all the master branches (LArContent, LArReco and MachineLearning data). Things look OK for me, but I have got the MonitoringRemaster feature folded into my default test setup. Thanks!

loressa:
Tested with master branch of all packages: SDK, LArContent, LArReco and MachineLearningData, and reproduced identical results.

PR #22, opened 2017-12-05
Added the LArFormattingHelper class

a-d-smith:
# LArFormattingHelper

A helper class to allow for modification of the colour / style of terminal output, and to draw tables. 

## Colour / style changes

Colour and style changes are made using ANSI escape sequences of the form:
```
\x1B[31m
```
- `\x1B` is the Hex ASCII escape character
- `[ & m` are the required characters to wrap the colour/style code
- `31` is an example colour/style code

When a user calls `SetColour` or `SetStyle`, the helper simply prints the ANSI sequence to the specified `ostream` (defaults to `std::cout`). A user must then use the `Reset` functions which print another ANSI sequence, resetting the formatting to default.

## Tables

LArFormattingHelper contains a `Table` class which can be used to quickly draw ASCII tables to the terminal. The table has three phases: definition, filling and printing.

### Table definition (construction)
The `Table` constructor takes the column headers as a vector of `std::string`. Users can use the special blank header `""` to draw a vertical rule. For example,
```c++
LArFormattingHelper::Table table(std::vector<std::string>({"Left", "", "Middle", "Right"}));
```
would produce a table with a header that looks like this
```
| Left || Middle | Right |
| ---- || ------ | ----- |
```
Users can also (optionally) specify, on construction, the precision to which numbers should be printed. This defaults to 3 significant figures.

### Table filling
The `Table` class flattens its data into vector. Users can only add elements in order (left to right, top to bottom). This is done using `AddElement`, which is templated to take any class that has a `<<` method defined.

Any `""` headers should be ignored when filling the table. For the above example, the table could be filled as follows:
```c++
table.AddElement(1);
table.AddElement("test");
table.AddElement(37.2);
```
This would produce the following output
```
| Left || Middle | Right |
| ---- || ------ | ----- |
|    1 ||   test |  37.2 |
```
Optionally, users can also specify a style and/or a colour for each table element added. 

### Printing
Once a table has been filled, it can be printed to the terminal using the `Print` method. Users must have added elements to fill rows entirely, otherwise an exception will be thrown. 

The tables *should* also be renderable as markdown (however tables are not part of the core Markdown spec, and some renderers will have trouble with the vertical rule).

### DIFF
--- a/larpandoracontent/LArHelpers/LArFormattingHelper.cc
+++ b/larpandoracontent/LArHelpers/LArFormattingHelper.cc
@@ -6,8 +6,6 @@
  *  $Log: $
  */
 
-#include "Pandora/StatusCodes.h"
-
 #include "larpandoracontent/LArHelpers/LArFormattingHelper.h"
 
 #include <iomanip>
--- a/larpandoracontent/LArHelpers/LArFormattingHelper.cc
+++ b/larpandoracontent/LArHelpers/LArFormattingHelper.cc
@@ -1,0 +1,184 @@
+/**
+ *  @file   larpandoracontent/LArHelpers/LArFormattingHelper.cc
+ *
+ *  @brief  Implementation of the formatting helper class.
+ *
+ *  $Log: $
+ */
+
+#include <iomanip>
+
+#include "Pandora/StatusCodes.h"
+
+#include "larpandoracontent/LArHelpers/LArFormattingHelper.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+    void LArFormattingHelper::SetStyle(const LArFormattingHelper::Style &style, std::ostream &stream)
+    {
+        LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(style), stream);
+    }
+   
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::SetColour(const LArFormattingHelper::Colour &colour, std::ostream &stream)
+    {
+        LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(colour), stream);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::ResetStyle(std::ostream &stream)
+    {
+        LArFormattingHelper::SetStyle(REGULAR, stream);
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::ResetColour(std::ostream &stream)
+    {
+        LArFormattingHelper::SetColour(DEFAULT, stream);
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::Reset(std::ostream &stream)
+    {
+        LArFormattingHelper::ResetColour(stream);
+        LArFormattingHelper::ResetStyle(stream);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::PrintFormatCharacter(const unsigned int &code, std::ostream &stream)
+    {
+        stream << LArFormattingHelper::GetFormatCharacter(code);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    std::string LArFormattingHelper::GetFormatCharacter(const unsigned int &code)
+    {
+        return ("\x1B[" + std::to_string(code) + "m");
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::PrintHeader(const std::string &title, const unsigned int &width)
+    {
+        LArFormattingHelper::SetStyle(BOLD);
+        std::cout << std::endl << std::string(width, '-') << "\r-" << title << std::endl;
+        LArFormattingHelper::Reset();
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::PrintRule(const unsigned int &width)
+    {
+        LArFormattingHelper::PrintHeader("", width);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    LArFormattingHelper::Table::Table(const std::vector<std::string> &columnTitles, const unsigned int &precision) :
+        m_columnTitles(columnTitles),
+        m_precision(precision)
+    {
+        m_stringstream.precision(m_precision);
+
+        for (const std::string &title : m_columnTitles)
+            m_widths.push_back(title.length());
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::CheckForSeparatorColumn()
+    {
+        if (m_columnTitles[m_elements.size() % m_columnTitles.size()] == "")
+        {
+            m_format.push_back("");
+            m_elements.push_back("");
+        }
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    bool LArFormattingHelper::Table::IsSeparatorColumn(const unsigned int &column)
+    {
+        if (column>=m_columnTitles.size())
+            throw StatusCodeException(STATUS_CODE_OUT_OF_RANGE);
+
+        return (m_columnTitles[column] == "");
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::UpdateColumnWidth()
+    {
+        const unsigned int currentElementIndex(m_elements.size()-1);
+        const unsigned int column(currentElementIndex % m_widths.size());
+        m_widths[column] = std::max(static_cast<unsigned int>(m_elements[currentElementIndex].length()), m_widths[column]);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+   
+    void LArFormattingHelper::Table::Print()
+    {
+        if (m_columnTitles.size() == 0)
+            return;
+
+        this->PrintColumnTitles();
+        this->PrintHorizontalLine();
+        this->PrintTableElements();
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintColumnTitles()
+    {
+        for (unsigned int i=0; i<m_columnTitles.size(); i++)
+            this->PrintTableCell(m_columnTitles[i], LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::BOLD), i);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintHorizontalLine()
+    {
+        for (unsigned int i=0; i<m_columnTitles.size(); i++) 
+            this->PrintTableCell(std::string(m_widths[i], '-'), LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::REGULAR), i);
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintTableElements()
+    {
+        const unsigned int nRows((m_elements.size() + m_columnTitles.size() - 1)/m_columnTitles.size());
+        if (nRows*m_columnTitles.size() != m_elements.size())
+        {
+            std::cerr << "LArFormattingHelper::Table::PrintTableElements - Error: Number of table elements added doesn't fill a whole row" << std::endl;
+            throw StatusCodeException(STATUS_CODE_OUT_OF_RANGE);
+        }
+
+        for (unsigned int i=0; i<nRows*m_columnTitles.size(); i++)
+            this->PrintTableCell(m_elements[i], m_format[i], i);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintTableCell(const std::string &value, const std::string &format, const unsigned int &index)
+    {
+        const unsigned int column(index % m_columnTitles.size());
+
+        std::string separator(this->IsSeparatorColumn(column) ? "" : " ");
+        std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator;
+        LArFormattingHelper::ResetStyle();
+
+        if (column == m_columnTitles.size()-1)
+            std::cout << "|" << std::endl;
+    }
+
+} // namespace lar_content
--- a/larpandoracontent/LArHelpers/LArFormattingHelper.cc
+++ b/larpandoracontent/LArHelpers/LArFormattingHelper.cc
@@ -61,7 +61,9 @@ void LArFormattingHelper::PrintFormatCharacter(const unsigned int code, std::ost
 
 std::string LArFormattingHelper::GetFormatCharacter(const unsigned int code)
 {
-    return ("\x1B[" + std::to_string(code) + "m");
+    const std::string startFormattingCode("\x1B[");
+    const std::string endFormattingCode("m");
+    return (startFormattingCode + std::to_string(code) + endFormattingCode);
 }
 
 //--------------------------------------------------------------------------------------------------------------------------------------
--- a/larpandoracontent/LArHelpers/LArFormattingHelper.cc
+++ b/larpandoracontent/LArHelpers/LArFormattingHelper.cc
@@ -120,6 +120,9 @@ bool LArFormattingHelper::Table::IsSeparatorColumn(const unsigned int column) co
 
 void LArFormattingHelper::Table::UpdateColumnWidth()
 {
+    if (m_widths.empty() || m_elements.empty())
+        throw StatusCodeException(STATUS_CODE_OUT_OF_RANGE);
+
     const unsigned int currentElementIndex(m_elements.size() - 1);
     const unsigned int column(currentElementIndex % m_widths.size());
     m_widths[column] = std::max(static_cast<unsigned int>(m_elements[currentElementIndex].length()), m_widths[column]);
@@ -141,8 +144,7 @@ void LArFormattingHelper::Table::Print() const
 
 void LArFormattingHelper::Table::PrintColumnTitles() const
 {
-    const unsigned int nColumns(m_columnTitles.size());
-    for (unsigned int i = 0; i < nColumns; ++i)
+    for (unsigned int i = 0, nColumns = m_columnTitles.size(); i < nColumns; ++i)
         this->PrintTableCell(m_columnTitles[i], LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::BOLD), i);
 }
 
@@ -150,8 +152,7 @@ void LArFormattingHelper::Table::PrintColumnTitles() const
 
 void LArFormattingHelper::Table::PrintHorizontalLine() const
 {
-    const unsigned int nColumns(m_columnTitles.size());
-    for (unsigned int i = 0; i < nColumns; ++i) 
+    for (unsigned int i = 0, nColumns = m_columnTitles.size(); i < nColumns; ++i)
         this->PrintTableCell(std::string(m_widths[i], '-'), LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::REGULAR), i);
 }
 
@@ -181,7 +182,7 @@ void LArFormattingHelper::Table::PrintTableCell(const std::string &value, const 
     if (m_columnTitles.empty())
         return;
 
-    const unsigned int column(index % m_columnTitles.size());
+    const unsigned int column(index % m_widths.size());
 
     const std::string separator(this->IsSeparatorColumn(column) ? "" : " ");
     std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator;
--- a/larpandoracontent/LArHelpers/LArFormattingHelper.cc
+++ b/larpandoracontent/LArHelpers/LArFormattingHelper.cc
@@ -1,184 +1,194 @@
 /**
  *  @file   larpandoracontent/LArHelpers/LArFormattingHelper.cc
  *
- *  @brief  Implementation of the formatting helper class.
+ *  @brief  Implementation of the lar formatting helper class.
  *
  *  $Log: $
  */
 
-#include <iomanip>
-
 #include "Pandora/StatusCodes.h"
 
 #include "larpandoracontent/LArHelpers/LArFormattingHelper.h"
+
+#include <iomanip>
 
 using namespace pandora;
 
 namespace lar_content
 {
 
-    void LArFormattingHelper::SetStyle(const LArFormattingHelper::Style &style, std::ostream &stream)
-    {
-        LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(style), stream);
-    }
-   
-    //--------------------------------------------------------------------------------------------------------------------------------------
+void LArFormattingHelper::SetStyle(const LArFormattingHelper::Style style, std::ostream &stream)
+{
+    LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(style), stream);
+}
 
-    void LArFormattingHelper::SetColour(const LArFormattingHelper::Colour &colour, std::ostream &stream)
-    {
-        LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(colour), stream);
-    }
-    
-    //--------------------------------------------------------------------------------------------------------------------------------------
+//--------------------------------------------------------------------------------------------------------------------------------------
 
-    void LArFormattingHelper::ResetStyle(std::ostream &stream)
+void LArFormattingHelper::SetColor(const LArFormattingHelper::Color color, std::ostream &stream)
+{
+    LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(color), stream);
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+void LArFormattingHelper::ResetStyle(std::ostream &stream)
+{
+    LArFormattingHelper::SetStyle(REGULAR, stream);
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+void LArFormattingHelper::ResetColor(std::ostream &stream)
+{
+    LArFormattingHelper::SetColor(DEFAULT, stream);
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+void LArFormattingHelper::Reset(std::ostream &stream)
+{
+    LArFormattingHelper::ResetColor(stream);
+    LArFormattingHelper::ResetStyle(stream);
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+void LArFormattingHelper::PrintFormatCharacter(const unsigned int code, std::ostream &stream)
+{
+    if (!(stream << LArFormattingHelper::GetFormatCharacter(code)))
+        throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+std::string LArFormattingHelper::GetFormatCharacter(const unsigned int code)
+{
+    return ("\x1B[" + std::to_string(code) + "m");
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+void LArFormattingHelper::PrintHeader(const std::string &title, const unsigned int width)
+{
+    LArFormattingHelper::SetStyle(BOLD);
+    std::cout << std::endl << std::string(width, '-') << "\r-" << title << std::endl;
+    LArFormattingHelper::Reset();
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+
+void LArFormattingHelper::PrintRule(const unsigned int width)
+{
+    LArFormattingHelper::PrintHeader("", width);
+}
+
+//--------------------------------------------------------------------------------------------------------------------------------------
+//--------------------------------------------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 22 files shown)


a-d-smith:
I just remembered... 

I forgot to Americanize: colour --> color!

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
+/**
+ *  @file   larpandoracontent/LArHelpers/LArFormattingHelper.cc
+ *
+ *  @brief  Implementation of the formatting helper class.
+ *
+ *  $Log: $
+ */
+
+#include <iomanip>
+
```
PandoraPFA: Normally put "standard" includes at the bottom of the list: purely a stylistic convention, trying to group includes in a sensible/useful/expected manner.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
+/**
+ *  @file   larpandoracontent/LArHelpers/LArFormattingHelper.cc
+ *
+ *  @brief  Implementation of the formatting helper class.
+ *
```
PandoraPFA: Of the *lar* formatting helper class.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+#include "Pandora/StatusCodes.h"
+
+#include "larpandoracontent/LArHelpers/LArFormattingHelper.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+    void LArFormattingHelper::SetStyle(const LArFormattingHelper::Style &style, std::ostream &stream)
```
PandoraPFA: Remove indentation here; functions start flush left.

PandoraPFA: If style is just an unsigned int, is there any point passing it by reference; just const by value should be fine.

On a related note, is style really an unsigned int or a signed int? I think you tell the compiler it's a signed int, but it's then used an unsigned throughout.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    {
+        LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(style), stream);
+    }
+   
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::SetColour(const LArFormattingHelper::Colour &colour, std::ostream &stream)
+    {
```
PandoraPFA: See comments for Style argument by reference vs. value, above.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        LArFormattingHelper::PrintFormatCharacter(static_cast<unsigned int>(colour), stream);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::ResetStyle(std::ostream &stream)
+    {
+        LArFormattingHelper::SetStyle(REGULAR, stream);
```
PandoraPFA: Resetting the style is interesting... This is probably just fine for all our needs, but it's potentially possible for a user with very non-standard style setup to call the LArFormattingHelper. The reset will presumably then reset everything to its own standard style, rather than the input style? Having said that, it would be ugly to have to cache the arguments upon initial usage... what if multiple callers, etc... Much cleaner in this approach.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+
+    void LArFormattingHelper::Reset(std::ostream &stream)
+    {
+        LArFormattingHelper::ResetColour(stream);
+        LArFormattingHelper::ResetStyle(stream);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    void LArFormattingHelper::PrintFormatCharacter(const unsigned int &code, std::ostream &stream)
```
PandoraPFA: No need for unsigned by const reference here.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    {
+        stream << LArFormattingHelper::GetFormatCharacter(code);
```
PandoraPFA: Can this operation fail? Doesn't the operator return something you can check, then raise an exception?

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    std::string LArFormattingHelper::GetFormatCharacter(const unsigned int &code)
+    {
+        return ("\x1B[" + std::to_string(code) + "m");
```
PandoraPFA: Is there a way to avoid the literals, e.g. can they be accessed from a standard header instead?

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+
+    void LArFormattingHelper::PrintRule(const unsigned int &width)
+    {
+        LArFormattingHelper::PrintHeader("", width);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    LArFormattingHelper::Table::Table(const std::vector<std::string> &columnTitles, const unsigned int &precision) :
```
PandoraPFA: Might want to use existing pandora::StringVector typedef?
unsigned by const reference again

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    {
+        m_stringstream.precision(m_precision);
+
+        for (const std::string &title : m_columnTitles)
+            m_widths.push_back(title.length());
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::CheckForSeparatorColumn()
```
PandoraPFA: Tentative: perhaps unexpected that this will be a non-const function, given the function name? CheckAndSet... instead?

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+            m_format.push_back("");
+            m_elements.push_back("");
+        }
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    bool LArFormattingHelper::Table::IsSeparatorColumn(const unsigned int &column)
+    {
+        if (column>=m_columnTitles.size())
```
PandoraPFA: Prefer spaces either side of >=

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+            throw StatusCodeException(STATUS_CODE_OUT_OF_RANGE);
+
+        return (m_columnTitles[column] == "");
```
PandoraPFA: Again non const, but name makes it sound like it should be a const member function.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::UpdateColumnWidth()
+    {
+        const unsigned int currentElementIndex(m_elements.size()-1);
```
PandoraPFA: spaces either side of minus sign(!)

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        const unsigned int column(currentElementIndex % m_widths.size());
+        m_widths[column] = std::max(static_cast<unsigned int>(m_elements[currentElementIndex].length()), m_widths[column]);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+   
+    void LArFormattingHelper::Table::Print()
+    {
+        if (m_columnTitles.size() == 0)
```
PandoraPFA: Prefer ```if (m_columnTitles.empty())```

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        this->PrintColumnTitles();
+        this->PrintHorizontalLine();
+        this->PrintTableElements();
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintColumnTitles()
+    {
+        for (unsigned int i=0; i<m_columnTitles.size(); i++)
```
PandoraPFA: Prefer ```i = 0```, ```i < m_``` and ```++i```

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+            this->PrintTableCell(m_columnTitles[i], LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::BOLD), i);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintHorizontalLine()
+    {
+        for (unsigned int i=0; i<m_columnTitles.size(); i++)
```
PandoraPFA: See comments above. Note that checking size of m_columnTitles repeatedly here, but this might be safest. This isn't likely to be performance-critical code.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    {
+        for (unsigned int i=0; i<m_columnTitles.size(); i++) 
+            this->PrintTableCell(std::string(m_widths[i], '-'), LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::REGULAR), i);
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintTableElements()
+    {
+        const unsigned int nRows((m_elements.size() + m_columnTitles.size() - 1)/m_columnTitles.size());
```
PandoraPFA: Need to check that m_columnTitles isn't empty somewhere, preferably right here, for safety.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        if (nRows*m_columnTitles.size() != m_elements.size())
+        {
+            std::cerr << "LArFormattingHelper::Table::PrintTableElements - Error: Number of table elements added doesn't fill a whole row" << std::endl;
```
PandoraPFA: We typically only write to std::cout in Pandora (but again convention...)

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        for (unsigned int i=0; i<m_columnTitles.size(); i++) 
+            this->PrintTableCell(std::string(m_widths[i], '-'), LArFormattingHelper::GetFormatCharacter(LArFormattingHelper::REGULAR), i);
+    }
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintTableElements()
+    {
+        const unsigned int nRows((m_elements.size() + m_columnTitles.size() - 1)/m_columnTitles.size());
+        if (nRows*m_columnTitles.size() != m_elements.size())
```
PandoraPFA: ```nRows * m_```

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        {
+            std::cerr << "LArFormattingHelper::Table::PrintTableElements - Error: Number of table elements added doesn't fill a whole row" << std::endl;
+            throw StatusCodeException(STATUS_CODE_OUT_OF_RANGE);
+        }
+
+        for (unsigned int i=0; i<nRows*m_columnTitles.size(); i++)
```
PandoraPFA: See comments above.

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+            this->PrintTableCell(m_elements[i], m_format[i], i);
+    }
+    
+    //--------------------------------------------------------------------------------------------------------------------------------------
+    
+    void LArFormattingHelper::Table::PrintTableCell(const std::string &value, const std::string &format, const unsigned int &index)
```
PandoraPFA: unsigned int by const reference

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+    {
+        const unsigned int column(index % m_columnTitles.size());
+
+        std::string separator(this->IsSeparatorColumn(column) ? "" : " ");
```
PandoraPFA: const std::string?

larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -0,0 +1,184 @@
...
+        std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator;
+        LArFormattingHelper::ResetStyle();
+
+        if (column == m_columnTitles.size()-1)
```
PandoraPFA: Space either side of minus sign; what happens if m_columnTitles is empty?

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+
+/**
+ *  @brief  LArFormattingHelper class
+ */
+class LArFormattingHelper
+{
+public:
+
+    enum Colour : int;
+    enum Style : int;
```
PandoraPFA: unsigned integers? Forward declaration for stylistic reasons? (would agree with this)

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+namespace lar_content
+{
+
+/**
+ *  @brief  LArFormattingHelper class
+ */
+class LArFormattingHelper
+{
+public:
+
```
PandoraPFA: No need for space

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+    enum Colour : int;
+    enum Style : int;
+
+    /**
+     *  @brief  Set the format style (to standard output stream)
+     *
+     *  @param  style the style of choice
+     */
+    static void SetStyle(const Style &style, std::ostream &stream=std::cout);
```
PandoraPFA: space either side of = throughout.

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+    static void PrintHeader(const std::string &title = "", const unsigned int &width = 140);
+
+    /**
+     *  @brief  Print a horizontal rule
+     *
+     *  @param  width the width of the rule line
+     */
+    static void PrintRule(const unsigned int &width = 140);
+
+    enum Style : int
```
PandoraPFA: doxygen: @brief, etc.

Where did all the magic numbers come from and can they be accessed via a nice header file somewhere?

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+    {
+        REGULAR = 0,
+        BOLD = 1,
+        DIM = 2,
+        UNDERLINED = 4,
+        INVERTED = 7
+    };
+    
+    enum Colour : int
```
PandoraPFA: See comments for Style enumeration, above.

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+        LIGHT_YELLOW = 93,
+        LIGHT_BLUE = 94,
+        LIGHT_MAGENTA = 95,
+        LIGHT_CYAN = 96,
+        WHITE = 97
+    };
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    class Table
```
PandoraPFA: Doxygen.

In Pandora code, we typically tend to add nested classes at "top" of parent class description.

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+        LIGHT_CYAN = 96,
+        WHITE = 97
+    };
+
+    //--------------------------------------------------------------------------------------------------------------------------------------
+
+    class Table 
+    {
+    public:
+
```
PandoraPFA: Remove space

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+        /**
+         * @brief  Table constructor
+         *
+         * @param  columnTitles the string columns titles use empty string for a separator column
+         * @param  precision the number of significant figures to display for number type table elements
+         */
+        Table(const std::vector<std::string> &columnTitles, const unsigned int &precision=3);
```
PandoraPFA: Maybe use existing pandora::StringVector typedef? precision by const value, rather than reference.

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+
+        /**
+         * @brief  Add an element to the table into the next (non-separator) column
+         *
+         * @param  value the element to add to the table
+         * @param  style the style of the element
+         * @param  colour the colour of the element
+         */
+        template<class T>
```
PandoraPFA: typically use typename, rather than class

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+        void AddElement(const T &value, const Style &style=REGULAR, const Colour &colour=DEFAULT);
```
PandoraPFA: Style and Colour by referenc

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+
+        /**
+         * @brief  Print the table
+         */
+        void Print();
+    private:
```
PandoraPFA: Add blank line here, between Print(); and private:

Can Print be a const member function?

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+        const std::vector<std::string>  m_columnTitles;     ///< The vector of columns titles in the table
```
PandoraPFA: Alignment of member variables.
Can use pandora::StringVector and pandora::UIntVector

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+         *  @param  widths a vector of the number of characters to include in each column 
+         */
+        void PrintColumnTitles();
+
+        /**
+         *  @brief  Print a horizontal line
+         *
+         *  @param  widths a vector of the number of characters to include in each column 
+         */
+        void PrintHorizontalLine();
```
PandoraPFA: Check to see which member functions can be const. In class description, typically order by:
1. public daughter classes
2. public and const member functions (and public constructor, destructor)
3. public and non-const member functions
4. public member variables (often deprecated)
5. private daughter classes
6. private and const member functions
7. private and non-const member functions
8. private member variables

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+         *  @param  value the string to print in the cell
+         *  @param  format a formatting string (see GetFormatCharacter(...))
+         *  @param  index the index of the table cell (0 --> number of elements) used to find the column
+         *  @param  widths a vector of the number of characters to include in each column 
+         */
+        void PrintTableCell(const std::string &value, const std::string &format, const unsigned int &index);
+    };
+    
+};
+
```
PandoraPFA: Separator lines between class description and inline implementation.

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+template<class T>
+inline void LArFormattingHelper::Table::AddElement(const T &value, const Style &style, const Colour &colour)
+{
+    this->CheckForSeparatorColumn();
+
+    m_stringstream << value;
```
PandoraPFA: Check if this succeeds, if possible.

larpandoracontent/LArHelpers/LArFormattingHelper.h
```diff
@@ -0,0 +1,222 @@
...
+         *  @param  format a formatting string (see GetFormatCharacter(...))
+         *  @param  index the index of the table cell (0 --> number of elements) used to find the column
+         *  @param  widths a vector of the number of characters to include in each column 
+         */
+        void PrintTableCell(const std::string &value, const std::string &format, const unsigned int &index);
+    };
+    
+};
+
+template<class T>
```
PandoraPFA: class vs. typename, as above

larpandoracontent/LArHelpers/LArFormattingHelper.cc
a-d-smith: This is a good point. To the best of my knowledge, this will reset back to the default colour / style used in the terminal preferences, but I haven't tested this. 

Maybe we could add a switch to allow users to turn off styles / colours globally? I am thinking something like IsMonitoringEnabled. I guess this isn't the nicest solution but if it was annoying for someone with a non-standard setup it would be good to be able to turn it off.

PandoraPFA: If this was in the SDK, we absolutely would need to address this head on, but it's in the algorithm library, for at monitoring for *development* purposes. As such, we can probably justify keeping things simple to cover 99.9+% of use cases!

larpandoracontent/LArHelpers/LArFormattingHelper.cc
a-d-smith: I can check if stream::fail() returns true!

larpandoracontent/LArHelpers/LArFormattingHelper.cc
a-d-smith: I did some googling, but I can't find any standard header. I don't know for sure though!

PandoraPFA: With a quick bit of googling, I did spot the following:

>These are ANSI escape sequences (also known as VT100 codes) are an early standardisation of control codes pre-dating ASCII.
>
>The escape sequence \x1BE, or Esc+E, is NEL or "Next line", and is used on older terminals and mainframes to denote CR+LF, or \r\n.

```\r\n``` is definitely more recognisable, but still a hard-coded string literal!

File: larpandoracontent/LArHelpers/LArFormattingHelper.cc (hunk 10 above)
Comment (a-d-smith) [reply larpandoracontent/LArHelpers/LArFormattingHelper.cc]: Renamed to CheckAndSetSeparatorColumn, and made other member functions const to be explicit

File: larpandoracontent/LArHelpers/LArFormattingHelper.cc (hunk 8 above)
Comment (PandoraPFA) [reply larpandoracontent/LArHelpers/LArFormattingHelper.cc]: Could just go with something like the following example, from [msdn](https://msdn.microsoft.com/en-us/library/69ze775t.aspx)?

```
    char newline = '\n';  
    char tab = '\t';  
    char backspace = '\b';  
    char backslash = '\\';  
    char nullChar = '\0';  
  
    cout << "Newline character: " << newline << "ending" << endl; // Newline character:  
                                                                  //  ending  
    cout << "Tab character: " << tab << "ending" << endl; // Tab character : ending  
    cout << "Backspace character: " << backspace << "ending" << endl; // Backspace character : ending  
    cout << "Backslash character: " << backslash << "ending" << endl; // Backslash character : \ending  
    cout << "Null character: " << nullChar << "ending" << endl; //Null character:  ending  
```

File: larpandoracontent/LArHelpers/LArFormattingHelper.cc (hunk 16 above)
Comment (a-d-smith) [reply larpandoracontent/LArHelpers/LArFormattingHelper.cc]: m_columnTitles is const, so replaced with

     const unsigned int nColumns(m_columnTitles.size());                                                                                      
     for (unsigned int i = 0; i < nColumns; ++i)

File: larpandoracontent/LArHelpers/LArFormattingHelper.cc (hunk 8 above)
Comment (a-d-smith) [reply larpandoracontent/LArHelpers/LArFormattingHelper.cc]: I think `\x1BE = ESC+E` (for next line as you say `\r\n`), but in this case we want `\x1B[ = ESC+[` which is the "control sequence introducer" for changing styles / colours.

I can make const static variables for the control characters, however, they are only used in one place. What do you think?

File: larpandoracontent/LArHelpers/LArFormattingHelper.cc (hunk 16 above)
Comment (a-d-smith) [reply larpandoracontent/LArHelpers/LArFormattingHelper.cc]: `m_columnTitles` is `const`, so replaced with
```c++
const unsigned int nColumns(m_columnTitles.size());                                                                                      
for (unsigned int i = 0; i < nColumns; ++i)  
```

larpandoracontent/LArHelpers/LArFormattingHelper.cc
a-d-smith: Added check for empty `m_columnTitles` for safety, although this function is only called by functions if `m_columnTitles` is not empty

larpandoracontent/LArHelpers/LArFormattingHelper.h
a-d-smith: Yep! they should be unsigned

larpandoracontent/LArHelpers/LArFormattingHelper.h
a-d-smith: I can't actually find any standard headers, and I think that's because the definitions are slightly different across platforms. From what I understand this seems to be mainly windows vs. non-windows, which isn't a problem for us! But it would be good if we could test on mac.

I did find a nifty snippet that we could use to check if a users terminal supports colours, by reading the `TERM` environment variable. 
```c++
bool supportsColor()
{
    if(const char *env_p = std::getenv("TERM")) {
        const char *const terms[] = {
            "xterm", "xterm-256", "xterm-256color", "vt100",
            "color", "ansi",      "cygwin",         "linux"};
        // Change 1
        for(auto const term: terms) {
            if(std::strcmp(env_p, term) == 0) return true;
        }
    }
    return false;
}
```
This bool would only need to used in `GetFormatCharacter`, which could always return `""` if colour isn't supported. What do you think?

larpandoracontent/LArHelpers/LArFormattingHelper.h
a-d-smith: Added const where possible, and modified order

a-d-smith:
Moving away from just the formatting helper to include updates to the validation.

larpandoracontent/LArHelpers/LArMonitoringHelper.cc
```diff
@@ -308,4 +313,224 @@ unsigned int LArMonitoringHelper::CountHitsByType(const HitType hitType, const C
...
+                // ATTN default to normal MCParticle sorting to avoid tie-breakers
+                return (a.first < b.first);
+            });
+
+            for (const MCParticleCaloHitListPair &mcParticleCaloHitPair : mcParticleToGoodHitsVect)
+                orderedMCParticleVector.push_back(mcParticleCaloHitPair.first);
+        }
+        catch (const StatusCodeException &exception)
+        {
+            throw StatusCodeException(exception);
```
PandoraPFA: Maybe just don't catch, if you're simply going to rethrow?

larpandoracontent/LArHelpers/LArMonitoringHelper.cc
```diff
@@ -308,4 +313,224 @@ unsigned int LArMonitoringHelper::CountHitsByType(const HitType hitType, const C
     return nHitsOfSpecifiedType;
 }
 
+//------------------------------------------------------------------------------------------------------------------------------------------
+    
+void LArMonitoringHelper::GetOrderedMCParticleVector(const std::vector<MCContributionMap> &selectedMCParticleToGoodHitsMaps, MCParticleVector &orderedMCParticleVector)
+{
+    for (const MCContributionMap &mcParticleToGoodHitsMap : selectedMCParticleToGoodHitsMaps)
```
PandoraPFA: This (and the next function) seems rather verbose for what it does...

larpandoracontent/LArMonitoring/PfoValidationAlgorithm.cc
```diff
@@ -0,0 +1,120 @@
...
+
+#include "larpandoracontent/LArMonitoring/PfoValidationAlgorithm.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+PfoValidationAlgorithm::PfoValidationAlgorithm()
+{
```
PandoraPFA: Might have to work on the name in future, but fine for now!

PandoraPFA:
Development now moved to `feature/MonitoringRemaster` branch.

PR #26, opened 2018-02-19
Neutrino ID

a-d-smith:
This is the "first" iteration of the neutrino ID. I realise the SliceFeatures constructor is a little messy, I can make it more readable but it might be less efficient. I thought I would just get something in the bag so we have a starting point!


larpandoracontent/LArCheating/CheatingNeutrinoIdTool.cc
```diff
@@ -19,8 +19,10 @@ using namespace pandora;
 namespace lar_content
 {
 
-void CheatingNeutrinoIdTool::SelectOutputPfos(const SliceHypotheses &nuSliceHypotheses, const SliceHypotheses &crSliceHypotheses, PfoList &selectedPfos)
+void CheatingNeutrinoIdTool::SelectOutputPfos(const MasterAlgorithm *const pAlgorithm, const SliceHypotheses &nuSliceHypotheses, const SliceHypotheses &crSliceHypotheses, PfoList &selectedPfos)
```
PandoraPFA: Presumably no need for this to be a MasterAlgorithm instance, rather than an Algorithm instance? Maybe ultimately don't need to take an Algorithm at all, but probably OK for first iteration. (And similar, below.)

larpandoracontent/LArCheating/CheatingNeutrinoIdTool.cc
```diff
@@ -19,8 +19,10 @@ using namespace pandora;
...
 {
+    (void) pAlgorithm;
```
PandoraPFA: Just remove (comment out better) algorithm variable, rather than add a null line to implementation. (And similar, below.)

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -10,11 +10,22 @@
...
 
 /**
  *  @brief  NeutrinoIdTool class
+ *
+ *  Compares the neutrino and cosmic hypotheses of all of the slices in the event. Uses an SVM to calculate the probability of each slice 
+ *  containing a neutrino interaction. The N slices with the highest probabilities are identified as a neutrino (if sufficiently probable)
+ *  all other slices are deemed cosmogenic.
+ *
+ *  If training mode is switched on, then the tool will write SVM training exmples to the specified output file. The events selected for 
+ *  training must pass (user congigurable) slicing quality cuts. Users may also select events based on their interaction type (nuance code).
```
PandoraPFA: Typo: "congigurable"

Is this all correctly formatted to be picked up by doxygen? It probably is, although I normally add n sets of four spaces (no tab characters) to indent so that the left text margin is in line the NeutrinoIdTool class. Some people add html commands, but we have never done this in Pandora.

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
      */
     NeutrinoIdTool();
 
-    void SelectOutputPfos(const SliceHypotheses &nuSliceHypotheses, const SliceHypotheses &crSliceHypotheses, pandora::PfoList &selectedPfos);
+    void SelectOutputPfos(const MasterAlgorithm *const pAlgorithm, const SliceHypotheses &nuSliceHypotheses, const SliceHypotheses &crSliceHypotheses, pandora::PfoList &selectedPfos);
 
 private:
+    class SliceFeatures;
```
PandoraPFA: Probably don't need forward declaration here; can just move typedef under the full class definition.

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
...
+
+        /**
+         *  @brief  Check if all features were calculable
+         */
+        bool IsFeatureVectorAvailable() const;
+
+        /**
+         *  @brief  Get the feature vector for the SVM
+         */
+        SupportVectorMachine::DoubleVector GetFeatureVector() const;
```
PandoraPFA: Check return by value vs. const reference (always preferable, if possible).

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
...
+        
+        /**
+         *  @brief  Get the probability that this slice contains a neutrino interaction
+         *
+         *  @param  supportVectorMachine the SVM used to calculate the probability
+         */
+        float GetNeutrinoProbability(const SupportVectorMachine &supportVectorMachine) const;
+    private:
```
PandoraPFA: Add empty line before private keyword.

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
...
+         *
+         *  @param  nuPfos input list of Pfos reconstructed under the neutrino hypothesis
+         *  @param  crPfos input list of Pfos reconstructed under the cosmic ray hypothesis
+         *  @param  pTool address of the tool using this class
+         */
+        SliceFeatures(const pandora::PfoList &nuPfos, const pandora::PfoList &crPfos, const NeutrinoIdTool *const pTool);
+
+        /**
+         *  @brief  Check if all features were calculable
+         */
```
PandoraPFA: Ought to add e.g. `@return    boolean`, and provide details for all other functions returning anything other than void, although you will find this hasn't been done comprehensively throughout our entire code base.

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
...
+        const pandora::ParticleFlowObject *GetNeutrino(const pandora::PfoList &nuPfos);
+
+        /**
+         *  @brief  Get the 3D space points in a given pfo
+         *
+         *  @param  pPfo input pfo
+         *  @param  spacePoints vector to hold the 3D space points associated with the input pfo
+         */
+        void GetSpacePoints(const pandora::ParticleFlowObject *const pPfo, pandora::CartesianPointVector &spacePoints) const;
```
PandoraPFA: Is there a function to do this in LArPfoHelper?

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
...
+     *  @param  hitList output list of all 2d hits in the input pfos
+     */
+    void Collect2DHits(const pandora::PfoList &pfos, pandora::CaloHitList &hitList) const;
+
+    /**
+     *  @brief  Count the number of neutrino induced hits in a given list
+     *
+     *  @param  hitList input list of calo hits
+     */
+    unsigned int CountNeutrinoInducedHits(const pandora::CaloHitList &hitList) const;
```
PandoraPFA: Ought to highlight whether this uses MC information or not.

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
```diff
@@ -24,13 +35,201 @@ class NeutrinoIdTool : public SliceIdBaseTool
...
+     *
+     *  @param  pfos the pfos to select
+     *  @param  selectedPfos the list of pfos to populate
+     */
+    void SelectPfos(const pandora::PfoList &pfos, pandora::PfoList &selectedPfos) const;
+
     pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
 
-    bool        m_selectAllNeutrinos;               ///< First approach: select all neutrinos, as opposed to selecting all cosmics
-    bool        m_selectOnlyFirstSliceNeutrinos;    ///< First approach: select first slice neutrinos, cosmics for all subsequent slices
```
PandoraPFA: Should probably maintain the old functionality, to be enabled via config. This may be by adding a simple tool (as we use in production right now), or, if you prefer, by maintaining these two Booleans in this tool.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+    SliceFeaturesVector sliceFeaturesVector;
+    this->GetSliceFeatures(this, nuSliceHypotheses, crSliceHypotheses, sliceFeaturesVector);
+
+    if (m_useTrainingMode)
+    {
+        // ATTN in training mode, just return everything as a cosmic-ray
+        this->SelectAllPfos(crSliceHypotheses, selectedPfos);
+
+        unsigned int bestSliceIndex(std::numeric_limits<unsigned int>::max());
+        if (!this->GetBestSliceIndex(pAlgorithm, nuSliceHypotheses, crSliceHypotheses, bestSliceIndex)) return;
```
PandoraPFA: Is this the MC truth part? If so, should make this clear via function name.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+{
+    for (unsigned int sliceIndex = 0, nSlices = nuSliceHypotheses.size(); sliceIndex < nSlices; ++sliceIndex)
+        sliceFeaturesVector.push_back(SliceFeatures(nuSliceHypotheses.at(sliceIndex), crSliceHypotheses.at(sliceIndex), pTool));
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool NeutrinoIdTool::GetBestSliceIndex(const MasterAlgorithm *const pAlgorithm, const SliceHypotheses &nuSliceHypotheses, const SliceHypotheses &crSliceHypotheses, unsigned int &bestSliceIndex) const
+{
+    unsigned int nHitsInBestSlice(0);
```
PandoraPFA: Might want to declare and initialise all three unsigned ints on one line. Have started to find this helpful when reading implementation (conceptually just doing the same thing three times over).

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
-        selectedPfos.insert(selectedPfos.end(), sliceOutput.begin(), sliceOutput.end());
+        CaloHitList reconstructedHits;
+        this->Collect2DHits(nuFinalStates, reconstructedHits);
+        this->Collect2DHits(crSliceHypotheses.at(sliceIndex), reconstructedHits);
+
+        const unsigned int nNuHits(this->CountNeutrinoInducedHits(reconstructedHits));
+        nuNHitsTotal += nNuHits;
+        
+        // TODO may need to be careful with reproducibility here?
+        if (nNuHits > nNuHitsInBestSlice)
```
PandoraPFA: Always good to look out for this. We should discuss if there's anything subtle going on. It looks like the input slices should always been in the same order, and then the input pfos in the same order... Is this right?

Otherwise need some sorting and/or tiebreakers.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+        {
+            nNuHitsInBestSlice = nNuHits;
+            nHitsInBestSlice = reconstructedHits.size();
+            bestSliceIndex = sliceIndex; 
+        }
+    }
+
+    // ATTN for events with no neutrino induced hits, default neutrino purity and completeness to zero
+    float purity(0);
```
PandoraPFA: Can use `? :` operator to collapse these down to two lines, and make them const floats. Usually set floats to e.g. `0.f` to make types explicit, rather than `0`

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+
+    return this->PassesQualityCuts(pAlgorithm, purity, completeness);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool NeutrinoIdTool::PassesQualityCuts(const MasterAlgorithm *const pAlgorithm, const float purity, const float completeness) const
+{
+    if (purity < m_minPurity || completeness < m_minCompleteness) return false;
+    if (m_selectNuanceCode && (this->GetNuanceCode(pAlgorithm) != m_nuance)) return false;
```
PandoraPFA: Nuance code use is actually under the PassesQualityCuts function? Maybe this is the best place, but would need to add either an ATTN, or additional comments in header.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+
+    return true;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void NeutrinoIdTool::Collect2DHits(const PfoList &pfos, CaloHitList &hitList) const
+{
```
PandoraPFA: OK, now I understand the need for a new Collect2DHits function :)

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+    LArPfoHelper::GetCaloHits(pfos, TPC_VIEW_V, collectedHits);
+    LArPfoHelper::GetCaloHits(pfos, TPC_VIEW_W, collectedHits);
+
+    for (const CaloHit *const pCaloHit : collectedHits)
+    {
+        // ATTN hits collected from Pfos are copies of hits passed from master instance, we need to access their parent to use MC info
+        const CaloHit *pParentCaloHit(static_cast<const CaloHit *>(pCaloHit->GetParentAddress()));
+
+        // Ensure no hits have been double counted
+        if (std::find(hitList.begin(), hitList.end(), pParentCaloHit) == hitList.end())
```
PandoraPFA: Can find this is incredibly slow, whilst checking a CaloHitSet (filled in parallel) is much faster. Depends on numbers of hits in use, though. We should profile again soon.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+{
+    const MCParticleList *pMCParticleList = nullptr;
+    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::GetCurrentList(*pAlgorithm, pMCParticleList));
+
+    MCParticleVector trueNeutrinos;
+    LArMCParticleHelper::GetTrueNeutrinos(pMCParticleList, trueNeutrinos);
+
+    if (trueNeutrinos.size() != 1)
+    {
+        std::cout << "NeutrinoIdTool::GetNuanceCode - Error: number of true neutrinos in event must be exactly one" << std::endl;
```
PandoraPFA: I did find a few events with multiple true neutrinos, with valid nuance codes.

First of all, is this OK with MCNeutrinos that don't have nuance codes - presumably LArMCParticleHelper doesn't pass these as true neutrinos?

Secondly, is everything OK if the exception is raised here? What happens, is the event just skipped from training?

Obvious question - no MC information is ever even collected for regular use case i.e. classification, right? :)

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+{
+    selectedPfos.insert(selectedPfos.end(), pfos.begin(), pfos.end());
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+// TODO think about how to make this function cleaner when features are more established
+NeutrinoIdTool::SliceFeatures::SliceFeatures(const PfoList &nuPfos, const PfoList &crPfos, const NeutrinoIdTool *const pTool) : 
+    m_isAvailable(false), m_pTool(pTool)
```
PandoraPFA: `m_pTool` on new line.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+{
+    const ParticleFlowObject *const pNeutrino(this->GetNeutrino(nuPfos));
+    const CartesianVector nuVertex(LArPfoHelper::GetVertex(pNeutrino)->GetPosition());
```
PandoraPFA: Can this function raise an exception?

Can this be a const reference?

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+    CartesianPointVector nuAllSpacePoints;
+    for (const ParticleFlowObject *const pPfo : nuFinalStates)
+    {
+        CartesianPointVector spacePoints;
+        this->GetSpacePoints(pPfo, spacePoints);
+
+        nuAllSpacePoints.insert(nuAllSpacePoints.end(), spacePoints.begin(), spacePoints.end());
+        nuNHitsTotal += spacePoints.size();
+
+        if (spacePoints.size() < 5) continue;
```
PandoraPFA: A few hardcoded numbers to (ultimately) replace via e.g. xml-configured parameters.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+
+        CartesianVector dir(this->GetDirectionFromVertex(spacePoints, nuVertex));
```
PandoraPFA: const?

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,464 @@
...
+        nuWeightedDirTotal += dir * static_cast<float>(spacePoints.size());
+        nuNHitsUsedTotal += spacePoints.size();
+    }
+
+    if (nuNHitsUsedTotal == 0) return;
+    CartesianVector nuWeightedDir(nuWeightedDirTotal * (1.f / static_cast<float>(nuNHitsUsedTotal)));
```
PandoraPFA: const?

larpandoracontent/LArControlFlow/NeutrinoIdTool.h
a-d-smith: Do you mean maintain these as options to bypass this neutrino ID tool? Maybe it makes most sense, as you say, to just keep the simple tool too.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: Yes I think you are right, provided the input slices are always given to the tool in the same order (which I believe they are) then this shouldn't be an issue. I just wanted to flag to to make sure!

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: The number of hits used will range from 10s to 1000s for the MicroBooNE case. Would this simply be a case of replacing CaloHitList -> CaloHitSet?

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: 1. Yes the LArMCParticleHelper insists that the neutrino has a nuance code
https://github.com/PandoraPFA/LArContent/blob/5a4156a0e7a0e7ddaad3d378a0a0a3f42204d5f7/larpandoracontent/LArHelpers/LArMCParticleHelper.cc#L83
2. You are right, no MC information is ever collected in the regular use case. If an exception is raised then the event will be skipped altogether from training. The idea is if you run this on a cosmic only sample in training mode it wouldn't produce any training samples at all.

PandoraPFA:
Hi Andy,

It’s more complicated than that, because a CaloHitSet is unordered. You’d need to work with the CaloHitList, but maintain a set in parallel for faster lookups of the hits in play. It would, obviously, only be for performance.

PandoraPFA:
Yes, let’s keep a simple neutrino id tool too (should call it simple, or a better, more descriptive name). Ensures we can maintain current functionality via config.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: You are right, good point, it can raise an exception. In fact, I want to bail out if the calculation of any feature isn't possible in any way, this could be due to numerical fails (e.g. all of the zero checks), or if *anything* raises and exception. I have now wrapped everything in a try block to ensure safety against this.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: I was wary about making these configurable parameters, because if they are changed then the SVM training could be thrown off. Maybe it is best for them to still be configurable nonetheless?

PandoraPFA:
feature/MonitoringRemaster has now been merged to master. Can you redirect the PR to master, and take a look at the CheatingNeutrinoIdTool.cc conflict. I think the version in feature/MonitoringRemaster should be the one to go with.

PR #27, opened 2018-02-21
feature/NeutrinoIdTool

a-d-smith:
I have rebased the code on top of the new master branch.


larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+{
+    unsigned int nNuHits(0);
+    for (const CaloHit *const pCaloHit : hitList)
+    {
+        try
+        {
+            if (LArMCParticleHelper::IsNeutrino(LArMCParticleHelper::GetParentMCParticle(MCParticleHelper::GetMainMCParticle(pCaloHit))))
+                nNuHits++;
+        }
+        catch (StatusCodeException &)
```
PandoraPFA: const StatusCodeException

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+// TODO think about how to make this function cleaner when features are more established
+NeutrinoIdTool::SliceFeatures::SliceFeatures(const PfoList &nuPfos, const PfoList &crPfos, const NeutrinoIdTool *const pTool) : 
+    m_isAvailable(false),
+    m_pTool(pTool)
+{
+    try
```
PandoraPFA: This is a very big try/catch block, so we'll need to watch out (as we develop this) how many different issues are being folded together to produce a SliceFeatures with isAvailable == false (which is at least safe).

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+        LArPcaHelper::EigenValues eigenValues(std::numeric_limits<float>::max(), std::numeric_limits<float>::max(), std::numeric_limits<float>::max());
+        LArPcaHelper::EigenVectors eigenVectors;
+        LArPcaHelper::RunPca(pointsInSphere, centroid, eigenValues, eigenVectors);
+
+
+        float nuNFinalStatePfos(static_cast<float>(nuFinalStates.size()));
+        float nuVertexY(nuVertex.GetY());
+        float nuWeightedDirZ(nuWeightedDir.GetZ());
+        float nuNSpacePointsInSphere(static_cast<float>(pointsInSphere.size()));
+        float nuEigenRatioInSphere(eigenValues.GetY() / eigenValues.GetX());
```
PandoraPFA: All const, plust check whether `eigenValues.GetX()` might be zero...

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+
+                crLongestTrackDirY = upperDir.GetY();
+                crLongestTrackDeflection = 1 - upperDir.GetDotProduct(lowerDir * (-1));
+            }
+        }
+
+        if (nCRHitsMax == 0) return;
+        if (nCRHitsTotal == 0) return;
+
+        float crFracHitsInLongestTrack = static_cast<float>(nCRHitsMax)/static_cast<float>(nCRHitsTotal);
```
PandoraPFA: const

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+    m_pTool(pTool)
+{
+    try
+    {
+        const ParticleFlowObject *const pNeutrino(this->GetNeutrino(nuPfos));
+        const CartesianVector &nuVertex(LArPfoHelper::GetVertex(pNeutrino)->GetPosition());
+        const PfoList &nuFinalStates(pNeutrino->GetDaughterPfoList());
+
+        // Neutrino features
+        CartesianVector nuWeightedDirTotal(0, 0, 0);
```
PandoraPFA: `(0.f, 0.f, 0.f)` preferred

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+            if (spacePoints.size() < 5) continue;
+            
+            if (spacePoints.size() > nCRHitsMax)
+            {
+                nCRHitsMax = spacePoints.size();
+                const CartesianVector upperDir(this->GetUpperDirection(spacePoints));
+                const CartesianVector lowerDir(this->GetLowerDirection(spacePoints));
+
+                crLongestTrackDirY = upperDir.GetY();
+                crLongestTrackDeflection = 1 - upperDir.GetDotProduct(lowerDir * (-1));
```
PandoraPFA: `1.f`, `-1.f`

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+bool NeutrinoIdTool::SliceFeatures::IsFeatureVectorAvailable() const
+{
+    return m_isAvailable;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+        
+void NeutrinoIdTool::SliceFeatures::GetFeatureVector(SupportVectorMachine::DoubleVector &featureVector) const
+{
+    if (!m_isAvailable)
```
PandoraPFA: Do we also need to check whether the feature vector is empty, or would that be OK?

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+
+void NeutrinoIdTool::SliceFeatures::GetSpacePoints(const ParticleFlowObject *const pPfo, CartesianPointVector &spacePoints) const
+{
+    ClusterList clusters3D;
+    LArPfoHelper::GetThreeDClusterList(pPfo, clusters3D);
+    
+    if (clusters3D.size() > 1)
+        throw StatusCodeException(STATUS_CODE_OUT_OF_RANGE);
+                                                                                                                          
+    if (clusters3D.size() == 0) return;
```
PandoraPFA: `clusters3D.empty()` is faster

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+    const CartesianVector dirMin(fit.GetGlobalMinLayerDirection());                                                                          
+    const CartesianVector dirMax(fit.GetGlobalMaxLayerDirection());                                                                          
+                                                                                                                                             
+    const bool isMinStart(fShouldChooseA(endMin, endMax));
+    const CartesianVector startPoint(isMinStart ? endMin : endMax);                                                                          
+    const CartesianVector endPoint(isMinStart ? endMax : endMin);                                                                            
+    const CartesianVector startDir(isMinStart ? dirMin : dirMax);                                                                            
+                                                                                                                                             
+    const bool shouldFlip((endPoint - startPoint).GetUnitVector().GetDotProduct(startDir) < 0);
```
PandoraPFA: `0.f`, and `-1.f` in the line below

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -10,30 +10,459 @@
...
+   
     for (unsigned int sliceIndex = 0, nSlices = nuSliceHypotheses.size(); sliceIndex < nSlices; ++sliceIndex)
     {
-        const PfoList &sliceOutput((m_selectAllNeutrinos || (m_selectOnlyFirstSliceNeutrinos && (0 == sliceIndex))) ?
-            nuSliceHypotheses.at(sliceIndex) : crSliceHypotheses.at(sliceIndex));
+        CaloHitList reconstructedHits;
+        this->Collect2DHits(crSliceHypotheses.at(sliceIndex), reconstructedHits);
 
-        selectedPfos.insert(selectedPfos.end(), sliceOutput.begin(), sliceOutput.end());
+        if (nuSliceHypotheses.at(sliceIndex).size() == 1)
```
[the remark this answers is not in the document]
a-d-smith: This is just trying to collect all and any hits that are in the slice, it doesn't really matter if there are one or more reconstructed neutrinos. So, I suppose I could actually loop over all reconstructed neutrinos?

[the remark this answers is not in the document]
PandoraPFA: Yes, let's just do that.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: Wow, I missed so many consts!

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
a-d-smith: The feature vectors all must have the same size and the order of the features within them must be the same. The idea is we will only get out of the SliceFeatures constructor with `m_isAvailable = true` if all zero check were passed an no exceptions were thrown. We then never touch `m_featureVector` again, all member functions are const. 

https://github.com/a-d-smith/LArContent/blob/8feae693de59e1f2d06b15f8a308fd65d55c54e0/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc#L326-L338

a-d-smith: The feature vectors all must have the same size and the order of the features within them must be the same. The idea is we will only get out of the SliceFeatures constructor with `m_isAvailable = true` if all zero check were passed an no exceptions were thrown. We then never touch `m_featureVector` again, all member functions are const. 

https://github.com/a-d-smith/LArContent/blob/8feae693de59e1f2d06b15f8a308fd65d55c54e0/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc#L326-L338

PR #30, opened 2018-03-06
Created cheating beam particle id tool.

StevenGreen1:


StevenGreen1:
The second commit includes a base tool for cheating the neutrino and beam particle id.  It was tested on ProtoDUNE events.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
```diff
@@ -0,0 +1,82 @@
...
+            LArPfoHelper::GetAllDownstreamPfos(pNeutrinoPfo, downstreamPfos);
+
+            float thisBeamParticleWeight(0.f), thisTotalWeight(0.f);
+            CheatingParticleIdBaseTool::GetTargetParticleWeight(&downstreamPfos, false, thisBeamParticleWeight, thisTotalWeight, LArMCParticleHelper::IsBeamParticle);
+
+            beamParticleWeight += thisBeamParticleWeight;
+            totalWeight += thisTotalWeight;
+        }
+
+        if (beamParticleWeight/totalWeight > m_minWeightFraction)
```
PandoraPFA: Need to check that `totalWeight` is not zero.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
```diff
@@ -0,0 +1,82 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void CheatingBeamParticleIdTool::SelectOutputPfos(const pandora::Algorithm *const /*pAlgorithm*/, const SliceHypotheses &nuSliceHypotheses, const SliceHypotheses &crSliceHypotheses, PfoList &selectedPfos)
+{
+    if (nuSliceHypotheses.size() != crSliceHypotheses.size())
+        throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
+
+    IntVector beamSliceIndices;
```
PandoraPFA: This could be a `std::set` of integers, given that we're only looking to see (later) if a given integer is in the set. However, no real problem with this being a vector as the number of slices is likely to only be of the order of a few tens.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
```diff
@@ -0,0 +1,82 @@
...
+            float thisBeamParticleWeight(0.f), thisTotalWeight(0.f);
+            CheatingParticleIdBaseTool::GetTargetParticleWeight(&downstreamPfos, false, thisBeamParticleWeight, thisTotalWeight, LArMCParticleHelper::IsBeamParticle);
+
+            beamParticleWeight += thisBeamParticleWeight;
+            totalWeight += thisTotalWeight;
+        }
+
+        if (beamParticleWeight/totalWeight > m_minWeightFraction)
+        {
+            beamSliceIndices.push_back(sliceIndex);
```
PandoraPFA: Shame that couldn't share this method with the neutrino case, but difference (single neutrino selected, vs. multiple beam particles) is clear.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
```diff
@@ -0,0 +1,82 @@
...
+        for (const Pfo *const pNeutrinoPfo : neutrinoPfoList)
+        {
+            if (!LArPfoHelper::IsNeutrino(pNeutrinoPfo))
+                throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
+
+            PfoList downstreamPfos;
+            LArPfoHelper::GetAllDownstreamPfos(pNeutrinoPfo, downstreamPfos);
+
+            float thisBeamParticleWeight(0.f), thisTotalWeight(0.f);
+            CheatingParticleIdBaseTool::GetTargetParticleWeight(&downstreamPfos, false, thisBeamParticleWeight, thisTotalWeight, LArMCParticleHelper::IsBeamParticle);
```
PandoraPFA: Is now the time that we need to introduce a difference between triggered beam particles and background beam particles, or can we stay as we are until we've reached some development milestones?

larpandoracontent/LArCheating/CheatingNeutrinoIdTool.cc
```diff
@@ -115,7 +40,7 @@ void CheatingNeutrinoIdTool::SelectOutputPfos(const pandora::Algorithm *const /*
             LArPfoHelper::GetAllDownstreamPfos(pNeutrinoPfo, downstreamPfos);
 
             float thisNeutrinoWeight(0.f), thisTotalWeight(0.f);
-            CheatingNeutrinoIdTool::GetNeutrinoWeight(&downstreamPfos, false, thisNeutrinoWeight, thisTotalWeight);
+            CheatingParticleIdBaseTool::GetTargetParticleWeight(&downstreamPfos, false, thisNeutrinoWeight, thisTotalWeight, LArMCParticleHelper::IsNeutrino);
             neutrinoWeight += thisNeutrinoWeight;
         }
```
PandoraPFA: Note, for reference: not dividing by total weight here (in neutrino id version), just using numerator.

larpandoracontent/LArCheating/CheatingParticleIdBaseTool.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+ *  @brief  CheatingParticleIdBaseTool class
+ */
+class CheatingParticleIdBaseTool : public SliceIdBaseTool
+{
+public:
+    /**
+     *  @brief  Constructor
+     */
+    CheatingParticleIdBaseTool();
```
PandoraPFA: Can just let the compiler provide this default constructor.

larpandoracontent/LArCheating/CheatingParticleIdBaseTool.cc
```diff
@@ -0,0 +1,107 @@
...
+        const MCParticle *const pParentMCParticle(LArMCParticleHelper::GetParentMCParticle(pMCParticle));
+
+        if (fCriteria(pParentMCParticle))
+            targetParticleWeight += weight;
+
+        totalWeight += weight;
+    }
+
+    // ATTN normalise arbitrary input weights at this point
+    if (totalWeight > std::numeric_limits<float>::epsilon())
```
PandoraPFA: Did you need to change any of these functions? There's only this last assignment of the weights that look different.

StevenGreen1: The only difference in these functions is that the neutrino labels have been changed to target particle labels and the IsNeutrino function replaced by fCriteria.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
StevenGreen1: Checked in restructuring of code.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
StevenGreen1: Removed in restructuring of code.

larpandoracontent/LArCheating/CheatingBeamParticleIdTool.cc
StevenGreen1: I think we need to do this soon, but I don't think that this needs to happen immediately for this pull request.  We should discuss the development milestones and figure out when would be best to do this.

StevenGreen1:
I've implemented a restructuring to remove the unnecessary for loop and have addressed the other comments.  I again tested this on a few ProtoDUNE events and everything seemed to be working well.

PandoraPFA:
Done, plus collected related commits (from a number of PRs) into a reduced number of individual 'features', each with a merge commit.

PR #31, opened 2018-03-07
Redesigned method for vetoing stitched particles in nu/beam particle id.

StevenGreen1:


StevenGreen1:
This branch is currently being tested on protodune events.  The events process as normal, but determining the impact of the changes is proving challenging.  Further updates will follow.

StevenGreen1:
There was a small error in the code in the original branch, which has been fixed.  When cutting on x0 to determine whether a particles is a clear cosmic ray, I was assuming the x0 being cut on was an unsigned float when it was actually a signed float.  Adding a std::fabs when checking resolved this issue and now all stitched particles are identified as clear cosmic rays.  I tested this by hand scanning a few ProtoDUNE events and I believe this is ready to be merged.  All work has been squashed into a single commit in the history.

PandoraPFA:
Great, glad this is understood. Continued testing, once this is in the master branch would obviously be greatly appreciated too. We can't be too careful when the issues are so very subtle! Thanks!

PR #32, opened 2018-03-11
CrossGapsExtensionAlgorithm

astblake:
Hi John,

Here (at last...) is a modified version of CrossGapsExtensionAlgorithm, which handles drift-type gaps as well as wire-type gaps.

The algorithm considers one gap at a time. For each gap, it tries to stitch together any associated clusters in proximity to the gap. The clusters don't actually have to go across the gap; they can enter and exit the gap from the same side if they have good pointing. I added an additional method in LArGeometryHelper to check whether a pair of vertices are in proximity to a gap.

If there are multiple associations, then the algorithm keeps tracks of ambiguities, and tries to decide between them based on the length of the clusters and their pointing information. It's fairly conservative about this decision, and will only join together "strongly associated" clusters. 

Anyway - let me know what you think.

Best Wishes,
Andy


larpandoracontent/LArHelpers/LArGeometryHelper.h
```diff
@@ -208,6 +208,20 @@ class LArGeometryHelper
...
+     *
+     *  @param  pandora the associated pandora instance
+     *  @param  p1 the first test point
+     *  @param  p2 the second test point
+     *  @param  hitType the hit type
+     *  @param  gapTolerance the gap tolerance
+     *
+     *  @return boolean
+     */
+    static bool IsInGap(const pandora::Pandora &pandora, const pandora::CartesianVector &p1, const pandora::CartesianVector &p2,
```
PandoraPFA: Just a thought - should the function name make it clear that the requirement is for the points to live in the *same* gap? Something like `IsInSingleGap`, `DoPointsLieInSingleGap`, etc.? I'm sure you can think of a better name than these two suggestions!

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -390,7 +390,7 @@ CartesianVector LArGeometryHelper::GetWireAxis(const Pandora &pandora, const Hit
 
 bool LArGeometryHelper::IsInGap(const Pandora &pandora, const CartesianVector &testPoint2D, const HitType hitType, const float gapTolerance)
 {
-    // ATTN: input test point MUST be a 2D position vector 
+    // ATTN: input test point MUST be a 2D position vector
```
PandoraPFA: Whitespace changes also in same pull request, but hey...

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -20,21 +20,21 @@ namespace lar_content
...
 CrossGapsExtensionAlgorithm::CrossGapsExtensionAlgorithm() :
-    m_minClusterLength(5.f),
-    m_minGapFraction(0.5f),
-    m_maxGapTolerance(2.f),
+    m_minClusterLength(10.f),
+    m_maxGapTolerance(2.5f),
     m_maxTransverseDisplacement(2.5f),
-    m_maxRelativeAngle(10.f)
+    m_maxRelativeAngle(15.f),
+    m_minCosRelativeAngle(0.966)
```
PandoraPFA: If a single float, rather than double float, normally end with `f`, e.g. `0.966f`

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -20,21 +20,21 @@ namespace lar_content
...
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 void CrossGapsExtensionAlgorithm::GetListOfCleanClusters(const ClusterList *const pClusterList, ClusterVector &clusterVector) const
 {
-    // ATTN May want to opt-out completely if no gap information available
-    // if (PandoraContentApi::GetGeometry(*this)->GetDetectorGapList().empty())
-    //     return;
+    // ATTN - Opt out completely if there is no gap information available
+    if (PandoraContentApi::GetGeometry(*this)->GetDetectorGapList().empty())
+         return;
```
PandoraPFA: Presumably this could change reconstruction output, although there are now few detector models which have no registered gaps (just CR steps for DUNE multi-volume detectors)? This, then, is the kind of innocuous line that will raise questions later on.

Are you sure this is the way forwards, and can vouch for it later? As an intermediate, can always make this on/off configurable. If you think it should behave in the new way, we should push to have it work in this new manner.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -51,72 +51,90 @@ void CrossGapsExtensionAlgorithm::GetListOfCleanClusters(const ClusterList *cons
...
-    LArPointingClusterList innerPointingClusterList, outerPointingClusterList;
-    this->BuildPointingClusterList(clusterVector, innerPointingClusterList, outerPointingClusterList);
+    // Build and loop over list of pointing clusters
+    LArPointingClusterList pointingClusterList;
+    this->BuildPointingClusterList(clusterVector, pointingClusterList);
 
-    // Form associations between pairs of pointing clusters
-    for (const LArPointingCluster &pointingClusterInner : innerPointingClusterList)
+    for (LArPointingClusterList::const_iterator iter1 = pointingClusterList.begin(), iterEnd1 = pointingClusterList.end();
+        iter1 != iterEnd1; ++iter1)
```
PandoraPFA: Could now use to c++11 range-based for loop, if you want. This looks clunky now! Ditto below.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -51,72 +51,90 @@ void CrossGapsExtensionAlgorithm::GetListOfCleanClusters(const ClusterList *cons
...
+                throw StatusCodeException(STATUS_CODE_FAILURE);
+
+            // Get closest pair of vertices from pointing clusters
+            LArPointingCluster::Vertex closestVertex1, closestVertex2;
+
+            try
+            {
+                LArPointingClusterHelper::GetClosestVertices(pointingCluster1, pointingCluster2, closestVertex1, closestVertex2);
+            }
+            catch (StatusCodeException &)
```
PandoraPFA: `catch (const StatusCodeException &)`

There are rather too many exceptions being raised, and caught as a matter of routine in LArContent...

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -51,72 +51,90 @@ void CrossGapsExtensionAlgorithm::GetListOfCleanClusters(const ClusterList *cons
 
 void CrossGapsExtensionAlgorithm::FillClusterAssociationMatrix(const ClusterVector &clusterVector, ClusterAssociationMatrix &clusterAssociationMatrix) const
 {
-    // Build lists of pointing clusters in proximity to gaps
-    LArPointingClusterList innerPointingClusterList, outerPointingClusterList;
-    this->BuildPointingClusterList(clusterVector, innerPointingClusterList, outerPointingClusterList);
+    // Build and loop over list of pointing clusters
+    LArPointingClusterList pointingClusterList;
+    this->BuildPointingClusterList(clusterVector, pointingClusterList);
```
PandoraPFA: Is the `pointingClusterList` sorted in a meaningful, and reproducible way?

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -51,72 +51,90 @@ void CrossGapsExtensionAlgorithm::GetListOfCleanClusters(const ClusterList *cons
...
-//------------------------------------------------------------------------------------------------------------------------------------------
+            const float cosTheta(-1.f * closestVertex1.GetDirection().GetDotProduct(closestVertex2.GetDirection()));
+            const float logTheta(-1.f * std::log10(1.0 - cosTheta + std::numeric_limits<float>::epsilon()));
 
-void CrossGapsExtensionAlgorithm::BuildPointingClusterList(const ClusterVector &clusterVector, LArPointingClusterList &innerPointingClusterList,
-    LArPointingClusterList &outerPointingClusterList) const
-{
-    // Convert each input cluster into a pointing cluster
-    LArPointingClusterList pointingClusterList;
+            const float figureOfMerit12(logTheta * (2.f * clusterLength2 - clusterSeparation));
```
PandoraPFA: Figure of merit might warrant a ATTN comment to (briefly) explain aims.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
```diff
@@ -51,72 +51,90 @@ void CrossGapsExtensionAlgorithm::GetListOfCleanClusters(const ClusterList *cons
...
-        const LArPointingCluster::Vertex &pointingVertex(useInner ? pointingCluster.GetInnerVertex() : pointingCluster.GetOuterVertex());
-        const HitType hitType(LArClusterHelper::GetClusterHitType(pointingCluster.GetCluster()));
-
-        if (LArGeometryHelper::IsInGap(this->GetPandora(), pointingVertex.GetPosition(), hitType, m_maxGapTolerance))
-            outputPointingClusterList.push_back(pointingCluster);
+        try
+        {
+            pointingClusterList.push_back(LArPointingCluster(pCluster));
+        }
+        catch (StatusCodeException &)
```
PandoraPFA: const again.

In the long term, we should consider getting into the habit of checking whether to bother attempting construction of a LArPointingCluster, rather than always trying and adding a blanket catch block, which can hide a number of things and can be so slow if an exception is raised.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
astblake: This won't change any output (will it?). The two CrossGaps algorithms will only modify clusters that are in proximity to detector gaps. If there are no gaps, the algorithms will ultimately do nothing. It seems most efficient to bail out straight away if there are no gaps.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
astblake: That's a good point. But these aren't full loops. The first loop is over i=[0,n] and the second loop is over j=[i,n] i.e. we only need to loop over half the n by n matrix. Is there a nice c++11 way of handling these loops?

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
astblake: The output pointingClusterList (std::vector) ought to have the same ordering as the input clusterVector, which has already been sorted.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
astblake: Yes, that's a good point - I've added some words of explanation. 

As a TODO for the future, we ought to think about how we analyse length and pointing information when stitching together tracks. By default, we've been making the joins that yield the longest tracks, but this isn't always the best decision.

astblake:
Hi John,

I've updated CrossGapsExtensionAlgorithm and LArGeometryHelper and replied to your comments. 

Let me know if there's a way that I can address the comment about c++11 loops.

This algorithms mostly stitches together long downstream tracks. The cuts are quite tight, so the upstream stitches look good. I think we'll need a new or modified 2D/3D tool to handle some of the remaining (more ambiguous) joins.

Best Wishes,
Andy

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
PandoraPFA: I agree, that it will only have an impact if the algorithm actually ended up making cluster merges even if there were no detector gaps in the vicinity! I don't know whether this remains the case for our gap-crossing algorithms, but I firmly recall that they did do this immediately after implementation, back in 2016. (Basically, they contained new cluster-association mechanisms, of which the gaps played a strong role, but weren't mandatory. Not ideal, certainly not from an aesthetic point of view, but the merges all looked good when I last examined this!)

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
PandoraPFA: OK, this probably remains the best approach for now.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsExtensionAlgorithm.cc
PandoraPFA: Great, this is super important! I can't see any loops over unordered containers, but if you recall such an operation then please let me know ASAP!!!

PandoraPFA:
I'll check this out locally, rebase, quickly check the diffs again and then aim to merge and push to master. If @loressa is then able to check the impact of this change in her ongoing DUNE FD validation, then that would be excellent team work! Thanks!

PandoraPFA:
Looking good, but I did spot that this will (most likely) change current performance, including for MicroBooNE. In addition to the whole whether the algorithm is enabled with/without gaps (which will affect only cosmic-ray worker instances for ProtoDUNE and other detectors without wire-type gaps), the following parameters have changed their default values:

In `CrossGapsExtensionAlgorithm`:
`m_minClusterLength` from 5.f to 10.f
`m_maxGapTolerance` from 2.f to 2.5f
`m_maxRelativeAngle` from 10.f to 15.f and a `m_minCosRelativeAngle` added.

So, how do we want to handle this? The route forwards has to be deciding what the default values should be, then making a pull request to LArReco with altered values in the XML file(s) where algorithms need to have non-zero values.

Finally, it's not clear to me that the same behaviour will be obtained, even if we restore the default values. We don't always need to maintain the same behaviour (improving is good!), but do we know that the performance should be at least comparable?

With all this in mind, I'm then ready to merge my local branch, but will hold off for now.

loressa:
Hi both,
I have tested this in DUNEFD and MicroBooNE. Details will follow by email. 
Lorena

PandoraPFA:
Any further progress on this, @astblake?

PR #33, opened 2018-03-12
AdaBoostDecisionTree

StevenGreen1:
Created the LArAdaBoostDecisionTree class, restructured the use of the LArSupportVectorMachine and converted the LArSvmHelper into the more generic LArMvaHelper.

This code contains the framework needed for applying a bdt in Pandora.  The beam particle id is not ready to be applied at this time (and so is not included in this PR) as more work is needed on training the model.  However, this framework is sound and was extensively tested when presenting results in the last group meeting.


larpandoracontent/LArHelpers/LArMvaHelper.h
```diff
@@ -1,26 +1,57 @@
...
-#include "larpandoracontent/LArObjects/LArSupportVectorMachine.h"
+#ifndef LAR_MVA_HELPER_H
+#define LAR_MVA_HELPER_H 1
 
 #include <fstream>
 #include <chrono>
 #include <ctime>
 
+class AdaBoostDecisionTree;
+class SupportVectorMachine;
```
PandoraPFA: Looks like these aren't needed (they aren't in namespace lar_content and everything still builds)

larpandoracontent/LArHelpers/LArMvaHelper.h
```diff
@@ -1,26 +1,57 @@
...
+
 namespace lar_content
 {
 
+typedef std::vector<double> DoubleVector;
```
PandoraPFA: Would prefer `DoubleVector` to be public, within `MvaFeatureTool` or `LArMvaHelper` classes. Would control use of this typedef.

larpandoracontent/LArHelpers/LArMvaHelper.h
```diff
@@ -1,26 +1,57 @@
...
+     *  @brief  Run the algorithm tool
+     *
+     *  @param  featureVector the vector of features to append
+     *  @param  args arguments to pass to the tool
+     */
+    virtual void Run(DoubleVector &featureVector, Ts... args) = 0;
+};
+
+template <typename ...Ts>
+using MvaFeatureToolVector = std::vector<MvaFeatureTool<Ts...> *>;
```
PandoraPFA: Need some thought about controlling typedefs, as per above (although equivalent to this has been in `LArSupportVectorMachine` for a while).

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+
+#include "Helpers/XmlHelper.h"
+
+#include "larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h"
+
+namespace lar_content
+{
+
+AdaBoostDecisionTree::AdaBoostDecisionTree() :
+    m_isInitialized(false)
```
PandoraPFA: Check all member variables initialised as appropriate; what about `m_pStrongClassifier`? Should be e.g. `nullptr`

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+{
+    std::cout << "AdaBoostDecisionTree::AdaBoostDecisionTree" << std::endl;
```
PandoraPFA: Debug relic?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+
+pandora::StatusCode AdaBoostDecisionTree::Initialize(const std::string &parameterLocation, const std::string &bdtName)
+{
+    if (m_isInitialized)
+    {
+        std::cout << "AdaBoostDecisionTree: bdt was already initialized" << std::endl;
+        return pandora::STATUS_CODE_FAILURE;
+    }
+
+    // ATTN: Add consistency check for number of features, bearing in min not all features in a bdt may be used
```
PandoraPFA: bearing in mind

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+    this->ReadXmlFile(parameterLocation, bdtName);
+
```
PandoraPFA: Would probably be nicer to allocate the strong classifier here, rather than in the reading of the xml file. Need to check for success. Also need to police use of `m_pStrongClassifier` throughout, in case accidentally still used uninitialised (which should now be `nullptr` as per above).

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+pandora::StatusCode AdaBoostDecisionTree::Initialize(const std::string &parameterLocation, const std::string &bdtName)
+{
+    if (m_isInitialized)
+    {
+        std::cout << "AdaBoostDecisionTree: bdt was already initialized" << std::endl;
+        return pandora::STATUS_CODE_FAILURE;
+    }
+
+    // ATTN: Add consistency check for number of features, bearing in min not all features in a bdt may be used
+    this->ReadXmlFile(parameterLocation, bdtName);
```
PandoraPFA: No return value, or exceptions that can be raised?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+ 
+    m_isInitialized = true;
+    return pandora::STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool AdaBoostDecisionTree::Classify(const DoubleVector &features) const
+{
+    return (m_pStrongClassifier->Predict(features) > 0.0 ? true : false);
```
PandoraPFA: What if `m_pStrongClassifier` is `not yet initialised? And in all other public functions, below.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+double AdaBoostDecisionTree::CalculateClassificationScore(const DoubleVector &features) const
+{
+    return m_pStrongClassifier->Predict(features);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+double AdaBoostDecisionTree::CalculateProbability(const DoubleVector &features) const
+{
+    return (m_pStrongClassifier->Predict(features) + 1.0) * 0.5;
```
PandoraPFA: Where does this come from?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void AdaBoostDecisionTree::ReadXmlFile(const std::string &bdtFileName, const std::string &bdtName)
+{
+    m_isInitialized = true;
```
PandoraPFA: Don't set this until done? Maybe back in `Initialize`?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+    {
+        if (pandora::STATUS_CODE_SUCCESS != this->ReadComponent(pCurrentXmlElement))
+        {
+            std::cout << "AdaBoostDecisionTree: Unknown component in xml file" << std::endl;
+            throw pandora::StatusCodeException(pandora::STATUS_CODE_FAILURE);
+        }
+
+        pCurrentXmlElement = pCurrentXmlElement->NextSiblingElement();
+    }
+    m_pStrongClassifier = new StrongClassifier(m_weakClassifiers);
```
PandoraPFA: See previous comments. When is `m_pStrongClassifier` deallocated?

Can we avoid the `m_weakClassifiers` member variable and just keep containers of weak classifiers on the local stack, passed via reference between functions, until this allocation is done, or does this not fit later usage?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+pandora::StatusCode AdaBoostDecisionTree::ReadComponent(pandora::TiXmlElement *pCurrentXmlElement)
```
PandoraPFA: Can this (and any/all other member functions) be a const member function? Recheck all.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+        }
+        else if ("Node" == pHeadTiXmlElement->ValueStr())
+        {
+            pandora::TiXmlHandle nodeHandle(pHeadTiXmlElement);
+            this->ReadNode(nodeHandle, idToNodeMap);
+        }
+    }
+
+    const WeakClassifier *pWeakClassifier = new WeakClassifier(idToNodeMap, boostWeight, treeIndex);
+    m_weakClassifiers.push_back(pWeakClassifier);
```
PandoraPFA: Do these need to be heap allocated? Are they definitely deleted at job end? What if an exception is raised during operations? Generally safer to place local instances in a local container, so that everything is tidied up when they drop off the stack. If they need to be heap based, so be it, but be super careful with the memory management throughout. Treat every function call as if it could raise an uncaught exception; are your objects tidied up?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool AdaBoostDecisionTree::WeakClassifier::Predict(const DoubleVector &features) const 
+{   
+    return this->EvaluateNode(0, features); 
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool AdaBoostDecisionTree::WeakClassifier::EvaluateNode(const int &nodeID, const DoubleVector &features) const
```
PandoraPFA: No need to have const int by reference, just const by value for built-ins.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+
+    bool isLeaf(leftDaughterID == -1 ? true : false);
+
+    const Node *pNode = new Node(nodeID, parentID, leftDaughterID, rightDaughterID, isLeaf, threshold, variableID, outcome);
+    idToNodeMap.insert(IDToNodeMap::value_type(nodeID, pNode));
+
+    return pandora::STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
```
PandoraPFA: Typically two dividers as now moving to implementation of a new class.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+
+AdaBoostDecisionTree::Node::Node(const int &nodeID, const int &parentNodeID, const int &leftChildNodeID, const int &rightChildNodeID, const bool &isLeaf, const double &threshold, const int &variableID, const bool outcome) :
```
PandoraPFA: Check when you need const by reference (big containers, typically) and const by value (built ins, typically), or non-const by reference (calling function wants to receive changed variable).

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+
+    if (m_idToNodeMap.find(nodeID) != m_idToNodeMap.end())
+    {
+        pActiveNode = m_idToNodeMap.at(nodeID);
+    }
+
+    if (pActiveNode->IsLeaf())
+        return pActiveNode->GetOutcome();
+
+    if (features.at(pActiveNode->GetVariableID()) <= pActiveNode->GetThreshold())
```
PandoraPFA: Exception handling; what if `features` doesn't have a key that matches `pActiveNode->GetVariableID()`?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+    }
+    else 
+    {
+        return this->EvaluateNode(pActiveNode->GetRightChildNodeID(), features);
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+AdaBoostDecisionTree::StrongClassifier::StrongClassifier(WeakClassifiers &weakClassifiers) :
```
PandoraPFA: Probably const reference argument here.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+{
+    double score(0.0);
+    double weights(0.0);
+
+    for (const WeakClassifier *pWeakClassifier : m_weakClassifiers)
+    {
+        weights += pWeakClassifier->GetWeight();
+
+        if (pWeakClassifier->Predict(features))
+	{
```
PandoraPFA: Whitespace alignment.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,299 @@
...
+             score += pWeakClassifier->GetWeight();
+        }
+        else
+        {
+            score -= pWeakClassifier->GetWeight();
+        }
+    }
+
+    score /= weights;
```
PandoraPFA: Care, could divide by zero here if e.g. `m_weakClassifiers` is empty or classifier weights are zero.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+#include "larpandoracontent/LArHelpers/LArMvaHelper.h"
+
+#include <functional>
+#include <map>
+#include <vector>
+
+namespace lar_content
+{
+
+//------------------------------------------------------------------------------------------------------------------------------------------
```
PandoraPFA: Not needed here.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ *  @brief  AdaBoostDecisionTree class
+ */
+class AdaBoostDecisionTree
+{
+public:
+//    typedef std::vector<double> DoubleVector;
```
PandoraPFA: Relic.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+
+namespace lar_content
+{
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+/**
+ *  @brief  AdaBoostDecisionTree class
+ */
+class AdaBoostDecisionTree
```
PandoraPFA: Is it possible for AdaBoostDecisionTree and SupportVectorMachine to have a common interface enforced, so that they *must* work within the constraints of the `LArMvaHelper`, rather than just working because you have made them the same? Would probably just need a purely abstract base class.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+         *  @param  nodeID id of node
+         *  @param  parentNodeID id of parent node
+         *  @param  leftChildNodeID id of left child node
+         *  @param  rightChildNodeID id of right child node
+         *  @param  isLeaf is node a leaf
+         *  @param  threshold threshold for cut
+         *  @param  variableID index
+         *  @param  outcome, only written if leaf node 
+         */
+        Node(const int &nodeID, const int &parentNodeID, const int &leftChildNodeID, const int &rightChildNodeID, const bool &isLeaf, const double &threshold, const int &variableID, const bool outcome = false);
```
PandoraPFA: See earlier comments r.e. const reference arguments.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+    /**
+     *  @brief  Read the bdt parameters from an xml file
+     *
+     *  @param  bdtFileName the sml file name
+     *  @param  bdtName the name of the bdt
+     */
+    void ReadXmlFile(const std::string &bdtFileName, const std::string &bdtName);
+
+private:
+    class Node
```
PandoraPFA: Doxygen brief for class.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+        const int       m_rightChildNodeID;     ///< Right child node id
+        const bool      m_isLeaf;               ///< Is node a leaf
+        const double    m_threshold;            ///< Threshold used for decision if decision node
+        const int       m_variableID;          ///< Variable cut on for decision if decision node
+        const bool      m_outcome;              ///< Outcome if leaf node
+    };
+
+    typedef std::map<int, const Node*> IDToNodeMap;
+
+    class WeakClassifier
```
PandoraPFA: Doxygen brief for class.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+    {
+    public: 
+        /**
+         *  @brief  Constructor, set hierarchy for nodes 
+         */
+        WeakClassifier(IDToNodeMap &idToNodeMap, const double &weight, const int &treeID);
```
PandoraPFA: const `IDToNodeMap`? Just const by value for other two arguments.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+         */
+        bool Predict(const DoubleVector &features) const;
+
+        /**
+         *  @brief  Evalute node and return outcome
+         *
+         *  @param  nodeID current node id 
+         *  @param  features the input features 
+         */
+        bool EvaluateNode(const int &nodeID, const DoubleVector &features) const;
```
PandoraPFA: Just `const int nodeID`. Prefer `nodeId`, but hey...

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+
+    private:
+        IDToNodeMap     m_idToNodeMap; ///< Decision tree nodes
+        const double    m_weight;      ///< Boost weight 
+        const int       m_treeID;      ///< Decision tree id        
+    };
+
+    typedef std::vector<const WeakClassifier*> WeakClassifiers;
+
+    class StrongClassifier
```
PandoraPFA: Doxygen brief.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+    {
+    public:
+        /**
+         *  @brief  Constructor
+         */
+        StrongClassifier(WeakClassifiers &weakClassifiers);
```
PandoraPFA: const argument

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+    };
+
+    /**
+     *  @brief  Read the component at the current xml element
+     *
+     *  @param  pCurrentXmlElement address of the current xml element
+     *
+     *  @return success
+     */
+    pandora::StatusCode ReadComponent(pandora::TiXmlElement *pCurrentXmlElement);
```
PandoraPFA: Check whether can be const member functions, as per earlier comment.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
```diff
@@ -0,0 +1,319 @@
...
+     *
+     *  @param  currentHandle the current xml handle
+     *  @param  idToNodeMap map of id to node
+     *
+     *  @return success
+     */
+    pandora::StatusCode ReadNode(pandora::TiXmlHandle &currentHandle, IDToNodeMap &idToNodeMap);
+
+    WeakClassifiers       m_weakClassifiers;             ///< Vector of all weak classifiers
+    StrongClassifier     *m_pStrongClassifier;           ///< Strong adaptive boost tree classifier
```
PandoraPFA: Where/when are the memory allocations tidied up?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Check added.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Added and exception thrown if unsuccessful reading of xml file.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Removed.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArHelpers/LArMvaHelper.h
StevenGreen1: I left this here so that when you add DoubleVector to the pandora typedefs you wouldn't have to change any code currently using DoubleVector.  I was under the impression that this was the plan, but if you don't want to add DoubleVector to the pandora typedefs I'll change the location of this to make sure it's use is policed.

larpandoracontent/LArHelpers/LArMvaHelper.h
StevenGreen1: Removed.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Normalising the BDT score by the weights, which I believe is standard procedure in things like TMVA, produced BDT scores between -1 and 1.  This linear mapping changes that to give a score between 0 and 1, which we can interpret as a probability.  I don't know if a more appropriate mapping is possible, but this is what's done in the pandora SVM use case also.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Moved to stack allocation.  Still need to clean data in case of exceptions being thrown.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.  Am I correct in assuming strings are not built-in types?

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Added check.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Corrected.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
StevenGreen1: Check added and exception thrown if weights sum zero.  This would be equivalent to m_weakClassifiers being empty so no additional check on this implemented.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Removed.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Removed.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Base class created and used inside MvaHelper.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
PandoraPFA: Definitely worth passing string arguments by const reference (character type is the basic type).

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
PandoraPFA: Probably best to add an `// ATTN` comment to explain this.

StevenGreen1: Done.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h
StevenGreen1: Restructured code to ensure memory management is taken care of.

StevenGreen1:
I believe that all the requested changes have been made with the only exception being the typedef for the DoubleVector.  I don't think that this change can be easily made for two reasons:

1. The feature tool vector class is templated and adding the typdef here means we have to arbitrarily specify some sort of specialization everywhere we use DoubleVector, which has no bearing at all on the DoubleVector object itself.  This seems like an unnecessary complication to enforce just to police the use of this typedef.
2. Adding DoubleVector to the MvaHelper class adds further complications to the data structure in place.  The main issue being that DoubleVector is needed in both the LArMvaHelper and the LArMultivariateAnalysisBaseClass classes and the LArMvaHelper class needs to know about the LArMultivariateAnalysisBaseClass.  Therefore, adding DoubleVector to the LArMvaHelper class means we have a circular dependence between these classes, which we should avoid.  I also don't believe it's possible to forward declare a typedef so there seems no way around this. 

I could create a separate class to sit alongside the FeatureTool class that contains just this typedef, possible called FeatureClass, and put DoubleVector in there?  That way it would control the use of the typedef, but prevent these issues.  It would be nice to have it in the FeatureTool, but having this templated I think makes this impossible (or at least extremely ugly).  Thoughts?

PandoraPFA:
Hi Steve,

Yes, this is exactly as we discussed and a little frustrating. I'll have a think this afternoon too.

Loose typedefs do create big maintenance problems later on, so it's worth thinking about this properly now. The idea of a separate class, containing just the typedef, is actually something I have done before, but it does add yet another name to the space.

You can actually further typedef the newly-defined typedef to map it into the LArMvaHelper, finally giving the LArMvaHelper::FeatureVector or LArMvaHelper::DoubleVector that we originally wanted, but algs could still choose to use MyRandomSpareClassName::FeatureVector if they wanted to!

I'll add some comments about your destructors/reset methods separately. The seminar is now coming in at 58 mins, which is quite a result given the first read through! :)

Thanks!

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,394 @@
...
+#include "larpandoracontent/LArObjects/LArAdaBoostDecisionTree.h"
+
+namespace lar_content
+{
+
+AdaBoostDecisionTree::AdaBoostDecisionTree() :
+    m_pStrongClassifier(nullptr)
+{
+}
+
```
PandoraPFA: Destructor needed to delete `m_pStrongClassifier`. Note that deleting `nullptr` is fine, so don't need to check before delete, but otherwise need to check before any attempt to deference.

Then need to add copy constructor and copy assignment operator, as the system defaults would e.g. give pointer to rhs-owned strong classifier to the lhs, which could proceed to delete it.

See https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming), now rule of five with C++11, although we probably don't adhere to the rule of five everywhere yet.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,394 @@
...
+    if (pandora::STATUS_CODE_SUCCESS != statusCode)
+    {
+        this->Reset();
+        std::cout << "AdaBoostDecisionTree: Unable to read xml" << std::endl;
+        throw pandora::StatusCodeException(statusCode);
+    }
+
+    // ATTN: Add consistency check for number of features, bearing in min not all features in a bdt may be used
+    m_pStrongClassifier = new StrongClassifier(weakClassifiers);
```
PandoraPFA: Can this constructor now raise an exception? If so, need to delete `m_pStrongClassifier` if construction fails.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,394 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void AdaBoostDecisionTree::Reset() const
+{
+    if (!m_pStrongClassifier)
+        return;
+
+    for (const WeakClassifier *pWeakClassifier : m_pStrongClassifier->GetWeakClassifiers())
```
PandoraPFA: This should go in the destructor for the strong classifier.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,394 @@
...
+    {
+        if (!pWeakClassifier)
+            continue;
+
+        pWeakClassifier->Reset();
+        delete pWeakClassifier;
+    }
+
+    delete m_pStrongClassifier;
```
PandoraPFA: More typically the body of the destructor for the decision tree.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,394 @@
...
+
+    try
+    {
+        score = m_pStrongClassifier->Predict(features);
+    }
+    catch (pandora::StatusCodeException &statusCodeException)
+    {
+        if (pandora::STATUS_CODE_NOT_FOUND == statusCodeException.GetStatusCode())
+        {
+            this->Reset();
```
PandoraPFA: Are you sure that you want to rip out everything in the classifier if you get an exception on one event? More typically, just need to ensure that everything is ready/reset to attempt to process next event without things being left in a mess.

larpandoracontent/LArObjects/LArAdaBoostDecisionTree.cc
```diff
@@ -0,0 +1,394 @@
...
+                "BoostWeight", boostWeight));
+        }
+        else if ("Node" == pHeadTiXmlElement->ValueStr())
+        {
+            pandora::TiXmlHandle nodeHandle(pHeadTiXmlElement);
+            this->ReadNode(nodeHandle, idToNodeMap);
+        }
+    }
+
+    const WeakClassifier *pWeakClassifiers = new WeakClassifier(idToNodeMap, boostWeight, treeIndex);
```
PandoraPFA: We should talk about this tomorrow. Still heap allocating weak classifiers in the decision tree class, rather than hiding all the weak classifier knowledge inside the strong classifier.

Can you just give the strong classifier (and then weak classifier) constructors that take xmlHandles?

The diff is getting hard to read, but it looks like the following should work:

-The decision tree should new up a strong classifier, and delete it in its destructor (and make sure that it doesn't leak the memory if the job terminates early).

-The strong classifier should new up its weak classifiers, and delete them in its destructor (and make sure that it doesn't leak memory if exceptions cause the job to terminate early).

-The weak classifiers should new up its nodes, and delete them in its destructor, and so on.

In this way, each object is encapsulated and there isn't a decision tree object that can mess about with the contents of each.

StevenGreen1:
Having just tested this code, it is now broken so I will close the pull request until these issues are resolved.

PR #34, opened 2018-03-20
Fixed small coverity defects

a-d-smith:
Hopefully fixed the three Coverity defects

### DIFF
--- a/larpandoracontent/LArHelpers/LArFormattingHelper.cc
+++ b/larpandoracontent/LArHelpers/LArFormattingHelper.cc
@@ -185,7 +185,7 @@ void LArFormattingHelper::Table::PrintTableCell(const std::string &value, const 
     const unsigned int column(index % m_widths.size());
 
     const std::string separator(this->IsSeparatorColumn(column) ? "" : " ");
-    std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator;
+    std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator << std::resetiosflags(std::ios::adjustfield);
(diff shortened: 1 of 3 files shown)


larpandoracontent/LArHelpers/LArFormattingHelper.cc
```diff
@@ -185,7 +185,7 @@ void LArFormattingHelper::Table::PrintTableCell(const std::string &value, const
     const unsigned int column(index % m_widths.size());
 
     const std::string separator(this->IsSeparatorColumn(column) ? "" : " ");
-    std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator;
+    std::cout << "|" << format << separator << std::setw(m_widths[column]) << std::internal << value << separator << std::resetiosflags(std::ios::adjustfield);
     LArFormattingHelper::ResetStyle();
```
PandoraPFA: Is this maybe something that is better within the ResetStyle method?

Presumably this resets to some system defaults, rather than whether the setup was before you called setw, so it will be interesting to see if Coverity still complains.

I'll leave the location of this up to you, just reply and I'll merge and retry Coverity. In some ways it's better to have the setw and resetiosflags on the very same line. In other ways, that's what the ResetStyle method is for.

Are they resetting the same "kinds" of things?

a-d-smith: My understanding is the Coverity defect was that I had set the adjustfield property to internal without setting it back afterwards (I guess the default is left?). If that's all it is then I think it makes sense to just add it on this line, the ResetStyle method prints the reset formatting character, which sets the colour etc back to default (not to do with text position)

PR #35, opened 2018-03-20
updated exception handling in svm track/shower id

loressa:
content with the performance with respect to latest version, committing the updated handling of exceptions in the track/shower id


larpandoracontent/LArTrackShowerId/PfoCharacterisationBaseAlgorithm.cc
```diff
@@ -57,21 +57,15 @@ StatusCode PfoCharacterisationBaseAlgorithm::Run()
...
-                    isTrackLike = this->IsClearTrack3x2D(pPfo);
-                }
+                isTrackLike = this->IsClearTrack(pPfo);
             }
-            catch (const StatusCodeException &)
+            else
             {
-                continue;
+                isTrackLike = this->IsClearTrack3x2D(pPfo);
             }
```
PandoraPFA: Even better as:
`const bool isTrackLike(m_useThreeDInformation ? this->IsClearTrack(pPfo) : this->IsClearTrack3x2D(pPfo));`

Much shorter :)

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -100,14 +100,26 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowOb
...
-    }
-    else
-    {
-        return (m_minProbabilityCut <= LArSvmHelper::CalculateProbability((wClusterList.empty() ? m_supportVectorMachineNoChargeInfo : m_supportVectorMachine), featureVector));
-    }
+	//check for failures in the calculation of features, which would give -1 in any of the features, if that happens, use current id
+	bool useSVM(true);
+	for (int i=0; i<featureVector.size(); i++)
+		if (std::fabs(featureVector[i] + 1.f) < std::numeric_limits<float>::epsilon())
+				useSVM = false;
```
PandoraPFA: Maybe better just removing the useSVM variable and just setting the value and returning immediately if this check succeeds and you can't proceed with the SVM as normal.

Also, need some parentheses here, and the indentation looks non-standard (at least in browser on GitHub, it does).

Finally, prefer e.g. `for (const double featureValue : featureVector)`, rather than use of operator[]

larpandoracontent/LArTrackShowerId/SvmPfoCharacterisationAlgorithm.cc
```diff
@@ -100,14 +100,26 @@ bool SvmPfoCharacterisationAlgorithm::IsClearTrack(const pandora::ParticleFlowOb
...
+			return LArSvmHelper::Classify((wClusterList.empty() ? m_supportVectorMachineNoChargeInfo : m_supportVectorMachine), featureVector);
+		}
+		else
+		{
+			return (m_minProbabilityCut <= LArSvmHelper::CalculateProbability((wClusterList.empty() ? m_supportVectorMachineNoChargeInfo : m_supportVectorMachine), featureVector));
+		}
+	}
+	else
+	{
+		return (pPfo->GetParticleId() == MU_MINUS);
```
PandoraPFA: As per previous comment, this logic probably belongs up above.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -268,6 +268,8 @@ const pandora::ParticleFlowObject *const pInputPfo)
     LArPfoHelper::GetTwoDClusterList(pInputPfo, clusterList);
     float diffWithStraightLineMean(0.f), maxFitGapLength(0.f), rmsSlidingLinearFit(0.f);
 
+	int nClustersUsed(0);
```
PandoraPFA: Is the indentation actually non-standard here, or is it just the display in GitHub? It looks like 8 spaces per old "tab", rather than our usual 4. Also other indentation to check in this file, below.

PandoraPFA: `unsigned int`, rather than `int`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -279,23 +281,32 @@ const pandora::ParticleFlowObject *const pInputPfo)
...
+			diffWithStraightLineMean   += diffWithStraightLineMeanCluster;
+			maxFitGapLength            += maxFitGapLengthCluster;
+			rmsSlidingLinearFit        += rmsSlidingLinearFitCluster;
+
+			++nClustersUsed;
+		}
     }
 
-    if (!clusterList.empty())
+	float lenght3D(-1.f);
```
PandoraPFA: `lenght` vs. `length`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -349,6 +360,10 @@ void ThreeDLinearFitFeatureTool::CalculateVariablesSlidingLinearFit(const pandor
                 if (correctedGapLength > maxFitGapLength)
                     maxFitGapLength = correctedGapLength;
             }
+			else
+			{
+				maxFitGapLength = 0.f;
```
PandoraPFA: Perhaps ought to initialise this value to its safe default above, and for *all* arguments you take as non-const reference so that the calling algorithm receives the new value. What you'd then be doing is guarding against all cases where the logic doesn't actually allow you to assign to the variable, which will then just maintain the value it was originally assigned in the calling algorithm.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -415,7 +430,7 @@ void ThreeDVertexDistanceFeatureTool::Run(SupportVectorMachine::DoubleVector &fe
 
     const Vertex *const nuVertex(pVertexList->front());
     //find the particle vertex
-   float vertexDistance(-1.f);
+	float vertexDistance(-1.f);
```
PandoraPFA: Indentation?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -463,25 +478,33 @@ void ThreeDOpeningAngleFeatureTool::Run(SupportVectorMachine::DoubleVector &feat
...
-
-    const float openingAngle(this->OpeningAngle(eigenVecsStart.at(0), eigenVecsStart.at(1), eigenValuesStart));
-    const float closingAngle(this->OpeningAngle(eigenVecsEnd.at(0), eigenVecsEnd.at(1), eigenValuesEnd));
-
-    featureVector.push_back(std::fabs(openingAngle-closingAngle));
-
+    float diffAngle(-1.f);
+    if (!threeDCaloHitList.empty())
+    {
+      CartesianPointVector pointVectorStart, pointVectorEnd;
```
PandoraPFA: Now only two spaces for a "tab"?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -463,25 +478,33 @@ void ThreeDOpeningAngleFeatureTool::Run(SupportVectorMachine::DoubleVector &feat
...
+			LArPcaHelper::RunPca(pointVectorStart, centroidStart, eigenValuesStart, eigenVecsStart);
+			LArPcaHelper::RunPca(pointVectorEnd, centroidEnd, eigenValuesEnd, eigenVecsEnd);
+
+			const float openingAngle(this->OpeningAngle(eigenVecsStart.at(0), eigenVecsStart.at(1), eigenValuesStart));
+			const float closingAngle(this->OpeningAngle(eigenVecsEnd.at(0), eigenVecsEnd.at(1), eigenValuesEnd));
+			diffAngle = std::fabs(openingAngle-closingAngle);
+		}
+		catch (const StatusCodeException &)
+		{
+			diffAngle = -1.f;
```
PandoraPFA: Is this actually needed, given the initialisation value, above? It's good to be paranoid sometimes, but I don't think the second assignment looks like it's ever needed.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -463,25 +478,33 @@ void ThreeDOpeningAngleFeatureTool::Run(SupportVectorMachine::DoubleVector &feat
...
-    const float closingAngle(this->OpeningAngle(eigenVecsEnd.at(0), eigenVecsEnd.at(1), eigenValuesEnd));
-
-    featureVector.push_back(std::fabs(openingAngle-closingAngle));
-
+    float diffAngle(-1.f);
+    if (!threeDCaloHitList.empty())
+    {
+      CartesianPointVector pointVectorStart, pointVectorEnd;
+      this->Divide3DCaloHitList(pAlgorithm, threeDCaloHitList, pointVectorStart, pointVectorEnd);
+      if ((pointVectorStart.size()>1)&&(pointVectorEnd.size()>1)) //otherwise not able to calculate angles if only 1 point provided
```
PandoraPFA: Prefer to add a space before e.g. `>` and `&&`

Prefer clear line between call to Divide3DCaloHitList and this if statement, and comment (if it's actually needed) above the if statement (and comment as terse as possible)

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -568,34 +591,41 @@ ThreeDPCAFeatureTool::ThreeDPCAFeatureTool()
...
+	LArPfoHelper::GetCaloHits(pInputPfo, TPC_3D, threeDCaloHitList);
 
-    // Run the PCA analysis
-    CartesianVector centroid(0.f, 0.f, 0.f);
-    LArPcaHelper::EigenVectors eigenVecs;
-    LArPcaHelper::EigenValues eigenValues(0.f, 0.f, 0.f);
-    LArPcaHelper::RunPca(threeDCaloHitList, centroid, eigenValues, eigenVecs);
-    const float principalEigenvalue(eigenValues.GetX()), secondaryEigenvalue(eigenValues.GetY()), tertiaryEigenvalue(eigenValues.GetZ());
+	if (threeDCaloHitList.empty())
+		return;
```
PandoraPFA: What happens if you return and then the feature vector doesn't have the expected size? Do we return anywhere else, in any of the other tools, rather than pushing back a null/dummy value into the feature vector? Won't this cause significant problems and mixing up of doubles as feature values?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -568,34 +591,41 @@ ThreeDPCAFeatureTool::ThreeDPCAFeatureTool()
...
+      const float principalEigenvalue(eigenValues.GetX()), secondaryEigenvalue(eigenValues.GetY()), tertiaryEigenvalue(eigenValues.GetZ()); 
+      if (principalEigenvalue > std::numeric_limits<float>::epsilon())
+        {
+          pca1 = secondaryEigenvalue/principalEigenvalue;
+          pca2 = tertiaryEigenvalue/principalEigenvalue;
+        }
+    }
+    catch (const StatusCodeException &)
+    {
+      pca1 = -1.f;
```
PandoraPFA: As above

loressa:
Hello! 

I have pushed several commits rebasing against recent updates to use LArMvaHelper::MvaFeature (i.e. initialised doubles) and fixed the exception handling so that it is now instead checked in SVMPfoCharacterisationAlgorithm.cc that all features are initialised before proceeding to use SVM classification, otherwise returns the previously cut-based assigned label, here: https://github.com/PandoraPFA/LArContent/pull/35/files#diff-6ebb01ab8c961084e72856a499e2ff0fR104

The initialisation of the features occur in the TrackShowerIdFeatureTool.cc for each tool and feature. I have made use of the default constructor https://github.com/loressa/LArContent/blob/master/larpandoracontent/LArObjects/LArMvaInterface.h#L130
to have empty and non initialised features to start with, e.g. here: 
https://github.com/PandoraPFA/LArContent/pull/35/files#diff-65a3aec7f7a2c8bc439fd21dfabe1234R644

which if there are no exceptions they get assigned the corresponding value (and initialised only then) by using the = operator here:
https://github.com/loressa/LArContent/blob/master/larpandoracontent/LArObjects/LArMvaInterface.h#L154
following the previous example:
https://github.com/PandoraPFA/LArContent/pull/35/files#diff-65a3aec7f7a2c8bc439fd21dfabe1234R654

I have tested both execution time and performance wrt my previous track/shower id efficiencies and I am satisfied with both, efficiencies only differ in <1%, typically 0.1%  as the number of events is slightly different in my two test samples. 

Since the svm track/shower id features handling has diverged considerably, my suggestion is to drop the 2D variables at this stage, and remove them from TrackShowerIdFeatureTool.cc - as the code stands now, they can't be used directly, and the only way that makes sense to use them anymore is to go back to a previous tag release in which they were used, so my suggestion is to remove them, but I will wait for feedback. 

All in all, I am ready to continue with this PR. 

Lorena

PandoraPFA:
Will let the previous feature merges go through the overnight tests (no changes expected), then merge this feature as it is tomorrow.

loressa:
Ta! 
As mentioned in my last comment, I found only differences of <1% in the track/shower efficiencies, typically 0.1% in my latest tests, affected by statistical fluctuations as number of events were slightly different.

PR #36, opened 2018-03-22
AdaBoostDecisionTree

StevenGreen1:
I have applied proper memory management to the heap variables and have tested this code against the python packages and it reproduced the same bdt scores to floating point accuracy.  The history needs to be tidied up, but I believe now is the time for another review.

### DIFF
--- a/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
+++ b/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
@@ -14,7 +14,7 @@
 
 #include "larpandoracontent/LArHelpers/LArMCParticleHelper.h"
 #include "larpandoracontent/LArHelpers/LArPfoHelper.h"
-#include "larpandoracontent/LArHelpers/LArSvmHelper.h"
+#include "larpandoracontent/LArHelpers/LArMvaHelper.h"
 #include "larpandoracontent/LArHelpers/LArPcaHelper.h"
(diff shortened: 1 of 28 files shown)


PandoraPFA:
Hi Steve,

This is basically just about there now. Let's sit down together on Monday afternoon to close out this feature. Without doing an exhaustive review, I think I spotted the following:
-No destructor for AdaBoostDecisionTree yet
-Is there any need for the StrongClassifier:: GetWeakClassifiers() function still, and why does it return a list of WeakClassifiers by value?
-We still need to sort out the feature/double typedef (I have some ideas we can try).

Following discussions with @loressa this morning, I was thinking we could use and pandora::InputType<double> as our Feature type, so that you also receive the information about whether the value has been explicitly assigned to, or is just holding its initialisation value

Thanks!

StevenGreen1:
Hi John,

That sounds good.  I'll work on the first two points now as that shouldn't take long and mean we can have that in place for Monday to do a proper review.  Using a Feature type sounds like a good idea too.  

Cheers
Steve

StevenGreen1:
Hi John,

The problem with the destructor of the base class was the lack of an implementation.  I did not realise that even virtual destructors require an implementation, but I added a null implementation in the LArMultivariateAnalysisBaseClass.h file (check to see if this fits stylistically) and now the test job works fine and gives the same results as the python implementation.  

Cheers

Steve

PR #37, opened 2018-03-29
Created the InitializedDouble class for use in Mva.

StevenGreen1:
Hi John,
Here is a first pass of the new simple double class containing a double and a bool for use in the MvaFeature.  I tested this branch on a few events and although the times were better than using an PandoraInputType, they were still significantly different from using just a double.  For example:

File ID 10018:
Double -  29.424s 
PandoraInputType - 47.167s
InitializedDouble - 34.867s

I tested a few files with similar results also.  If you're happy using the InitializedDouble class then we can iterate any code changes and get the request merged, but if you want to discuss this further then we can do that too.

Cheers
Steve


PandoraPFA:
Hi Steve,

Can you try removing the MvaInterface.cc and inline all the function implementation in MvaInterface.h, then retest for performance?

Thanks!

StevenGreen1:
Unfortunately, I have to leave now as we have a carpenter coming to give us a quote for some work, but I'll get on with that later this afternoon if I've got time and if not next week.

StevenGreen1:
Hi John,  

I managed to find time to make this check.  It seems like removing the LArMvaInterface.cc speeds things up quite a lot and recovers almost the same performance as using just a double.  I'll update this branch now to remove the .cc file. 

Cheers,
Steve

PandoraPFA:
I thought inlining it would help, otherwise you were forcing a function jump for each call (unless the compiler had opted-in to inlining it for you). I'll just flag up a couple of things I spotted for you too.

Thanks!

larpandoracontent/LArObjects/LArMvaInterface.cc
```diff
@@ -0,0 +1,78 @@
...
+
+MvaTypes::InitializedDouble::InitializedDouble() : 
+    m_number(0.0),
+    m_isInitialized(false)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+MvaTypes::InitializedDouble::InitializedDouble(double number) :
```
PandoraPFA: const double

larpandoracontent/LArObjects/LArMvaInterface.cc
```diff
@@ -0,0 +1,78 @@
...
+    m_number(number),
+    m_isInitialized(true)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+MvaTypes::InitializedDouble::InitializedDouble(const InitializedDouble &rhs)
+{
+    m_number = rhs.m_number;
```
PandoraPFA: In initialiser list, rather than body of constructor.

larpandoracontent/LArObjects/LArMvaInterface.h
```diff
@@ -21,10 +21,66 @@ namespace lar_content
 class MvaTypes
 {
 public:
-    typedef pandora::PandoraInputType<double> MvaFeature;
+    class InitializedDouble
```
PandoraPFA: Add doxygen brief.

Could even call this the MvaFeature class and then remove the typedef below, but this still work for me as is!

larpandoracontent/LArObjects/LArMvaInterface.h
```diff
@@ -21,10 +21,66 @@ namespace lar_content
...
+         *  @brief  Default constructor.
+         */
+        InitializedDouble();
+
+        /**
+         *  @brief  Constructor.
+         *
+         *  @param  number to hold in class
+         */
+        InitializedDouble(double number);
```
PandoraPFA: const double, as per above comment

StevenGreen1:
HI John,

I've addressed the comments you made, squashed the new commit so there's just one commit here and pushed it to the branch that this pull request is for.  I left the class name as InitializedDouble, but I can see your point of view with the name changing.  This has been a good learning experience on how compilers work and how big a difference this can make!

Cheers
Steve

PandoraPFA:
Great work! I just did some rebasing (added this commit under the original AdaBoostDecisionTree feature) and pushed it to the master branch.

We should probably create a separate LArMva subdirectory in the near future, along the lines of our previous grouping together of implementation and header files in subdirectories.

Thanks again!

PR #39, opened 2018-04-04
Removed neutrino reference in CheatingCosmicRayTaggingTool.

StevenGreen1:
This pull request addresses the issues of the CheatingCosmicRayTaggingTool defining clear cosmic rays using neutrino weights.  By cutting on the cosmic ray weight instead, this tool then becomes generic and usable for ProtoDUNE. 

I tested this feature branch with respect to the original master.  For MicroBooNE events both with and without cosmics the reconstruction (PandoraSettings_Master_MicroBooNE.xml with LArCheatingCosmicRayTagging) the reconstruction output was floating point identical when running over 10 events.  

When testing this on ProtoDUNE events (PandoraSettings_Master_ProtoDUNE.xml with LArCheatingCosmicRayTagging) the results were significantly different.  This is expected as the original cheated cosmic ray tagging calls everything a cosmic ray (as the neutrino weight for all pfos).  While a direct diff was too messy to compare, the output using the new code gave some neutrino Pfos for test beam particles, which indicates that this new code is working as expected.  

One comment I would put up for discussion is whether we still use the m_minNeutrinoFraction variable in this tool as in the new code we cut on 1 - m_minNeutrinoFraction.  I think it's fairly clear what's going on here from a readability perspective, but I'd be happy to change this to a m_maxCosmicRayFraction that could be configured via xml if this is preferred.

### DIFF
--- a/larpandoracontent/LArCheating/CheatingCosmicRayTaggingTool.cc
+++ b/larpandoracontent/LArCheating/CheatingCosmicRayTaggingTool.cc
@@ -20,7 +20,7 @@ namespace lar_content
 {
 
 CheatingCosmicRayTaggingTool::CheatingCosmicRayTaggingTool() :
-    m_minNeutrinoFraction(0.75f)
+    m_maxCosmicRayFraction(0.25f)
 {
 }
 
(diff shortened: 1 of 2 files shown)


PandoraPFA:
This looks fine as it is, but is it worth replacing the XML key entirely with e.g. `MaxCosmicRayFraction` instead?

StevenGreen1:
Good point, have made that change and pushed to pull request branch.

PandoraPFA:
Thanks. Will merge after CI check.

PR #44, opened 2018-04-17
Moved the GetSigmaUVW function to the LArGeometryHelper.

StevenGreen1:
Moved the GetSigmaUVW function to the LArGeometryHelper.  This prevents an exception being thrown when calling the GetLArTPC function in the GeometryManager when trying to find sigmaUVW in the case of multiple drift volumes.

### DIFF
--- a/larpandoracontent/LArHelpers/LArGeometryHelper.cc
+++ b/larpandoracontent/LArHelpers/LArGeometryHelper.cc
@@ -485,7 +485,7 @@ float LArGeometryHelper::CalculateGapDeltaZ(const Pandora &pandora, const float 
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-double LArGeometryHelper::GetSigmaUVW(const Pandora &pandora, const double maxSigmaDiscrepancy) 
+float LArGeometryHelper::GetSigmaUVW(const Pandora &pandora, const float maxSigmaDiscrepancy) 
 {
     const LArTPCMap &larTPCMap(pandora.GetGeometry()->GetLArTPCMap());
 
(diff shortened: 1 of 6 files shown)


larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -138,7 +138,7 @@ void LArGeometryHelper::MergeTwoPositions(const Pandora &pandora, const HitType
     const float Z3(LArGeometryHelper::MergeTwoPositions(pandora, view1, view2, Z1, Z2));
 
     position3.SetValues(X3, 0.f, Z3);
-    const float sigmaUVW(pandora.GetGeometry()->GetLArTPC().GetSigmaUVW());
+    const float sigmaUVW(LArGeometryHelper::GetSigmaUVW(pandora));
```
PandoraPFA: Should be a single float (as in LArTPC class), except when used in mathematical manipulation in RotationalTransformationPlugin.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -483,4 +483,33 @@ float LArGeometryHelper::CalculateGapDeltaZ(const Pandora &pandora, const float
     return gapDeltaZ;
 }
 
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+double LArGeometryHelper::GetSigmaUVW(const Pandora &pandora, const double maxSigmaDiscrepancy) 
+{
```
PandoraPFA: Review, but maybe only here that needs to be single float, rather than double.

StevenGreen1: I've made this change.  The only place that doesn't use a double sigmaUVW are the LArGeometryHelper::MergeTwo/ThreePositions functions.  There functions use a float sigmaUVW, but do the same manipulation to it as the other functions using double sigmaUVW.  

Having done a bit of searching the chi2 in the LArGeometryHelper::MergeTwo/ThreePositions functions is redundant.  There are 2-3 places inside LArContent where these functions are called and nothing is done with the chi2 in any of them.  Therefore, I think we need to review and either:
1) Don't bother calculating chi2.
2) Change these floats chi2 to doubles if we think the chi2 might be needed later.

PR #42, opened 2018-04-16
Created BDT beam particle id tool.

StevenGreen1:
Created a beam particle id tool that uses a bdt for classification.  There is also an accompanying trained bdt xml file that will be added to the machine learning library once this code has been reviewed.

PR #45, opened 2018-04-19
Neutrino Id Training Purity & Completeness

StevenGreen1:
Modifications to purity and completeness calculation in neutrino id training.

There is potential for a merge conflict in the LArMCParticleHelper.h file with the Bdt pull request (https://github.com/PandoraPFA/LArContent/pull/42) where the same change was made (making CollectReconstructable2DHits a public function).  This is the only change though and should be easy to fix up with some rebasing.  

This was tested on a few neutrino + cosmic events and the changes were minimal.  The purities and completenesses did change slightly, but in all cases seen the best slice index was unchanged. 

@a-d-smith If you have time to have a quick look over this to make sure there are no obvious issues that'd be great.  Thanks.

### DIFF
--- a/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
+++ b/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
@@ -31,6 +31,8 @@ NeutrinoIdTool::NeutrinoIdTool() :
     m_nuance(-std::numeric_limits<int>::max()),
     m_minPurity(0.9f),
     m_minCompleteness(0.9f),
+    m_selectInputHits(true),
+    m_maxPhotonPropagation(2.5f),
     m_minProbability(0.0f),
     m_maxNeutrinos(1),
     m_filePathEnvironmentVariable("FW_SEARCH_PATH")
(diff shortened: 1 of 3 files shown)


a-d-smith:
Looks good to me, the only thing I would say is the LArMCParticleHelper::PrimaryParameters already has default values hardcoded:
https://github.com/PandoraPFA/LArContent/blob/133054eb310b215a18a1cae2f74d68c9f12d1d6e/larpandoracontent/LArHelpers/LArMCParticleHelper.cc#L31-L39

Maybe instead of doing this twice, the constructor could use these values instead?
What do you think?

StevenGreen1:
Thanks for looking over this.  Yeah that's a good idea, I'll make that change now.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -141,6 +158,9 @@ void NeutrinoIdTool::Collect2DHits(const PfoList &pfos, CaloHitList &hitList) co
         // ATTN hits collected from Pfos are copies of hits passed from master instance, we need to access their parent to use MC info
         const CaloHit *pParentCaloHit(static_cast<const CaloHit *>(pCaloHit->GetParentAddress()));
 
+        if (std::find(reconstructableCaloHitList.begin(), reconstructableCaloHitList.end(), pParentCaloHit) == reconstructableCaloHitList.end())
+            continue;
```
PandoraPFA: Consider use of CaloHitSet (note: unordered!) here for lookup.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -167,6 +187,8 @@ unsigned int NeutrinoIdTool::CountNeutrinoInducedHits(const CaloHitList &hitList
     return nNuHits;
 }
 
+
```
PandoraPFA: Two empty lines! Two!

StevenGreen1: I thought it helped the fung shui of the code, but happy to remove.

larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
```diff
@@ -158,21 +159,21 @@ void NeutrinoIdTool::Collect2DHits(const PfoList &pfos, CaloHitList &hitList, Ca
...
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-unsigned int NeutrinoIdTool::CountNeutrinoInducedHits(const CaloHitList &hitList) const
+unsigned int NeutrinoIdTool::CountNeutrinoInducedHits(const CaloHitSet &caloHitSet) const
 {
     unsigned int nNuHits(0);
-    for (const CaloHit *const pCaloHit : hitList)
+    for (const CaloHit *const pCaloHit : caloHitSet)
     {
```
PandoraPFA: As discussed, use unordered_set for the look-ups, but list (or vector) for ordered loops (making sure that contents of list of vector hadn't been in a set as n intermediate step!).

StevenGreen1:
I've tested these changes with a few events and the reported purity and completenesses appear as you would expect.

PR #51, opened 2018-05-03
Resolving issue with 3D hit creation using fully cheated reconstruction

StevenGreen1:
Issue relates to accessing geometry information in the case of multiple LArTPCs in a single pandora instance.

This reconstructs the ci test event fully including 3D hits.

### DIFF
--- a/larpandoracontent/LArHelpers/LArGeometryHelper.cc
+++ b/larpandoracontent/LArHelpers/LArGeometryHelper.cc
@@ -331,35 +331,6 @@ CartesianVector LArGeometryHelper::ProjectDirection(const Pandora &pandora, cons
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-float LArGeometryHelper::GetWireZPitch(const Pandora &pandora, const float maxWirePitchWDiscrepancy)
-{
-    const LArTPCMap &larTPCMap(pandora.GetGeometry()->GetLArTPCMap());
-
-    if (larTPCMap.empty())
-    {
-        std::cout << "LArGeometryHelper::GetWireZPitch - LArTPC description not registered with Pandora as required " << std::endl;
-        throw StatusCodeException(STATUS_CODE_NOT_INITIALIZED);
-    }
-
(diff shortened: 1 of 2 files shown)


larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -331,31 +331,64 @@ CartesianVector LArGeometryHelper::ProjectDirection(const Pandora &pandora, cons
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-float LArGeometryHelper::GetWireZPitch(const Pandora &pandora)
+float LArGeometryHelper::GetWireZPitch(const Pandora &pandora, const float maxWirePitchWDiscrepancy)
```
PandoraPFA: Should just make this function call the second function with specified view TPC_VIEW_W. This kind of function is then typically inline in Pandora.

Note that this is the kind of operation where we get into trouble when w no longer necessarily is equivalent to z, as Andy mentioned in his e-mail yesterday.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -331,31 +331,64 @@ CartesianVector LArGeometryHelper::ProjectDirection(const Pandora &pandora, cons
...
+    const LArTPC *const pFirstLArTPC(larTPCMap.begin()->second);
+    const float wirePitchW(pFirstLArTPC->GetWirePitchW());
+
+    for (const LArTPCMap::value_type &mapEntry : larTPCMap)
+    {
+        const LArTPC *const pLArTPC(mapEntry.second);
+
+        if (std::fabs(wirePitchW - pLArTPC->GetWirePitchW()) > maxWirePitchWDiscrepancy)
+        {
+            std::cout << "LArGeometryHelper::GetWireZPitch - Plugin does not support provided LArTPC configurations " << std::endl;
```
PandoraPFA: Pedantic. Does it make sense for this message to use the term Plugin? Maybe. Probably not quite.

PR #56, opened 2018-05-30
Added vertices for cheated (non-neutrino) particles.

StevenGreen1:
This fix allows Track and Shower objects to be made from cheated test beam particles.  Currently, cheated cosmic rays and test beam particles cannot be used to make tracks or showers inside LArSoft as there is no vertex made when cheating.  

This change consists of a configurable option to add a start vertex to all PFOs made in the LArCheatingPfoCreation algorithm.  If this flag is set to true in the cosmic ray pass of the LArCheatingPfoCreation algorithm, start vertices are added to all cosmic rays and test beam particles, which then allows them to be made into tracks and showers.

There are other options available to us on how we could proceed with this fix, such as only adding start vertices to test beam particles (as the start vertex for a cosmic ray could be ambiguous), which we should discuss before merging this request.  This fix has been tested inside LArSoft and does work for single electron showers.

### DIFF
--- a/larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
+++ b/larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
@@ -20,6 +20,7 @@ namespace lar_content
 CheatingPfoCreationAlgorithm::CheatingPfoCreationAlgorithm() :
     m_collapseToPrimaryMCParticles(false),
     m_useOnlyAvailableClusters(true),
+    m_addVertices(false),
     m_minGoodHitTypes(0),
     m_nHitsForGoodHitType(10)
 {
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
 
             const ParticleFlowObject *pPfo(nullptr);
             PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ParticleFlowObject::Create(*this, pfoParameters, pPfo));
+
+            if (m_addVertices)
+            {
+                const VertexList *pVertexList = NULL; std::string vertexListName;
+                PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::CreateTemporaryListAndSetCurrent(*this, pVertexList, vertexListName));
+
+                PandoraContentApi::Vertex::Parameters parameters;
+                parameters.m_position = CartesianVector(pMCParticle->GetVertex().GetX(), pMCParticle->GetVertex().GetY(), pMCParticle->GetVertex().GetZ());
+                parameters.m_vertexLabel = VERTEX_START;
+                parameters.m_vertexType = VERTEX_3D;
+
+                const Vertex *pVertex(NULL);
+                PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::Vertex::Create(*this, parameters, pVertex));
+
+                if (!pVertexList->empty())
+                {
+                    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::SaveList<Vertex>(*this, m_vertexListName));
+                    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ReplaceCurrentList<Vertex>(*this, m_vertexListName));
+                    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::AddToPfo<Vertex>(*this, pPfo, pVertex));
+                }
+            }
         }
         catch (const StatusCodeException &)
         {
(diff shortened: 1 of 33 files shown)


larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
```diff
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
 
             const ParticleFlowObject *pPfo(nullptr);
             PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ParticleFlowObject::Create(*this, pfoParameters, pPfo));
+
+            if (m_addVertices)
+            {
+                const VertexList *pVertexList = NULL; std::string vertexListName;
```
PandoraPFA: `nullptr` now preferred (has a pointer type), and via copy constructor.

larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
```diff
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
...
+                PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::CreateTemporaryListAndSetCurrent(*this, pVertexList, vertexListName));
+
+                PandoraContentApi::Vertex::Parameters parameters;
+                parameters.m_position = CartesianVector(pMCParticle->GetVertex().GetX(), pMCParticle->GetVertex().GetY(), pMCParticle->GetVertex().GetZ());
```
PandoraPFA: Can just assign `pMCParticle->GetVertex()` directly.

larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
```diff
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
...
+                parameters.m_vertexLabel = VERTEX_START;
+                parameters.m_vertexType = VERTEX_3D;
+
+                const Vertex *pVertex(NULL);
```
PandoraPFA: `nullptr`

larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
```diff
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
 
             const ParticleFlowObject *pPfo(nullptr);
             PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ParticleFlowObject::Create(*this, pfoParameters, pPfo));
+
+            if (m_addVertices)
+            {
+                const VertexList *pVertexList = NULL; std::string vertexListName;
+                PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::CreateTemporaryListAndSetCurrent(*this, pVertexList, vertexListName));
```
PandoraPFA: Probably better to just create the temporary list once, as for the Pfos, above. Then fill it as we build Pfos and save it at the end.

larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
```diff
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
...
+
+                PandoraContentApi::Vertex::Parameters parameters;
+                parameters.m_position = CartesianVector(pMCParticle->GetVertex().GetX(), pMCParticle->GetVertex().GetY(), pMCParticle->GetVertex().GetZ());
+                parameters.m_vertexLabel = VERTEX_START;
+                parameters.m_vertexType = VERTEX_3D;
+
+                const Vertex *pVertex(NULL);
+                PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::Vertex::Create(*this, parameters, pVertex));
+
+                if (!pVertexList->empty())
```
PandoraPFA: Suggest saving all vertices in one go, at end, just as for Pfos.

Can still add Vertex to Pfo at this point, but an exception will be raised at the end of the algorithm if you don't save a vertex that has been added to a Pfo. Pandora will attempt to delete the temporary Vertex, and find it is in use!

larpandoracontent/LArCheating/CheatingPfoCreationAlgorithm.cc
```diff
@@ -129,6 +130,27 @@ void CheatingPfoCreationAlgorithm::CreatePfos(const MCParticleToClusterListMap &
...
+                {
+                    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::SaveList<Vertex>(*this, m_vertexListName));
+                    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ReplaceCurrentList<Vertex>(*this, m_vertexListName));
```
PandoraPFA: Might want to add option as to whether you really want to replace the current list. Basically, and as for most of the above, suggest that you just follow what is already done for the Pfos.

StevenGreen1:
Thanks for the comments, I've implemented the changes, but would appreciate another quick check just to make sure I understood what you meant with the vertex list.  

I tested this out on an event and it seemed to work as expected (vertices produced for everything with an option of whether to replace the current list or not).  Also as we discussed I have changed the default behaviour to be to create the vertices for the PFO, but to not replace the current vertex list.  

I believe this means anyone wanting to build Tracks and Showers in LArSoft using cheated PFParticles won't have to alter PandoraSettings_Cheat.xml too, which is nice.

PandoraPFA:
All good - merged via commandline.

PR #62, opened 2018-06-14
Set the test beam score in the pfo properties map.

StevenGreen1:
This commit fills the IsTestBeam property in PfoProperties map for all parent pfos that go through the test beam particle selection (either bdt or cut based).  

In the cut based approach all scores are set to +1 if they pass the cuts and -1 if they don't.  This has been checked also for the simple use cases of selecting all slices as test beam particles and just selecting the first slice as a test beam particle.  It seemed appropriate to add a score for the cut based id to prevent a discontinuity in what is contained in the properties map for any larsoft analysers comparing new and old beam particle ids (not that I think this is a study anyone will be doing any time soon, if ever though).

For the bdt id the bdt scores are passed into the map.  These scores are confined between -1 and +1, hence the use of +1 and -1 for test beam and cosmic ray scores respectively for the cut based approach.  This has been tested for both the default use case and in training mode where everything is defaulted to a comsic ray.  The SelectAllPfos function in the BdtBeamParticleId class implicitly assumes you're selecting all Pfos as cosmic rays, hence giving a score of -1, but this is only used if running in the training mode where everything is being set to cosmic rays.   Are you happy with this or would you like me to add further constrains on this function to allow you to define a nominal score? 

A minor addition to the TestBeamParticleCreation algorithm was required as the test beam score is attached to the neutrino pfo and this needs to e persisted into the pfo that becomes the parent in the test beam (previously a daughter of the neutrino and so it doesn't have the score attached).

### DIFF
--- a/larpandoracontent/LArControlFlow/BdtBeamParticleIdTool.cc
+++ b/larpandoracontent/LArControlFlow/BdtBeamParticleIdTool.cc
@@ -87,7 +87,7 @@ void BdtBeamParticleIdTool::SelectOutputPfos(const pandora::Algorithm *const pAl
     if (m_useTrainingMode)
     {
         // ATTN in training mode, just return everything as a cosmic-ray
-        this->SelectAllPfos(crSliceHypotheses, selectedPfos);
+        this->SelectAllPfos(pAlgorithm, crSliceHypotheses, selectedPfos);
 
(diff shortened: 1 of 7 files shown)


PandoraPFA:
Merged via command line.

PR #65, opened 2018-06-15
Altering method for finding intercepts between fit and TPC boundaries.

StevenGreen1:
This method applies a limit on the isContained function and then selected the two intercepts that are furthest apart from each other.  I'm currently testing, but so far there are no changes in the results apart from the events that previously threw exceptions now no longer do.


PandoraPFA:
Merged.

PR #69, opened 2018-07-26
Vertex selection now uses GetInputEnergy

jackanth:
Vertex selection now uses `GetInputEnergy` instead of `GetElectromagneticEnergy`.


PandoraPFA:
Decision is to retrain vertexing using new samples with corrected input energy estimators, rather than work around legacy position in C++.

PR #71, opened 2018-08-17
Added method to get the MCParticle best matched to a PFO

jackanth:
Unlike `LArMCParticleHelper::GetMainMCParticle`, there's no need for agreement in multiple views. Every 2D hit in each view votes with a float between 0 and 1 (from their MC particle weight map), then the MC particle with the highest weight is chosen. The method should return `nullptr` iff the MC particle weight map is empty for every considered hit.

A variation on this could be weighting each hit's vote by its energy.

This is simplified from similar logic in my `LArPhysicsContent` repo.

### DIFF
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -216,6 +216,42 @@ const MCParticle *LArMCParticleHelper::GetMainMCParticle(const ParticleFlowObjec
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+const MCParticle *LArMCParticleHelper::GetBestMCParticle(const ParticleFlowObject *const pPfo)
+{
+    CaloHitList caloHitList;
+    LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_U, caloHitList);
+    LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_V, caloHitList);
+    LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_W, caloHitList);
+
+    const MCParticleWeightMap mcParticleWeightMap(GetMCParticleWeightMap(caloHitList));
+
+    float             bestWeight(0.f);
+    const MCParticle *pBestMCParticle(nullptr);
+
+    // Iterate through map in a reproducible order
+    MCParticleVector mcParticleVector;
+
+    for (const MCParticleWeightMap::value_type &mapEntry : mcParticleWeightMap)
+        mcParticleVector.push_back(mapEntry.first);
+
+    std::sort(mcParticleVector.begin(), mcParticleVector.end(), PointerLessThan<MCParticle>());
+
+    for (const MCParticle *const pMCParticle : mcParticleVector)
+    {
+        const float currentWeight(mcParticleWeightMap.at(pMCParticle));
+
+        if (currentWeight > bestWeight)
+        {
+            pBestMCParticle = pMCParticle;
+            bestWeight      = currentWeight;
+        }
+    }
+
+    return pBestMCParticle; // can be nullptr
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
 bool LArMCParticleHelper::SortByMomentum(const MCParticle *const pLhs, const MCParticle *const pRhs)
 {
     // Sort by momentum (prefer higher momentum)
@@ -602,6 +638,39 @@ CaloHitList LArMCParticleHelper::GetSharedHits(const CaloHitList &hitListA, cons
     }
 
     return sharedHits;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+MCParticleWeightMap LArMCParticleHelper::GetMCParticleWeightMap(const CaloHitList &caloHitList)
+{
+    MCParticleWeightMap mcParticleWeightMap;
+
+    for (const CaloHit *const pCaloHit : caloHitList)
+    {
+        // Synthesize the weights from every hit into the main map
+        const MCParticleWeightMap &hitMCWeightMap(pCaloHit->GetMCParticleWeightMap());
+        MCParticleVector           mcParticleVector;
+
+        // Iterate through map in a reproducible order
+        for (const MCParticleWeightMap::value_type &mapEntry : hitMCWeightMap)
+            mcParticleVector.push_back(mapEntry.first);
+
+        std::sort(mcParticleVector.begin(), mcParticleVector.end(), PointerLessThan<MCParticle>());
+
+        for (const MCParticle *const pMCParticle : mcParticleVector)
+        {
+            const auto findIter = mcParticleWeightMap.find(pMCParticle);
+
+            if (findIter == mcParticleWeightMap.end())
+                mcParticleWeightMap.emplace(pMCParticle, hitMCWeightMap.at(pMCParticle));
+
+            else
+                findIter->second += hitMCWeightMap.at(pMCParticle);
+        }
+    }
+
+    return mcParticleWeightMap;
 }
 
 } // namespace lar_content
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -164,6 +164,15 @@ public:
     static const pandora::MCParticle *GetMainMCParticle(const pandora::ParticleFlowObject *const pPfo);
 
     /**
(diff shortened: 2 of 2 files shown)


StevenGreen1:
Hi Jack, this code looks good to me.  There's two stylistic things that we might want to consider changing.  Firstly, there's a lot of unusual whitespace indentation that doesn't fit the typical Pandora style so fixing that would be great.  Secondly, there's a place where you've used auto (in place of MCParticleWeightMap::iterator) and another place where you fully specified a MCParticleWeightMap::value_type, it would be good to just pick one and use it consistently for a commit.  Finally, have you run this over some events to check everything works as expected and if so is there a plot/figure/comment that we could add to the PR to show things are as expected?  Thanks.

PandoraPFA:
Hi, I'll add some comments later this afternoon. Please still feel free to also add your thoughts before or after I do @loressa @astblake @a-d-smith

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -216,6 +216,42 @@ const MCParticle *LArMCParticleHelper::GetMainMCParticle(const ParticleFlowObjec
...
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+const MCParticle *LArMCParticleHelper::GetBestMCParticle(const ParticleFlowObject *const pPfo)
+{
+    CaloHitList caloHitList;
+    LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_U, caloHitList);
+    LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_V, caloHitList);
+    LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_W, caloHitList);
+
+    const MCParticleWeightMap mcParticleWeightMap(GetMCParticleWeightMap(caloHitList));
```
PandoraPFA: Is it better to create an empty MCParticleWeightMap locally, then pass it by non const reference to the function, which fills it? Or some equivalent, using move semantics.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -216,6 +216,42 @@ const MCParticle *LArMCParticleHelper::GetMainMCParticle(const ParticleFlowObjec
...
+    std::sort(mcParticleVector.begin(), mcParticleVector.end(), PointerLessThan<MCParticle>());
+
+    for (const MCParticle *const pMCParticle : mcParticleVector)
+    {
+        const float currentWeight(mcParticleWeightMap.at(pMCParticle));
+
+        if (currentWeight > bestWeight)
+        {
+            pBestMCParticle = pMCParticle;
+            bestWeight      = currentWeight;
```
PandoraPFA: Whitespace!

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -216,6 +216,42 @@ const MCParticle *LArMCParticleHelper::GetMainMCParticle(const ParticleFlowObjec
...
+        }
+    }
+
+    return pBestMCParticle; // can be nullptr
```
PandoraPFA: I think our convention is to raise a StatusCodeException, with STATUS_CODE_NOT_FOUND, rather than asking the algorithm to deal with a nullptr. Should try to make sure this approach is adopted throughout, rather than risking dereferencing nullptr one day!

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -604,4 +640,37 @@ CaloHitList LArMCParticleHelper::GetSharedHits(const CaloHitList &hitListA, cons
...
+
+MCParticleWeightMap LArMCParticleHelper::GetMCParticleWeightMap(const CaloHitList &caloHitList)
+{
+    MCParticleWeightMap mcParticleWeightMap;
+
+    for (const CaloHit *const pCaloHit : caloHitList)
+    {
+        // Synthesize the weights from every hit into the main map
+        const MCParticleWeightMap &hitMCWeightMap(pCaloHit->GetMCParticleWeightMap());
+        MCParticleVector           mcParticleVector;
```
PandoraPFA: Whitespace!

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -604,4 +640,37 @@ CaloHitList LArMCParticleHelper::GetSharedHits(const CaloHitList &hitListA, cons
...
+
+        // Iterate through map in a reproducible order
+        for (const MCParticleWeightMap::value_type &mapEntry : hitMCWeightMap)
+            mcParticleVector.push_back(mapEntry.first);
+
+        std::sort(mcParticleVector.begin(), mcParticleVector.end(), PointerLessThan<MCParticle>());
```
PandoraPFA: See above comment.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -604,4 +640,37 @@ CaloHitList LArMCParticleHelper::GetSharedHits(const CaloHitList &hitListA, cons
...
+
+        for (const MCParticle *const pMCParticle : mcParticleVector)
+        {
+            const auto findIter = mcParticleWeightMap.find(pMCParticle);
+
+            if (findIter == mcParticleWeightMap.end())
+                mcParticleWeightMap.emplace(pMCParticle, hitMCWeightMap.at(pMCParticle));
+
```
PandoraPFA: Whitespace, plus normally do use parentheses for if/else statements.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -604,4 +640,37 @@ CaloHitList LArMCParticleHelper::GetSharedHits(const CaloHitList &hitListA, cons
...
+        // Iterate through map in a reproducible order
+        for (const MCParticleWeightMap::value_type &mapEntry : hitMCWeightMap)
+            mcParticleVector.push_back(mapEntry.first);
+
+        std::sort(mcParticleVector.begin(), mcParticleVector.end(), PointerLessThan<MCParticle>());
+
+        for (const MCParticle *const pMCParticle : mcParticleVector)
+        {
+            const auto findIter = mcParticleWeightMap.find(pMCParticle);
```
PandoraPFA: Our use of auto is still a little inconsistent. This is indeed OK, but consistent use (or selective use, or even disuse) is something to think about in general.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -163,6 +163,15 @@ class LArMCParticleHelper
      */
     static const pandora::MCParticle *GetMainMCParticle(const pandora::ParticleFlowObject *const pPfo);
 
+    /**
+     *  @brief  Find the best mc particle making the largest contribution to 2D hits in a specified pfo
```
PandoraPFA: Find vs. Get?

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -306,6 +315,15 @@ class LArMCParticleHelper
      *  @return The hits that are found in both hitListA and hitListB
      */
     static pandora::CaloHitList GetSharedHits(const pandora::CaloHitList &hitListA, const pandora::CaloHitList &hitListB);
+
+    /**
+     *  @brief  Get the MCParticle weight map for a set of CaloHits
```
PandoraPFA: set vs. list?

StevenGreen1:
One other thing I noticed here is that you don't consider IsolatedCaloHits in these new functions.  Is this something you purposefully decided to do or does this needed to be included too?

PandoraPFA:
Ah, yes - thanks. I forgot to mention this above. I think the isolated hits are probably ignored in nearly everything apart from the very low-level validation that went into the paper studies. This is a problem (as I always knew isolated hits would be!).

It ultimately needs a search through the codebase, looking for cases where Clusters are used to fill CaloHitLists, and reviewing whether isolated hits should be added too. Hopefully this search would find most usages, although there may be some cases where algorithms manually extract the hits.

It's not a big problem (isolated hits only appear at the very end of the neutrino reco chain; not cosmics and not slicing), but it is a problem and one that will eventually cause someone a small headache. We should add an issue about this, so that it's ultimately followed up.

I don't think it will take long to sort through, but it will play a part in the subsequent release, etc.

PandoraPFA:
Old PR now superseded, closed.

PR #72, opened 2018-09-06
Added IsClearCosmic label to metadata for stitched cosmic rays.

StevenGreen1:


PandoraPFA:
Hi Steve,

Out of interest, does this have an important impact/function downstream?

This should work, but the overall metadata addition is now verbose and presumably repetitive. This is because of the way this flag is set, just above, for all `nonStitchedParentCosmicRayPfos` (to 1 or 0 as required). It's then set again (to 1) for all Pfos in the `clearCosmicRayPfos` list, which will presumably include many of those for which it's already 1?

I was therefore wondering if you might be better just removing the two current attempts to set this metadata, and just adding a new loop over all the Pfos in the `pRecreatedCRPfos` list, checking whether they also exist in the `ambiguousPfos` list, if you think that makes sense?

It's a balance between making the code nice and simple and easier to maintain in future, and needing to do exhaustive checks because tidying up the code accidentally makes a change in behaviour. It is particularly important that the Master algorithm remains easy to follow, though.

Just my two-pence worth, anyhow.

StevenGreen1:
Hi John,

Yes it does for Andy's external event building code in LArPandora.  It selects the PFParticle collections that are tagged as IsClearCosmic from the slicing instance and adds them together with the TestBeam/Nu/AmbiguousCRs from the Nu/CR workers.  Therefore, without this tag the stitched cosmic rays don't appear in the externally built events, which is what's been causing difficulties for him testing it on ProtoDUNE.  It works fine for MicroBooNE as there's no stitching.  

That's a good point about the code, I didn't think about the duplicates, I'll tidy the for loops up a bit now! 

Cheers
Steve

PandoraPFA:
Hi Steve,

Thanks. I think the `clearCosmics` list now only exists for the purposes of visualisation and so its creation should go inside the later `m_visualizeOverallRecoStatus` block.

Others should ideally look over this too (I only had a look as a quick break from writing slides about helicity and chirality!).

StevenGreen1:
Hi John,

Oh yeah you are right, there's no need for that function to read in the, initially empty, clearCosmicRayPfos PFOList, populate it and return it at all as it isn't used outside of this function.  Is there a reason for returning the clearCosmicRayPfos list or shall I remove this argument and just create a local PFOList in the function?  Either way I'll finish tidying this function and test it tomorrow.

PandoraPFA:
I'd forgotten the list was actually created outside of the function. Maybe we'd best leave this in for now, as you can bet that one day there will be a use for it.

It is, conceptually, the obvious interface for the function. That would mean that, we two think the PR is now ready. Would be good for someone else to check too.

The common lecturer trick seems to be to confine yourself to the regime where helicity and chirality eigenstates are equivalent, and then go "oooooh, you need to be careful, though" :)

StevenGreen1:
Hi John,

Yeah I think that sounds good.  It is redundant for the moment, but not wasteful and retaining the nicer interface might come in handy later.  I've got an event dump for a clean event and the same event build with all outcomes turned on and reconstructed using Andy's external event building algorithm and they are the same modulo changes in ID and Key number, which is good.  I think we should let @loressa have the final review of this once she's back next week and then we can merge.

Cheers

PR #73, opened 2018-11-14
Resolve compiler error on MacOS

StevenGreen1:
Refactoring bdt beam particle id class to resolve compiler error on MacOS.


StevenGreen1:
I tested this by running a few (~10) ProtoDUNE events through with and without the change and there were no differences.  As you can see it's pretty minimal, but makes Pandora work on the new MacOS, so it's worth putting the change in.  If someone else could take a quick look to sanctify check this it would be appreciated.

Cheers, Steve

PandoraPFA:
Thanks. This can presumably go in alongside a next v03.15.0 release, or distributed specifically in a quiet week for a v03.14.03 release (should be able to push to develop). Have a think and proceed as you feel is best.

StevenGreen1:
I'll have a think about releases at some point soon, as we also have a few other things queued up so might be easiest to sneak this in alongside.  Will double check the plan with everyone once we have it!

PR #74, opened 2018-11-28
Updated numbering convention for slice indexes to match how output to…

loressa:
… LArSoft is handled in larpandora


PandoraPFA:
Looks exactly as we discussed via e-mail - Many thanks!

PandoraPFA:
Thanks for the test output files! We should consider using a PR template soon, so that testing status is one of the items addressed in initial comments when submitting the PR. It's good practice to provide accompanying info anyway, but this would provide a handy reminder when opening at PR. It would, however, mean more .md files in the repo, which we could maybe put in a .github directory.

PR #75, opened 2018-11-30
APA Stitching

StevenGreen1:
Modification to sign of x0 to allow for stitching across APA and CPA boundaries.  Below are screenshots of stitching demonstrating that stitching now works as expected across CPAs and APAs.

screenshot 2018-11-30 at 10 57 48
screenshot 2018-11-30 at 10 57 32
screenshot 2018-11-30 at 10 57 23

### DIFF
--- a/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
+++ b/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
@@ -532,13 +532,18 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
             {
                 if (pfoToLArTPCMap.find(pPfoToShift) != pfoToLArTPCMap.end())
                 {
+                    const LArTPC *const pLArTPC(pfoToLArTPCMap.at(pPfoToShift));
+
+                    if (pLArTPC && (stitchedLArTPCs.first == pLArTPC || stitchedLArTPCs.second == pLArTPC))
+                        continue;
+
                     if (!stitchedLArTPCs.first)
                     {
-                        stitchedLArTPCs.first = pfoToLArTPCMap.at(pPfoToShift);
+                        stitchedLArTPCs.first = pLArTPC;
                     }
                     else if (!stitchedLArTPCs.second)
                     {
-                        stitchedLArTPCs.second = pfoToLArTPCMap.at(pPfoToShift);
+                        stitchedLArTPCs.second = pLArTPC;
                     }
                     else
                     {
--- a/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
+++ b/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
@@ -524,6 +524,44 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
         for (const ParticleFlowObject *const pPfo : pfoList) pfoVector.push_back(pPfo);
         std::sort(pfoVector.begin(), pfoVector.end(), LArPfoHelper::SortByNHits);
 
+        std::pair<const LArTPC*, const LArTPC*> stitchedLArTPCs;
+
+        try
+        {
+            for (const ParticleFlowObject *const pPfoToShift : pfoVector)
+            {
+                if (pfoToLArTPCMap.find(pPfoToShift) != pfoToLArTPCMap.end())
+                {
+                    if (!stitchedLArTPCs.first)
+                    {
+                        stitchedLArTPCs.first = pfoToLArTPCMap.at(pPfoToShift);
+                    }
+                    else if (!stitchedLArTPCs.second)
+                    {
+                        stitchedLArTPCs.second = pfoToLArTPCMap.at(pPfoToShift);
+                    }
+                    else
+                    {
+                        throw StatusCodeException(STATUS_CODE_FAILURE);
+                    }
+                }
+                else
+                {
+                    throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+                }
+            }
+        }
+        catch (const pandora::StatusCodeException &statusCodeException)
+        {
+            if (STATUS_CODE_FAILURE == statusCodeException.GetStatusCode())
+                std::cout << "StitchingCosmicRayMergingTool: Attempting to stitch a pfo across more than two LArTPCs" << std::endl;
+
+            if (STATUS_CODE_NOT_FOUND == statusCodeException.GetStatusCode())
+                std::cout << "StitchingCosmicRayMergingTool: Unable to find LArTPC for a pfo" << std::endl;
+
+            continue;
+        }
+
         float x0(0.f);
 
         if (!m_useXcoordinate)
@@ -538,7 +576,16 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
             }
 
             for (const ParticleFlowObject *const pPfoToShift : pfoVector)
-                pAlgorithm->ShiftPfoHierarchy(pPfoToShift, pfoToLArTPCMap, x0);
+            {
+                if (pfoToLArTPCMap.find(pPfoToShift) != pfoToLArTPCMap.end())
+                {
+                    const LArTPC *const pActiveLArTPC(pfoToLArTPCMap.at(pPfoToShift));
+                    const LArTPC *const pPartnerLArTPC(stitchedLArTPCs.first == pActiveLArTPC ? stitchedLArTPCs.second : stitchedLArTPCs.first);
+                    const float sign(pActiveLArTPC->GetCenterX() < pPartnerLArTPC->GetCenterX() ? 1.f : -1.f);
+                    x0 = std::fabs(x0) * sign;
+                    pAlgorithm->ShiftPfoHierarchy(pPfoToShift, pfoToLArTPCMap, x0);
+                }
+            }
         }
 
         for (const ParticleFlowObject *const pPfoToDelete : pfoVector)
--- a/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
+++ b/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
@@ -542,6 +542,7 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
         }
 
         float x0(0.f);
+        const float tpcBoundaryCenterX(LArStitchingHelper::GetTPCBoundaryCenterX(*stitchedLArTPCs.first, *stitchedLArTPCs.second));
 
         if (!m_useXcoordinate)
         {
@@ -556,16 +557,25 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
 
             for (const ParticleFlowObject *const pPfoToShift : pfoVector)
             {
-                PfoToLArTPCMap::const_iterator iter(pfoToLArTPCMap.find(pPfoToShift));
+                CaloHitList caloHitList3D;
+                LArPfoHelper::GetCaloHits(pPfoToShift, TPC_3D, caloHitList3D);
 
-                if (iter != pfoToLArTPCMap.end())
+                float xsum(0.f);
+                int count(0);
+
+                for (const CaloHit *const pCaloHit : caloHitList3D)
                 {
-                    const LArTPC *const pActiveLArTPC(iter->second);
-                    const LArTPC *const pPartnerLArTPC(stitchedLArTPCs.first == pActiveLArTPC ? stitchedLArTPCs.second : stitchedLArTPCs.first);
-                    const float sign(pActiveLArTPC->GetCenterX() < pPartnerLArTPC->GetCenterX() ? 1.f : -1.f);
-                    x0 = std::fabs(x0) * sign;
-                    pAlgorithm->ShiftPfoHierarchy(pPfoToShift, pfoToLArTPCMap, x0);
+                    xsum += pCaloHit->GetPositionVector().GetX();
+                    ++count;
                 }
+
+                if (count == 0)
+                    continue;
+
+                const float averageX(xsum / static_cast<float>(count));
+                const float sign(averageX < tpcBoundaryCenterX ? 1.f : -1.f);
+                x0 = std::fabs(x0) * sign;
+                pAlgorithm->ShiftPfoHierarchy(pPfoToShift, pfoToLArTPCMap, x0);
             }
         }
 
--- a/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
+++ b/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
@@ -663,7 +663,7 @@ void StitchingCosmicRayMergingTool::CalculateX0(const PfoToLArTPCMap &pfoToLArTP
                 PfoToPointingVertexMap::iterator pfoToPointingVertexMapIter1(pfoToPointingVertexMap.find(pPfo1));
                 if (pfoToPointingVertexMapIter1 == pfoToPointingVertexMap.end())
                 {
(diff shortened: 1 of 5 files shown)


StevenGreen1:
There are a few places in this code where a map is accessed without checking the object exists.  I've added a try/catch block around where I do this in the first instance.  I repeat this accessing later, but I don't use the try/catch as I'm running the same loop as in the first instance so all errors would have been caught.  If anyone has a more elegant solution to this then I'd be happy to implement it.

StevenGreen1:
MCC10 ProtoDUNE reco metrics for this branch.
beamparticlecompleteness_5_gev_beam_cosmics
beamparticleefficiencyvsnhits_5_gev_beam_cosmics
beamparticlepurity_5_gev_beam_cosmics

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -524,6 +524,49 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
         for (const ParticleFlowObject *const pPfo : pfoList) pfoVector.push_back(pPfo);
         std::sort(pfoVector.begin(), pfoVector.end(), LArPfoHelper::SortByNHits);
 
+        std::pair<const LArTPC*, const LArTPC*> stitchedLArTPCs;
```
PandoraPFA: Do make sure the values in the pair are explicitly initialised (you rely on this below), e.g. via the constructor.

Typically make this a typedef, here probably a private typedef defined in the class prototype. Can here then make an instance via the more self-describing e.g. `LArTPCPair stitchedLArTPCs(nullptr, nullptr);`

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -524,6 +524,49 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+
+        try
```
PandoraPFA: Suggest take this entire block and make it into a private member function, which takes a non-const reference to the LArTPCPair and sets the values in the pair.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -524,6 +524,49 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+        {
+            for (const ParticleFlowObject *const pPfoToShift : pfoVector)
+            {
+                if (pfoToLArTPCMap.find(pPfoToShift) != pfoToLArTPCMap.end())
```
PandoraPFA: Code is nice and clean by checking to see if pfoToShift is there, then using `.at(...)` below. I typically use `.count(...)` to check if something is there, as it reads a bit more cleanly than checking whether you got a valid iterator.

More efficient than doing the map lookup twice, though, is to actually store the iterator returned by `.find(...)` and then using it to get the LArTPC from the map too. i.e. only doing the look up once.

Probably doesn't matter much here, as the map is small, but when they get big this is important.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -524,6 +524,49 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+        }
+        catch (const pandora::StatusCodeException &statusCodeException)
+        {
+            if (STATUS_CODE_FAILURE == statusCodeException.GetStatusCode())
+                std::cout << "StitchingCosmicRayMergingTool: Attempting to stitch a pfo across more than two LArTPCs" << std::endl;
+
+            if (STATUS_CODE_NOT_FOUND == statusCodeException.GetStatusCode())
+                std::cout << "StitchingCosmicRayMergingTool: Unable to find LArTPC for a pfo" << std::endl;
+
+            continue;
```
PandoraPFA: I think it looks like you do indeed want to continue, as you'll keep on trying to stitch pPfoToEnlarge to those Pfos in the pfoVector. (This kind of logic is often a source of problems, but looks good here!)

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -538,7 +581,16 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
             }
 
             for (const ParticleFlowObject *const pPfoToShift : pfoVector)
-                pAlgorithm->ShiftPfoHierarchy(pPfoToShift, pfoToLArTPCMap, x0);
+            {
+                if (pfoToLArTPCMap.find(pPfoToShift) != pfoToLArTPCMap.end())
```
PandoraPFA: (Can see earlier comment about map iterators)

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -538,7 +581,16 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+                {
+                    const LArTPC *const pActiveLArTPC(pfoToLArTPCMap.at(pPfoToShift));
+                    const LArTPC *const pPartnerLArTPC(stitchedLArTPCs.first == pActiveLArTPC ? stitchedLArTPCs.second : stitchedLArTPCs.first);
+                    const float sign(pActiveLArTPC->GetCenterX() < pPartnerLArTPC->GetCenterX() ? 1.f : -1.f);
```
PandoraPFA: Just looking for reassurance that you've checked this through, and thought about e.g. any cosmic ray reconstruction (rare, but required) that might need to happen in DUNE FD, with different APA, CPA structure.

Isn't it possible that you might want to move both tracks (in both TPCs) in the opposite sense in order to achieve a good stitch (so cosmic ray arrived after Tbeam, rather than before Tbeam)?

Or is there something somewhere else that precludes this? @astblake?

StevenGreen1: Hi John,
Thank you for the comments.  They were really useful and I've implemented them.  I think what I didn't particularly like was the try-catch block structure as I felt it was too big, but by moving the try block logic to a separate function it looks quite clean now.  

I think you might be right about there being an issue if the CPA/APA structure is different as this logic assumes that the 3D particle appears in the 3D drift volume it was measured in, when that might not be the case.  I will have more of a think about this and look at a few more events to check everything still makes sense.  Especially for DUNEFD too.  

Cheers
Steve

StevenGreen1:
I've looked into this a bit more (and hand scanned more events that I initially did) and I did find a problem in the logic.  I've assumed that the 3D particle reconstructed in each drift volume will be in the drift volume, which isn't always the case.  You can reconstruct a 3D particle in the adjacent drift volume, which means the sign needs to chance too.  I've got a few ideas on how to fix this and will make a new commit shortly.  Here's also a picture of an event that I found where this PR gives the wrong sign 
screenshot 2018-12-07 at 13 36 22

StevenGreen1:
Hi John,

I've now fixed the logic here so that the stitching moved the 3D pfo toward the boundary that is being stitched across.  I will rerun the metrics to make sure this doesn't change anything, but I'd appreciate having a quick check over the changes when you have time.

Here's a screenshot of this branch fixing the previous issue.  The other stitching examples shown above also work correctly.

screenshot 2018-12-10 at 14 24 35

Cheers
Steve

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.h
```diff
@@ -5,7 +5,7 @@
  *
  *  $Log: $
  */
-#ifndef LAR_STITCHING_COSMIC_RAY_MERGING_TOOL_H
+# ifndef LAR_STITCHING_COSMIC_RAY_MERGING_TOOL_H
```
PandoraPFA: A space crept in here - should undo this.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -538,7 +556,27 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+                for (const CaloHit *const pCaloHit : caloHitList3D)
+                {
+                    xsum += pCaloHit->GetPositionVector().GetX();
+                    ++count;
+                }
+
+                if (count == 0)
+                    continue;
+
+                const float averageX(xsum / static_cast<float>(count));
```
PandoraPFA: I would personally remove the variable `count` (just another thing that could go wrong), get `xSum` (should ideally have capital S!) in a for loop with a single line body, i.e. without parentheses and check `caloHitList3D.size()` (could do this check first, and cache value in a const unsigned int, and use it in the calculation of `averageX`).

I wondered whether this was just another way of characterising a cluster position (we probably have too many different ways to do this), but mean X is just so simple, we probably don't need to worry here.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -538,7 +556,27 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+                const float sign(averageX < tpcBoundaryCenterX ? 1.f : -1.f);
+                x0 = std::fabs(x0) * sign;
```
PandoraPFA: Yes, this seems like a neat way to do it. I think this works for all cases now, but if you can reply to say you've checked it through one more time, that'd be great!

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
StevenGreen1: Good idea, caching the int first and then checking the size is more efficient.  The xsum has now been changed to xSum, which I'm very happy to do, but I think I copied this from a helper (I think) so we might need to update that too in the future!

StevenGreen1:
Argh! This is the most fiddly negative sign change I've ever dealt with!  I found a case now where the centre of the pfo actually needs to move away from the boundary because it's a really long pfo and the mean position is actually on the opposite side to the pointing vertex being stitched! 

I will have another go at fixing this now.  Attached is an event display of the problem event.
 
screenshot 2018-12-11 at 10 08 27

PandoraPFA:
Hi Steve,

Great that you've caught this before it was merged (better that it's released right than released fast!). Ultimately, don't you want to use the same logic to sign the x0 shifts as was actually used to identify a good stitch in the first place? I think this is here:
https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc#L594-L597

If this is an expensive calculation, we could even think about storing the info you need at the point of the original calculation. This is, after all, where the x0 is originally found. Here, we're just trying to implement the shift, and so need to know which Pfo gets shifted in which direction.

Thanks!

John

StevenGreen1:
Hi John, 

Thanks, I'm glad we caught this too.  Those lines you pointed me to are where I'm trying to get the proper fix out.  The problem is that there's no limit on the number of Pfos being stitched at one time (i.e. can be more than two), therefore, finding the single vertex to use to find the relevant shift direction for a given pfo is tricky as there are lots of combinations to consider when shifting.  Also the x0 calculation is ambivalent to the shift sign meaning the information again is hard to extract.  This is a much more in depth problem that I'd initially expected and these complexities are the reason as to why I didn't just start with this (as the info is buried in that calculation).

I've got another idea on what to do now though and will test this again once I've got it working.  

Cheers
Steve

StevenGreen1:
Ok, so I think I'm there now.  I've added code that fills a map of pfo to the pointing vertex used in the stitching calculation.  If more than one pointing vertex is used for a single pfo, then we throw an exception because each Pfo should be stitched using only one vertex at any given instance (in such a case the x0 will be 0 and stitching will run but do nothing, this piggy backs on a catch that's already there).  Then I decide the sign of x0 based on the location of the vertex and demand that it gets closer to the tpc boundary.  I believe this is now the correct logic and it definitely fixes the issues raised above.  

I will run the metrics to double check again though!  

Cheers
Steve

StevenGreen1:
Have hand scanned the 10 events that I have with the APA stitching and all the stitching cases work properly.  Waiting for metrics on condor now.

StevenGreen1:
Hi, 

I think I've finally got this working properly now.  The difficulty comes with (i) having to find the relevant sign of the x0 shift for each PFO in opposite drift volumes, which I resolved by making sure the pointing vertices move toward the stitching boundary and (ii) introducing a signed X0 into the pfo particle metadata that has the correct sign for calculation of t0 in LArPandora.  Point (ii) is particularly challenging because pfos being stitched on either side of the drift volume are shifted by a +x0 and -x0 in their own right in 3D space, but when considered as a pair result in an t0 shift of the same sign that tells you whether the cosmic arrive before or after the beam.  

What I did to check this was working was to make distributions of the reco vs mc T0 (using my modified LArMCParticles).  As you can see this change doesn't affect the CPA stitches.  Unfortunately, we don't have MC data for the APAs so I can't confirm that they're totally working, but I've introduced the negative sign to account for the fact that a T_Cosmic > T_Beam in a CPA stitch looks like a T_Cosmic < T_Beam in an APA stitch and vice versa.  Also the signedT0 metadata calculation requires the stitch to be across an APA or CPA (and no funky combinations), however, the calculation of X0 actually enforces this earlier in the stitching calculation (in the LArStitchingHelper::CalculateX0).  Therefore, I suggest we go with this fix now so it works for ProtoDUNE and long term if anyone decides to build TPCs next to each other with drift volumes pointing in the same direction we can address that at a later date.

Also if there are no objections I would suggest removing the LArStitchingHelper::GetCorrectedPosition function as it is no longer valid for APA stitches and it isn't used in LArPandora or LArPandoraContent.  

deltat0cosmicrays_5_gev_beam_cosmics
recovsmct0cosmicrays_5_gev_beam_cosmics

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -594,10 +660,36 @@ void StitchingCosmicRayMergingTool::CalculateX0(const PfoToLArTPCMap &pfoToLArTP
                 LArStitchingHelper::GetClosestVertices(*pLArTPC1, *pLArTPC2, pointingCluster1, pointingCluster2,
                     pointingVertex1, pointingVertex2);
 
+                PfoToPointingVertexMap::iterator pfoToPointingVertexMapIter1(pfoToPointingVertexMap.find(pPfo1));
+                if (pfoToPointingVertexMapIter1 == pfoToPointingVertexMap.end())
+                {
+                    pfoToPointingVertexMap.insert(std::make_pair(pPfo1, pointingVertex1));
```
PandoraPFA: Cosmetic: For a map of type `MyMap`, I tend to use:
`myMap.insert(MyMap::value_type(key, value));`

The insertion also returns a pair, with the first element being an iterator to the inserted value (or existing value), and the second element being a Boolean indicating whether a new element was inserted (true), or whether an equivalent key already existed (false). I generally check the Boolean, but here it's not useful, as you know the map didn't already contain a key matching `pPfo1`, so in this case, I normally indicate that I'm throwing away the return value:
`(void) myMap.insert(MyMap::value_type(key, value));`

PandoraPFA:
Hi Steve,

I think everything's looking good now, to the extent that we're probably now beyond looking at the code and so we're relying on all your testing.

I agree that you should remove the redundant helper function (we'll remember that it was there, in case we need to extract it from the history again!). If you push a commit to do this, and drop me a message saying that you think we're good for the merge, we'll get it in and look to get the release to LArSoft going.

I think this is just a patch release, v03_14_04, or would you say that this should be a minor revision increment, to v03_15_00? This obviously goes into the Pandora-only bit of CMakeLists.txt, *not* in the ups bit (LArSoft need to do that!), and in an update to the ChangeLog, as usual.

Thanks,

John

PR #76, opened 2019-01-29
Adding option to always calculate T0 when stitching cosmics rays.

StevenGreen1:


StevenGreen1:
Here's a figure showing the effect of adding this option on the reconstructed T0 distribution for ProtoDUNE MCC11 (5 GeV test beam sample).

[Plot5.pdf](https://github.com/PandoraPFA/LArContent/files/2806870/Plot5.pdf)

PandoraPFA:
Perfect - thanks!

StevenGreen1:
Thanks for sorting that so quickly!

PR #77, opened 2019-02-05
Reset Hit Positions

StevenGreen1:
Post cosmic ray tagging, reset calo hit positions for ambiguous pfos that have been stitched.

### DIFF
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -18,6 +18,7 @@
 #include "larpandoracontent/LArHelpers/LArFileHelper.h"
 #include "larpandoracontent/LArHelpers/LArMCParticleHelper.h"
 #include "larpandoracontent/LArHelpers/LArPfoHelper.h"
+#include "larpandoracontent/LArHelpers/LArStitchingHelper.h"
 
 #include "larpandoracontent/LArObjects/LArCaloHit.h"
 #include "larpandoracontent/LArObjects/LArMCParticle.h"
@@ -420,7 +421,7 @@ StatusCode MasterAlgorithm::RunCosmicRayHitRemoval(const PfoList &ambiguousPfos)
         const VertexList vertexList(pPfoToDelete->GetVertexList());
 
         // ATTN: If an ambiguous pfo has been stitched, reset the calo hit positions in preparation for subsequent algorithm chains
-        if (pPfoToDelete->GetPropertiesMap().count("X0"))
+        if (LArStitchingHelper::HasPfoBeenStitched(pPfoToDelete))
         {
             PfoList downstreamPfos;
             LArPfoHelper::GetAllDownstreamPfos(pPfoToDelete, downstreamPfos);
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -423,16 +423,13 @@ StatusCode MasterAlgorithm::RunCosmicRayHitRemoval(const PfoList &ambiguousPfos)
         // ATTN: If an ambiguous pfo has been stitched, reset the calo hit positions in preparation for subsequent algorithm chains
         if (LArStitchingHelper::HasPfoBeenStitched(pPfoToDelete))
         {
-            PfoList downstreamPfos;
-            LArPfoHelper::GetAllDownstreamPfos(pPfoToDelete, downstreamPfos);
-
             CaloHitList caloHitList2D;
-            LArPfoHelper::GetCaloHits(downstreamPfos, TPC_VIEW_U, caloHitList2D);
-            LArPfoHelper::GetCaloHits(downstreamPfos, TPC_VIEW_V, caloHitList2D);
-            LArPfoHelper::GetCaloHits(downstreamPfos, TPC_VIEW_W, caloHitList2D);
-            LArPfoHelper::GetIsolatedCaloHits(downstreamPfos, TPC_VIEW_U, caloHitList2D);
-            LArPfoHelper::GetIsolatedCaloHits(downstreamPfos, TPC_VIEW_V, caloHitList2D);
-            LArPfoHelper::GetIsolatedCaloHits(downstreamPfos, TPC_VIEW_W, caloHitList2D);
+            LArPfoHelper::GetCaloHits(pPfoToDelete, TPC_VIEW_U, caloHitList2D);
+            LArPfoHelper::GetCaloHits(pPfoToDelete, TPC_VIEW_V, caloHitList2D);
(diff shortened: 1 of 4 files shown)


larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -418,6 +418,29 @@ StatusCode MasterAlgorithm::RunCosmicRayHitRemoval(const PfoList &ambiguousPfos)
     {
         const ClusterList clusterList(pPfoToDelete->GetClusterList());
         const VertexList vertexList(pPfoToDelete->GetVertexList());
+
```
PandoraPFA: I think this entire function is just about under the line limit for a single function. If the x0 reset gets any bigger, it could be extracted as a separate function, but I think it's OK right now.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -418,6 +418,29 @@ StatusCode MasterAlgorithm::RunCosmicRayHitRemoval(const PfoList &ambiguousPfos)
...
+        // ATTN: If an ambiguous pfo has been stitched, reset the calo hit positions in preparation for subsequent algorithm chains
+        if (pPfoToDelete->GetPropertiesMap().count("X0"))
```
PandoraPFA: Using the presence of a string in the map means this algorithm relies on the implementation in other parts of the code. I wonder whether we could instead go for a single definition in e.g. `LArPfoHelper::HasPfoBeenStitched(const Pfo *const);` and use this in *all* cases where we want to know this.

One way to avoid this would be to just look at the hits, but this means looping over all the hits, all the time. Alternatively, could push the actual string into xml at the point of adding the property and when checking it.

I think I'm in favour of consolidating this functionality into one place for reduced maintenance. What do you think?

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -418,6 +418,29 @@ StatusCode MasterAlgorithm::RunCosmicRayHitRemoval(const PfoList &ambiguousPfos)
...
+        {
+            PfoList downstreamPfos;
+            LArPfoHelper::GetAllDownstreamPfos(pPfoToDelete, downstreamPfos);
```
PandoraPFA: Are we doing this multiple times for some Pfos, given the call to `GetAllConnectedPfos` above, then `GetAllDownstreamPfos` here? Could use a PfoSet to keep track of which Pfos have been processed (but watch out, as these are unordered, so only insert when a Pfo has been processed, then look up to see whether to proceed with processing future Pfos).

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
StevenGreen1: Hi John, 

This is a good point, I like the idea of having the additional LArPfoHelper function as it will reduce complications in the future.  I'll start work on adding that now.  

I'm not sure I understand what you mean by pushing the string into xml at the point of adding the property and when checking it though.  Could you clarify as I'm probably just misunderstanding something obvious!

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
StevenGreen1: We don't actually do this multiple times because the X0 shift is only set for the parent cosmic ray and not the daughters.  This means that we need to reset the X0 position in the calo hits for the for the daughter pfos too when we encounter a pfo with an X0 because if we don't the daughter hits are still shifted.  

I believe the logic behind not setting the X0 for the daughters was that the shift is calculated from the parent track and so isn't a property of the daughters per say, but the daughter hits will be shifted because of it.  Perhaps it would be cleaner to set the x0 for the daughter pfos too as it is slightly confusing to shift the hits for the pfo, but not to set the x0.  What are your thoughts on this?  If you agree with that suggestions I'll add that another commit to this PR setting the X0 for the daughters.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
PandoraPFA: Hi Steve,

Sorry this wasn't clear (unless you're a mind reader!). Basically I'm rarely comfortable when we have literals, especially when we're relying on filling the map via a string literal key in one algorithm and then accessing it, via the same literal, in another. It can be difficult to maintain when changes are made in future.

One idea is to move the literal into just one place (or maybe one place for setting, one place for accessing) e.g. in a helper, to ease maintenance. Another idea would be to set and read the map using a string that is read in via an XML config file, to provide the key. This still means setting and accessing can get out of sync, but it can be recovered via config, rather than being irrevocably compiled in. I hope this clarifies things!

It looks like LArStitchingHelper would actually be the best place to consolidate this. If every use/check goes through this helper, then it's much easier to maintain.

What do you think?

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
PandoraPFA: OK, thanks for clarifying. I think you have most experience now with where/how this is used, so please go with the convention you think makes most sense and will help ease maintenance in the future. If you want to bounce ideas around about this via e-mail or next week then I'd be very happy to discuss. Thanks.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
StevenGreen1: Hi John,
Thanks for the feedback, sorry for the slow response.  Ah yes, now I understand about defining the string to be set via xml.  I'll add these additional functions to the Stitching helper for now to keep things under control as you suggest.  
Cheers

StevenGreen1:
Hi John, 

I've attached some plots showing the impact of this PR on the test beam and cosmic ray reconstruction metrics.  There is almost no change, but one test beam particle is lost and another two cosimcs are correctly identified.  I'll have a look at what's going on with this event, but I expect that it's due to that fact that in the updates we reset the hit positions for cosmics that are stitched and then aren't tagged as clear, which will lead to a small shift in calo hit positions.  

Cheers
Steve 

beamparticlecompleteness_5_gev_beam_cosmics
beamparticleefficiencyvsnhits_5_gev_beam_cosmics
beamparticlepurity_5_gev_beam_cosmics
cosmicraycompleteness_5_gev_beam_cosmics
cosmicrayefficiencyvsnhits_5_gev_beam_cosmics
cosmicraypurity_5_gev_beam_cosmics
cosmicrayt0
cosmicrayt0resolustion

StevenGreen1:
I've found the event that changed with this PR.  In the default sample, because the initial hits were shifted by stitching and not reset, there was a different outcome in the bdt id for this one event.  Running the default sample without the AlwaysApplyT0Calucaltion flag in the stitching gives exactly the same outcome as this new PR, so I believe this is good to be merged now.

PR #78, opened 2019-04-01
ProtoDUNE Hierarchy Mode

StevenGreen1:
Hi, 

This is the feature branch that I've developed to allow Pandora to produce metrics showing the reconstruction efficiency for the hierarchy in protoDUNE test beam events.  

Key Points: 

- The parent track is treated as an additional primary.
- The tier information is written out in addition to the other variables.
- Invisible particles are suppressed in the hierarchy (so that a pizero will appear as two gammas).
- The printout in default (neutrino) mode, should be identical and has been tested on a few MicroBooNE and DUNEFD events.  The normal test beam mode is identical apart from the test beam pfo id is printed out also.  
- The interaction type helper requires more options to be comprehensive.  ProtoDUNE hierarchy events can easily involve more than 10 particle types and the current method for labelling them requires all permutations to be coded in.  This might be an area where we need to change the method, however, for now I've added in the interaction types of interest (e.g. K0L, PiZeros and small particle permutations).

There are a few places where I think this code can be improved/shortened, but this does work (and John requested the PR once ready) and this branch will be extensively tested now (also used for Stefano's poster if time).

I will update this PR if changes are made to the branch based on the testing or if further feedback is requested.  (It should not be merged just yet but feedback is welcome!)

Cheers
Steve


larpandoracontent/LArHelpers/LArInteractionTypeHelper.cc
```diff
@@ -44,17 +44,130 @@ LArInteractionTypeHelper::InteractionType LArInteractionTypeHelper::GetInteracti
         else return COSMIC_RAY_OTHER;
     }
 
-    if ((1 == mcPrimaryList.size()) && LArMCParticleHelper::IsBeamParticle(mcPrimaryList.front()))
+    if (!testBeamHierarchyMode)
```
PandoraPFA: Suggest separate these two helpers into GetInteractionType and GetHierarchyInteractionType (or a better name), then use the algorithm logic to decide which to call. Part of a general pattern of teasing apart implementation.

larpandoracontent/LArHelpers/LArInteractionTypeHelper.cc
```diff
@@ -437,6 +550,44 @@ std::string LArInteractionTypeHelper::ToString(const InteractionType interaction
     case BEAM_PARTICLE_KAON_PLUS: return "BEAM_PARTICLE_KAON_PLUS";
     case BEAM_PARTICLE_KAON_MINUS: return "BEAM_PARTICLE_KAON_MINUS";
     case BEAM_PARTICLE_OTHER: return "BEAM_PARTICLE_OTHER";
+    case BEAM_PARTICLE_PI_PLUS_PI_PLUS: return "BEAM_PARTICLE_PI_PLUS_PI_PLUS";
```
PandoraPFA: Could ultimately consider going for one of our preprocessor list tricks here, with one table that is used to populate the enumeration and ToString function automatically.

larpandoracontent/LArHelpers/LArInteractionTypeHelper.h
```diff
@@ -198,10 +235,11 @@ class LArInteractionTypeHelper
      *
      *  @param  nuanceCode the nuance code
      *  @param  mcPrimaryList the list of primary mc particles
+     *  @param  testBeamHierarchyMode
```
PandoraPFA: We'll have to see if we can come up with some neat text to describe exactly what's different under this mode.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -57,6 +57,18 @@ bool LArMCParticleHelper::IsBeamParticle(const MCParticle *const pMCParticle)
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+bool LArMCParticleHelper::IsLeadingBeamParticle(const MCParticle *const pMCParticle)
+{
+    // ATTN: Only the parent triggered beam particle has nuance code 2001
+    const int parentNuance(LArMCParticleHelper::GetNuanceCode(LArMCParticleHelper::GetParentMCParticle(pMCParticle)));
+
+    bool isLeading(LArMCParticleHelper::IsLeading(pMCParticle) && (parentNuance == 2001 || parentNuance == 2000));
```
PandoraPFA: Need to review where the actual values of the nuance codes are used, and ensure that it's in the absolute minimum number of places: ideally just in one or two helper functions, so that the values don't appear anywhere else. (const bool)

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -294,33 +430,58 @@ void LArMCParticleHelper::SelectReconstructableMCParticles(const MCParticleList
...
 
     // Select MCParticles matching criteria
     MCParticleVector candidateTargets;
     LArMCParticleHelper::SelectParticlesMatchingCriteria(mcPrimaryVector, fCriteria, candidateTargets);
 
     // Ensure the MCParticles have enough "good" hits to be reconstructed
     LArMCParticleHelper::SelectParticlesByHitCount(candidateTargets, mcToTrueHitListMap, mcToPrimaryMCMap, parameters, selectedMCParticlesToHitsMap);
+
+    // ATTN: The parent particle must be in the hierarchy map, event if not reconstructable
+    if (testBeamHierarchyMode)
```
PandoraPFA: Was this not required before the introduction of the test beam hierarchy mode? (I'm here just looking for ways of sharing implementation between the with and without test beam hierarchy modes). Currently leaning towards two separate functions, each maybe then calling a block of common implementation, rather than one function with a Boolean argument/switch.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -294,33 +430,58 @@ void LArMCParticleHelper::SelectReconstructableMCParticles(const MCParticleList
...
+    {
+        for (const MCParticle *const pMCParticle : candidateTargets)
+        {
+            if (!LArMCParticleHelper::IsBeamParticle(pMCParticle))
+                continue;
+
+            const MCParticle *const pParentMCParticle(LArMCParticleHelper::GetParentMCParticle(pMCParticle));
+            if (selectedMCParticlesToHitsMap.find(pParentMCParticle) == selectedMCParticlesToHitsMap.end())
+            {
+                CaloHitList caloHitList;
```
PandoraPFA: Can combine these steps using the `operator[]`, which will create a new key, value pair if it doesn't exist, or return a reference to the value if the key is present.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -388,11 +549,20 @@ void LArMCParticleHelper::GetPfoMCParticleHitSharingMaps(const PfoContributionMa
...
 void LArMCParticleHelper::CollectReconstructable2DHits(const ParticleFlowObject *const pPfo, const MCContributionMapVector &selectedMCParticleToHitsMaps,
-    pandora::CaloHitList &reconstructableCaloHitList2D)
+    pandora::CaloHitList &reconstructableCaloHitList2D, const bool testBeamHierarchyMode)
 {
     // Collect all 2D calo hits in pfo hierarchy
     PfoList pfoList;
-    LArPfoHelper::GetAllDownstreamPfos(pPfo, pfoList);
+
+    // ATTN: Only collect downstream pfos for daughter test beam particles, cosmics, neutrinos
+    if (testBeamHierarchyMode && pPfo->GetParentPfoList().empty() && LArPfoHelper::IsTestBeam(pPfo))
```
PandoraPFA: Suggest create a function in the LArPfoHelper to sit alongside the `GetAllDownstreamPfos` function, but for this use case. In this way, takes away the need for any actual logic (no matter how small/trivial) hidden inside this function.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -25,6 +25,7 @@ class LArMCParticleHelper
 {
 public:
     typedef std::unordered_map<const pandora::MCParticle*, const pandora::MCParticle*> MCRelationMap;
+    typedef std::unordered_map<const pandora::MCParticle*, int> MCParticleIntMap;
```
PandoraPFA: int or unsigned int for hierarchy tiers?

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -75,6 +76,11 @@ class LArMCParticleHelper
      */
     static bool IsBeamParticle(const pandora::MCParticle *const pMCParticle);
 
+    /**
+     *  @brief  Returns true if passed a leading beam MCParticle
+     */
+    static bool IsLeadingBeamParticle(const pandora::MCParticle *const pMCParticle);
```
PandoraPFA: Should provide a good definition of leading beam particle in the header file. It's subtly different to a primary beam particle, and the primary definition could be refreshed too.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -103,6 +109,24 @@ class LArMCParticleHelper
...
+    static bool IsLeading(const pandora::MCParticle *const pMCParticle);
+
+    /**
+     *  @brief  Determine the position in the hierarchy for the MCParticle
+     *
+     *  @param  pMCParticle the address of the mc particle
+     *
+     *  @return integer
+     */
+    static int GetHierarchyTier(const pandora::MCParticle *const pMCParticle);
```
PandoraPFA: I like hierarchy tier, as (to me) it seems nicely self-describing.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -129,6 +153,15 @@ class LArMCParticleHelper
...
     static const pandora::MCParticle *GetPrimaryMCParticle(const pandora::MCParticle *const pMCParticle);
 
+    /**
+     *  @brief  Get the leasding particle in the hierarchy, for use at ProtoDUNE
+     *
+     *  @param  pMCParticle the input mc particle
+     *
+     *  @return address of the primary parent mc particle
+     */
+    static const pandora::MCParticle *GetLeadingMCParticle(const pandora::MCParticle *const pMCParticle, const int hierarchyTierLimit = 1);
```
PandoraPFA: Is the limit likely to be a parameter of our algorithms or, realistically, always set to 1? (Definitely not saying we should hard code it!!!)

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -190,29 +239,33 @@ class LArMCParticleHelper
...
      *  @brief  Get mapping from Pfo to reconstructable 2D hits (=good hits belonging to a selected reconstructable MCParticle)
      *
      *  @param  pfoList the input list of Pfos
      *  @param  selectedMCParticleToHitsMap the input mapping from selected reconstructable MCParticles to their hits
      *  @param  pfoToReconstructable2DHitsMap the output mapping from Pfos to their reconstructable 2D hits
+     *  @param  testBeamHierarchyMode test beam hierarchy mode
      */
     static void GetPfoToReconstructable2DHitsMap(const pandora::PfoList &pfoList, const MCContributionMap &selectedMCParticleToHitsMap,
-        PfoContributionMap &pfoToReconstructable2DHitsMap);
+        PfoContributionMap &pfoToReconstructable2DHitsMap, const bool testBeamHierarchyMode = false);
```
PandoraPFA: Anything we can do to push logic into the algorithms, and then maybe have two functions, or one function that calls a second function for extra processing when in test beam hierarchy mode, would help. This is as opposed to a Boolean in the function itself. This would fit with the idea of having clean helper functions that can be reviewed in isolation, together with separate algorithms that drive particular use cases.

larpandoracontent/LArMonitoring/EventValidationAlgorithm.cc
```diff
@@ -24,6 +24,7 @@ namespace lar_content
 EventValidationAlgorithm::EventValidationAlgorithm() :
     m_useTrueNeutrinosOnly(false),
     m_testBeamMode(false),
+    m_testBeamHierarchyMode(false),
```
PandoraPFA: How well do you thing this would work if we tried to factorise out the old and new/hierarchy modes into separate algorithms? It would avoid e.g. the re-use of the `*Nu` variables for TB usage, and the Nu vs. TB string name, but will lead to repetition.

Some repetition is OK if it's really straightforward and simple. I wondered about introducing algorithm tools to interpret the final matching info. Could have one tool to print comprehensive matching to terminal, one to print interpreted matching, one to write a tree and one to show things in an event display. Then, instead of setting Boolean flags, simply enable/disable the tools in XML.

Might also need tools to populate the info maps differently for different use cases, or maintain separate algs or the Boolean switch. I'm thinking aloud here and generally liking the idea of getting the event display that matches the matching print out to terminal back in action again!

larpandoracontent/LArMonitoring/EventValidationAlgorithm.cc
```diff
@@ -100,15 +101,17 @@ void EventValidationAlgorithm::FillValidationInfo(const MCParticleList *const pM
         parameters.m_maxPhotonPropagation = m_maxPhotonPropagation;
         LArMCParticleHelper::MCContributionMap targetMCParticleToHitsMap;
         LArMCParticleHelper::SelectReconstructableMCParticles(pMCParticleList, pCaloHitList, parameters, LArMCParticleHelper::IsBeamNeutrinoFinalState, targetMCParticleToHitsMap);
-        if (!m_useTrueNeutrinosOnly) LArMCParticleHelper::SelectReconstructableMCParticles(pMCParticleList, pCaloHitList, parameters, LArMCParticleHelper::IsBeamParticle, targetMCParticleToHitsMap);
+        if (!m_useTrueNeutrinosOnly && !m_testBeamHierarchyMode) LArMCParticleHelper::SelectReconstructableMCParticles(pMCParticleList, pCaloHitList, parameters, LArMCParticleHelper::IsBeamParticle, targetMCParticleToHitsMap);
```
PandoraPFA: Leaving rest of algorithm for now, until we've discussed thoughts from above.

StevenGreen1:
Hi John, 

Having read your comments above, I'm going to work on splitting this into two separate algorithms with as much shared code factored out into functions as possible.  I'll then update the PR.  Long term, I imagine we will want a coupled piece of code, because of the complex DUNE hierarchies, but if we can shape the code so that the actual algorithms are small then it should be semi-easy to implement in the future.  

Cheers
Steve

StevenGreen1:
Hi @johnmarshall80, 

I've separated out the current logic so that there is now an event validation base algorithm that is inherited by the current event validation algorithm and a new test beam event creation algorithm.  The internal LArMCParticleHelpers are also now separated and do not contain any testBeamMode bools.  I've checked this for a ProtoDUNE event and everything appears to work as it did when I previously tested it.  

Could I please request a review of this code?  It would be great to have this done asap as I would ideally like to use this code for testing a BDT vertex algorithm that I've created and that will appear as another PR in the very near future.  Furthermore, if we could get these two PRs in quickly it would be great to show improved results at the next DUNE CM for the correct event fraction (Stefano's work).  It would also really help all of the studies currently being shown in the ProtoDUNE DRA and analysis meetings, which don't look as good as I think we can do.  Although I do appreciate you're very busy too and understand if you can't review this for some time.  Thanks in advance. 

Cheers
Steve

PandoraPFA:
Hi Steve,

Thanks for the nice work on this complex, but very important feature. I've had a look through the new diffs. I was wondering how you thought it "felt" as you separated out the single EventValidation algorithm into two algorithms? Did it feel "right", as if order was emerging from chaos, or "wrong" because you could see code repetition appearing?

It looks like the division into a base class (run method and controls the generic matching process), and derived algorithms is quite clean. I did wonder whether there was anything further that was common to be factored out when filling the validation info object, or when performing the main analysis/printout/tree writing? I also wondered whether it was worth completely decoupling the neutrino vs. testbeam uses and having another separate algorithm? Or do you think these are better left entwined as it's a price worth paying to reduce code repetition?

Other than that, I think that the sheer number of interaction types in the enum and ToString does warrant one of our preprocessor list tricks, and that the header file includes in the three EventValidation-related algorithm header files can be reviewed and pruned (and maybe additions made, where currently relying on picking up required header files "accidentally").

I think we all agree that a key step is then to get this tested for high stats (happy to help as soon as I'm free again), including for MicroBooNE and DUNE FD, where we expect zero changes (so a clear test) and with/without hierarchy mode for ProtoDUNE. At that point, we have to decide whether to keep on agonising about making it perfect, or be a bit more pragmatic and merge as is (which isn't bad, just reflects the complexity of the problem - I've always thought this was one of the more difficult problems we tackled!).

Given that this event validation is only used by developers (I think), we could keep this in a feature branch for a little while longer whilst we move it forwards (certainly for high-stats testing), but I do agree that it mustn't stay in a feature branch too long or stagnate. The feature for supporting either a BDT or a SVM approach for vertex selection is presumably a much simpler, less-sprawling change and can hopefully go straight in rather quickly (key test: MicroBooNE performance should be unchanged).

Thanks again, and any/all thoughts appreciated!

Best,

John

StevenGreen1:
Hi John, 

Thanks for having a look at this code so quickly, I know how busy you are at the moment and I do really appreciate this.  

I felt that order really was appearing from the chaos when splitting this into a base class and the two derived algorithms.  What I would say that became quite evident was that although there was a repeated code structure most of the lines are in fact unique.  I played around using the comm command this morning to find shared lines of code between the two .cc files for the derived algorithms and I found that there was only 205 lines of shared code in total.  I've attached the text file showing the common lines for interest.  

From these common lines there are roughly 3 sizeable blocks:

- Firstly the setting of parameters in the validation info class, which has to be in the derived algorithm so we can define test beam/test beam hierarchy/neutrino targets.
- Secondly the filling of the information for the mc/reco primaries to be saved in the root tree.
- Finally the writing of variables to the root tree. 

There is some other shared coder appearing due to a different naming scheme, but the main point is that a lot of the code is different and so I felt this was a good suggestion by you and that we should go with this.  In terms of reducing the common lines for the tree writing, I think that's going to be a bit tricky as we'd need to have access to all the vectors/floats/ints being filled and the names that we want to call them.  We would also need some customisable access because there are neutrino variables that don't going into the test beam tree and vice versa.  For simplicity I would suggest leaving the duplicated code for now and perhaps restructuring later on if the common code beings to grow further.  

I agree that we need to use some tricky to sort the enum particle interaction type.  That list isn't close to being exhaustive either.  I'm not familiar with the method you suggest so perhaps we should have a chat about that at some point whenever is convenient for you.  

I'll definitely go over the includes in these files to reduce what's there for the next iteration.  I didn't spend long pruning this now as we were still discussing the higher level decisions (e.g. base algorithms vs no base etc.) and so didn't want to spend long fixing code that might be removed in the end.  

I'm going to use this feature branch with the BDT vertexing one to update the efficiencies Stefano was showing so that things will hopefully look a lot better for the CM next week as I do feel it's important we address this asap.  I'll make sure we have preliminary on things and say that work is ongoing to get both these features out (but avoid giving a deadline).  

Thanks again for going over this, I do think this will make a big difference for ProtoDUNE and really help us to make the Reco better.

Cheers
Steve 

Common Lines:
[Common.txt](https://github.com/PandoraPFA/LArContent/files/3181633/Common.txt)

PandoraPFA:
Hi Steve,

That's excellent to hear that you felt order was emerging from chaos, and just what I'd hoped you'd say! It's good that this isn't too entangled, and can be teased into separate algorithms without excessive repetition. The long-term goal, of course, is clarity of implementation (reduces mistakes and eases debugging) and ease of maintenance (e.g. how many algs to update when we change a helper function?).

I wonder whether we can further reduce code duplication when filling the ValidationInfo by adding a virtual function that returns the criteria to be used in the selection. The rest of the function implementation is common and is implemented in the base class, which then calls e.g. `this->GetMCParticleCriteria()` or similar.

I agree that the ROOT tree writing might be more challenging to tease apart! There may be some more than can be done, but, rather than agonising too much, we should set our main criteria for release as: ensure that all test results are good (as per earlier discussion) and satisfy ourselves that a sensible (but pragmatic) amount of thought has gone into clarity of implementation and future maintenance.

In terms of the preprocessor tricks (optional, of course, but reduces code), a good example is our Pandora StatusCodes and PdgCodes (or our algorithm registration, which we've all modified many times before), which you can find here:
[table](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/PdgTable.h#L19), [example macro](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/PdgTable.h#L58), [example function](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/PdgTable.h#L172) or [table](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/StatusCodes.h#L72), [example macro](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/StatusCodes.h#L87), [example enum](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/StatusCodes.h#L99) or [more fiddly table](https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArContent.cc#L179)

I think, for clarity, though, the modifications we're discussion are now optional (but should help us in the long term), but good test results are obviously required, and maybe factoring out a third TestBeam algorithm is a sensible thing too?

I hope that helps! Thanks again!

John

StevenGreen1:
Hi @johnmarshall80 , I know it's been a while since I looked at this, but want to try and address the comments now and get this into our Git.  Also the fact that Stefano's work using this seems to have been well received  I think it would be nice if we could make this official soon.  So in terms of the work needed I believe that I need to:

1. Split up the current event validation algorithm into a test beam only and a neutrino only option.
2. Add in the virtual function GetMCParticleCriteria.
3. Preprocessor trick for reducing the interaction types.  

It seems to me like these are all sensible and so are worth doing.  I'll get on these now, but do let me know if I've missed anything and if not I'll update this PR once finished.  

Cheers
Steve

PandoraPFA:
Hi Steve,

That all sounds good - many thanks. As we discussed before, these changes are sufficiently involved that testing is a very important aspect (i.e. reproduce old results without new hierarchy mode; is all working as expected with new hierarchy mode for high-stats samples?). Obviously we always do this anyway :) but we frequently try to have the code under such tight control that the test results are a given. Here, with rather more involved changes, there's more potential for unexpected issues.

Thanks again,

John

StevenGreen1:
Hi John,

When attempting to move the FillValidationInfo function the to base class, I think there is a problem with the this->GetMCParticleCriteria() function (and the equivalent for Pfos) idea.  In order to pass this function to the SelectReconstructableMCParticles LArMCParticleHelper, the selection criteria function has to be static.  This means we can't make this->GetMCParticleCriteria() a virtual function in the base class and have each instance of a validation algorithm overwrite it.  If you know of a workaround that would allow us to do this them I'm happy to try and follow that up though.

When looking again at whether it's worth doing this I did note that the FillValidationInfo functions do differ between the EventValidation, TestBeamEventValidation and TestBeamHierarchyEventValidation algorithms by more than just the this->GetMCParticleCriteria() function (and the equivalent for Pfos).  In the TestBeamHierarchyEventValidation different LArMCParticleHelper functions are called as the default ones look to fold hierarchies too far back up the chain otherwise, so my inclination is that it might be worth just leaving this for now and accepting the little bit of duplicate code.  

The only other option I can think of to further reduce the code duplication would be to split the FillValidationInfo function into steps.  We could then make each step a function in the base class apart from duplicate code which could be defined in the base class.  One of these functions would call the SelectReconstructableMCParticles using a static LArMCParticleHelper function in each instance of a EventValidation algorithm.  The issue I have with this is that I think it might make the FillValidationInfo function hard to read, while currently it's quite clean.  

I'd appreciate your thoughts on this one when you get some time and in the mean time I'll move onto trying to do the preprocessor trick stuff.
Cheers
Steve

StevenGreen1:
Hi John,

So I've been looking into the preprocessor trick and I think I understand how this works and how I could go about implementing this.  The problem that I have found with this is the combinations tend to explode.  As from a pion interaction you could feasibly get several e+, e-, mu+, mu-, proton, photon, pi+, pi-, pi0, K+, K- and K0, if you wanted to cover all combinations up to 5 of each particle you need 244 million enums!  If you dropped down to 3 of each you'd still need 4000 lines of code to define the enums in the preprocessor table.  

This makes me think that the interaction type enum is not a sensible thing to include for the hierarchy problem (fundamentally I think for complex events it's a bit redundant).  Instead I'd suggest scrapping it from the output variable for the hierarchy method and then adding additional variables to the root tree showing how many of these types of particles come from the primary interaction.  This would mean that Validation.C will no longer work for the test beam hierarchy mode root files, but I don't think that's a problem worth forcing thousands of lines of additional enum definition code.  Also as we've now split the test beam/hierarchy/neutrino validation algorithms into separate algorithms then we can easily do this for the hierarchy, but leave the neutrino and current test beam options unchanged. 

How does that sound to you?  Also I understand there's a few things where it might be easier to discuss in person so if you'd prefer to arrange a Skype chat to go over this just let me know.  

Cheers
Steve

StevenGreen1:
Hi,

The additional commits here apply the preprocessor trick for defining interaction types (and the ToString function) in the LArInteractionTypeHelper class.  Furthermore, the GetInteraction type function has been split into GetInteractionType and GetTestBeamHierarchyInteractionType and the common code refactored into single functions.  A new class was needed to set the interaction type parameters cleanly instead of passing around the 9 ints for nMuon, nElectron, nPiPlus etc... 

Testing of this feature branch is underway and results will follow when ready.

Cheers
Steve

StevenGreen1:
Hi,

The first stage of testing this PR is now complete.  I tested mcc11 ProtoDUNE events at 1, 3, 5 and 7 GeV with space charge and with beam and cosmics.  The differences are as follows:

- The isSplitTB parameter is now set where previously is was not (no issue).
- The interactionType integer differed as the calls to the relevant functions in the validation algorithms were wrong (resolved now, 4253e1d).
- The mcPrimaryIndex was not being incremented (resolved now, 4253e1d). 

I will rerun the testing, but expect this to fix the issues and then move onto comparing DUNEFD events.

Cheers
Steve

PandoraPFA:
Hi Steve,

Thanks for the nice update - sounds like the testing is proving very useful. Thanks for checking DUNEFD events too, and it'd be helpful if you could quickly test a few MicroBooNE events as well, with and without cosmics (all ought to be straightforward for this, after the other tests).

Thanks again,

John

StevenGreen1:
Hi, 

The new ProtoDUNE tests are complete and with the latest modifications (4253e1d), the event validation root files are identical.  Will update again once the DUNEFD checks are complete.

Cheers
Steve

StevenGreen1:
Hi, 

The DUNEFD testing is complete and the majority, >90%, of the output root files (200 events per file, I believe) were identical using the new event validation algorithm.  Any difference in root file output can be solely due to a single event in the file and so on a per event basis I expect identical performance percentage to be well in excess of 90%.  This testing was done on the entire mcc11 dune dataset.  It is believe that the differences occurring for some files is due to a reproducibility issue that is being examined now and not the changes in this feature.

This branch was also tested for MicroBooNE on ~150 BNB+Cosmic and 1,500 BNB events and the output was identical.  

I believe this feature is now ready for a final review of the code before being merged.  In that time I hope we can resolve the reproducibility issue and guarantee identical performance for DUNEFD also.

Cheers, Steve

larpandoracontent/LArHelpers/LArMCParticleHelper.h
StevenGreen1: I think that it is likely that it will be a parameter of the algorithms because for neutrino interactions we currently drill down to tier 1 (tier 0 being the neutrino and 1 being the primaries, but this is a weird case as we don't see tier 0 particles) and for ProtoDUNE we have been going to tier 0 (the parent test beam particle) and now we want to go to 1.  Ultimately, the next step for this is to go to higher tiers for DUNE (secondary interactions etc), so with that in mind, I wanted to make these changes easily extendible for future use cases, which is why I wanted this to be configurable.

StevenGreen1:
Rebased and merged into master branch.

PR #79, opened 2019-04-05
Define out of time slices in CR tagging using all pfos in slice.

StevenGreen1:
A minor fix to the CR tagging that identifies whether a slice is out of time based on the information from all pfos in that slice as opposed to a single pfo in the slice.  

Tests of this branch have been done for ProtoDUNE and showed that this change resulted in a very small improvement to the cosmic ray and beam particle reconstruction metrics:

Cosmic Ray:
CosmicRayEfficiencyVsNHits
CosmicRayCompleteness
CosmicRayPurity

Beam Particle: 
BeamParticleEfficiencyVsNHits
BeamParticleCompleteness
BeamParticlePurity

### DIFF
--- a/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.cc
+++ b/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.cc
@@ -462,12 +462,23 @@ void CosmicRayTaggingTool::CheckIfTopToBottom(const CRCandidateList &candidates,
 void CosmicRayTaggingTool::GetNeutrinoSlices(const CRCandidateList &candidates, const PfoToBoolMap &pfoToInTimeMap, const PfoToBoolMap &pfoToIsContainedMap,
     UIntSet &neutrinoSliceSet) const
 {
+    IntBoolMap sliceIdToIsInTimeMap;
+
+    for (const CRCandidate &candidate : candidates)
+    {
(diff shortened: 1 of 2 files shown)


StevenGreen1:
Also for reference the raw matching numbers: 

Event Description    : Default Reco
nBeamParticles       : 3576
nBeamParticleMatches : 2886
Efficiency [%]       : 80.7047+-0.659897
nCosmic              : 281231
nCosmicMatches       : 266301
Efficiency [%]       : 94.6912+-0.0744121

Event Description    : CR In Time Tagging Update
nBeamParticles       : 3576
nBeamParticleMatches : 2902
Efficiency [%]       : 81.1521+-0.654007
nCosmic              : 281231
nCosmicMatches       : 267430
Efficiency [%]       : 95.0927+-0.0737479

PandoraPFA:
Hi Steve,

We've now discussed this directly, so here's just a few notes of record:

-As a result of this change, if any pfo in a slice is out of time then the entire slice is now marked as being out of time. The performance impact of this change has been demonstrated for ProtoDUNE.

-Assignment via non-const reference from map.at() requires that the relevant key already be present in the map, but this is enforced here, with a placeholder insertion made in `sliceIdToIsInTimeMap` for each `sliceId`. Similarly `pfoToInTimeMap` must already contain a key for each pfo address.

Many thanks for the fast fix to address a rather subtle issue!

John

PR #80, opened 2019-05-07
feature/ClearTrackFragments

PandoraPFA:
Hi all,

We'll want this to go out across the whole codebase soon, and I'd really appreciate some cross checks. Many thanks!

Restructuring and simplification to resolve issues with a (previously) excessively complicated tool. Danger was that reclustering operations were being performed on clusters that were present/active in an OverlapTensor. The tool was responsible for ensuring that the tensor was updated correctly and remains valid for subsequent tools (or repeat passes of the same tool).

Now perform at most one particle creation per pass of the ClearTrackFragments tool. Extract and cache all relevant parts of the tensor, then remove all affected clusters from the tensor before allowing operations to proceed. [Need to be so careful with reclustering operations, as e.g. the address of a previously deleted cluster can be reused for a newly allocated cluster, so nothing is straightforward!]

Tested with MicroBooNE events, yielding unchanged performance (as per EventValidation algorithm and Validation.C) but reduced complexity of tool and memory use issues addressed.

John


larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -52,52 +52,52 @@ bool ClearTrackFragmentsTool::FindTrackFragments(ThreeDTrackFragmentsAlgorithm *
...
 
-        this->RebuildClusters(pAlgorithm, modifiedClusters, clustersToAddToTensor);
-        this->UpdateTensor(pAlgorithm, overlapTensor, clustersToRemoveFromTensor, clustersToAddToTensor);
+        if (!(pClusterU->IsAvailable() && pClusterV->IsAvailable() && pClusterW->IsAvailable()))
+            throw StatusCodeException(STATUS_CODE_FAILURE);
 
-        if (particlesMade)
-            return true;
+        // ATTN For safety, remove all clusters associated with this fragment particle from the tensor
+        ClusterList fragmentClusterList, affectedKeyClusters;
```
StevenGreen1: This is a minor comment, but what is the difference between clusters and keyClusters?  I believe keyClusters are a subset of the fragmented clusters that have sufficient length and number of hits to be considered in the overlap tensor.  Is this correct?

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -280,31 +257,47 @@ void ClearTrackFragmentsTool::ProcessTensorElement(ThreeDTrackFragmentsAlgorithm
 
         this->Recluster(pAlgorithm, pCluster, daughterHits, separateHits, deletedClusters, badClusters, pFragmentCluster);
 
-        if (badClusters.count(pFragmentCluster))
+        // ATTN Fragment cluster will be used to build particle, so it shouldn't ever be bad, or deleted
+        if (badClusters.count(pFragmentCluster) || deletedClusters.count(pFragmentCluster))
             throw StatusCodeException(STATUS_CODE_FAILURE);
```
StevenGreen1: This check makes sense.  If this exception gets thrown, do you know where it is caught and what the affect on the remaining reconstruction is?  Does it bail out completely or just skip the remaining logic in this algorithm and how does this impact the tensor if it's thrown?

StevenGreen1:
Hi John,

I've had a look through these changes and I believe that everything you've done makes sense in terms of caching the tensor information prior to the reclustering.  I added two comments asking questions about the code as I didn't have anything else major to contribute in terms of changes.  It's great that is looks a lot simpler now than it did before though so well done on that!  

One slightly more open question is that, now the code only performs one creation of a particle per pass at most, does the cut requiring that the tool be run less than 1000 times (https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArThreeDReco/LArTrackFragments/ThreeDTrackFragmentsAlgorithm.cc#L22) impact the reconstruction?  I imagine that for MicroBooNE this will probably not happen as the topologies are simpler, but potentially for DUNEFD there could be the odd case or two where things now change.  I imagine the easiest way to check this would be to run a batch of DUNEFD jobs and check those metrics prior to release too.  How does that sound?

Cheers,
Steve

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
PandoraPFA: Thanks a lot for looking! Yes, the keyClusters are those clusters upon which the tensor is keyed. The tensor is like a big, nested map, keyed on U clusters, then V clusters, then W clusters and finally having an OverlapResult as its ultimate value.

Most algorithms will place some quality cuts on the clusters that are used as key clusters. Typically, the key clusters are the only clusters that are ever considered, and which will be modified and/or used to build particles.

In this one, ambitious alg/tool, however, the plan was to link good key clusters in two views, e.g. V, W and associate them to lots of fragment clusters in the third view e.g. U.

It's still a good idea, and it makes good changes (well, examining event displays during development suggested this), but you can see how this makes things more complicated, especially as the alg/tool together request the fragments to be reclustered and merged, and then needs to update the tensor for the next passes!

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
PandoraPFA: I think this will terminate the algorithm, i.e. exit the tool, exit the algorithm, so it's all over and the tensor is destroyed. I searched for this failure during my uB tests and it didn't occur during the tests with the code as is (but was triggered during earlier tests involved with these updates).

PandoraPFA:
Hi Steve,

Thanks so much for carefully checking through this. I really appreciate informed second opinions here, as this alg/tool tries to do something very complicated and clearly the complexity of the earlier implementation made it hard to keep track of what is really happening.

Regarding the number of tool iterations, I don't think it's very likely that an algorithm will run the tensor tool more than a handful of times each event - I've only seen it run twice so far in my tests with uB and in the problematic DUNE FD events that triggered this development.

I completely agree with running the DUNE FD metrics too, though. My thinking is that it'd be fine to stop after any large-ish number provided that the code is deterministic and reproducible, doing the same thing every time it runs.

[I don't think we can visualise what it would mean to be making 1000+ changes here. 1000 is basically a large number to get rid of the rare possibility of getting stuck in an infinite loop where two tensor tools make conflicting changes - I have seen this happen twice, both times in uB cosmic ray events a few years ago now, but seems rare.]

Thanks,

John

StevenGreen1:
Hi John,

Thanks for having a look at the DUNE metrics too.  I think you're probably right in that the tool shouldn't get close to running 1000 times even for DUNE, but it's good to double check this with the testing you've got on the go.  It'll be good to get this released neat week if all goes smoothly with the final testing.

Cheers
Steve

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -52,52 +52,52 @@ bool ClearTrackFragmentsTool::FindTrackFragments(ThreeDTrackFragmentsAlgorithm *
         IteratorList iteratorList;
         this->SelectClearElements(elementList, iteratorList);
 
-        bool particlesMade(false);
-        ClusterList deletedClusters, modifiedClusters, clustersToRemoveFromTensor, clustersToAddToTensor;
-        ClusterSet badClusters;
-
-        for (IteratorList::const_iterator iIter = iteratorList.begin(), iIterEnd = iteratorList.end(); iIter != iIterEnd; ++iIter)
```
loressa: I think I understand and agree with the simplification here. In the broader picture, the tensor will be updated after the tool makes a change, so this iteration inside the tool main function is not lost by the replacement with iteratorList.front() as the tool would run again and evaluate if more elements are connected by SelectClearElements. Also, the iteratorList is sorted by overlapResult (from GetConnectedElements in LArOverlapTensor and its operator < definition), which makes sense to start with the largest overlap result in the list of elements. And it solves potential problems so I like the simplification.

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -52,52 +52,52 @@ bool ClearTrackFragmentsTool::FindTrackFragments(ThreeDTrackFragmentsAlgorithm *
...
+            pAlgorithm->UpdateUponDeletion(pCluster);
+
+        // Now make the particle
+        ProtoParticle protoParticle;
+        ProtoParticleVector protoParticleVector;
+        protoParticle.m_clusterListU.push_back(pClusterU);
+        protoParticle.m_clusterListV.push_back(pClusterV);
+        protoParticle.m_clusterListW.push_back(pClusterW);
+        protoParticleVector.push_back(protoParticle);
+        return pAlgorithm->CreateThreeDParticles(protoParticleVector);
```
loressa: is it worth having the result of CreateThreeDParticles checked before the affected clusters go under UpdateUponDeletion? What would happen if for any reason the creation of the threed particle fails (and we return false at this point)?

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -132,32 +132,9 @@ bool ClearTrackFragmentsTool::GetAndCheckElementList(const TensorType &overlapTe
...
-
-            allMatchedHits.push_back(*hIter);
-        }
-    }
-
-    return true;
-}
-
-//------------------------------------------------------------------------------------------------------------------------------------------
-
```
loressa: Well spotted, as this appears to never being used.

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -367,30 +360,6 @@ void ClearTrackFragmentsTool::RebuildClusters(ThreeDTrackFragmentsAlgorithm *con
 
 //------------------------------------------------------------------------------------------------------------------------------------------
```
loressa: nice change by using sets!

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
```diff
@@ -367,30 +360,6 @@ void ClearTrackFragmentsTool::RebuildClusters(ThreeDTrackFragmentsAlgorithm *con
...
-
-    for (ClusterList::const_iterator iter = affectedKeyClusters.begin(), iterEnd = affectedKeyClusters.end(); iter != iterEnd; ++iter)
-        pAlgorithm->UpdateUponDeletion(*iter);
-
-    for (ClusterList::const_iterator iter = affectedKeyClusters.begin(), iterEnd = affectedKeyClusters.end(); iter != iterEnd; ++iter)
-        pAlgorithm->UpdateForNewCluster(*iter);
-}
-
-//------------------------------------------------------------------------------------------------------------------------------------------
-
```
loressa: I agree with removing this, its clearer by adding the relevant operations inside ProcessTensorElement

loressa:
Hello, 

I have also had a look at the changes and I think I understand their logic and agree with them. I only had a couple of minor comments. 

I have tested their impact in high stats for DUNE FD (MCC11 production) and the results can be compared here:
https://www.hep.phy.cam.ac.uk/~escudero/DUNEFD_MCC11_clean.txt
https://www.hep.phy.cam.ac.uk/~escudero/DUNEFD_MCC11_JohnPR.txt

Some files are still running for this PR, so note that some number of events differ. For those with the same (or enough statistics) some changes appear, which are very small. I think this is not unreasonable, as the initial ground in which something triggering these changes was found in DUNE FD events. 

Have a look and see if you are happy, but on my side I think it should be good to go. 

Best,
Lornea

larpandoracontent/LArThreeDReco/LArTrackFragments/ClearTrackFragmentsTool.cc
PandoraPFA: I thought about this, but the checks above in the tool essentially mean that false can _currently_ never be returned (we check cluster availabilities and indeed ensure that there are some clusters provided). If false is returned, the control is passed to the next tensor tool to have a go.

Even if we changed the implementation so that false could be returned, everything would be well defined: the worst that can happen is that some clusters could be taken out of the tensor, but (somehow) _not_ turned into a particle and would be unavailable for the next tensor tool.

With the current implementation, false cannot happen (exceptions could be raised inside CreateThreeDParticles, but this would terminate the algorithm entirely) and there also is no other tool to run next in the chain. I'm therefore minded to go with this as it has been well tested and is consistent with other tools used by other tensor-related algorithms.

[Being pedantic, we could raise an exception if CreateThreeDParticles doesn't return true. If it does return true, we'd then manually need to return true in the tool. Would probably want to review all tools from all tensor-related algorithms in a specific new piece of maintenance.]

PR #82, opened 2019-06-10
Changed the cuts used to select which clusters are considered in the …

phuncr:
…stitching


PandoraPFA:
Merged (into feature branch feature/larpandoracontent_v03_15_00). Selection cut values set such that (by default) no change in functionality is observed.

PR #83, opened 2019-07-03
Feature/stitching

loressa:
Reduce vector of potential stitches to two to keep stitching across CPA now that APA hits are availalbe. This is a short term fix, prior to extend stitches across APAs in addition. This is a work in progress PR, not yet fully validated, but wanted to open it already for people to have a look. Feedback about map vs double loop (see TODO comments) are especially welcome. It has been tested on the single muon particle gun events created by Leigh. See a couple of examples in action [here](https://www.hep.phy.cam.ac.uk/~escudero/ProtoDUNEStitchingExample1.png) and [here](https://www.hep.phy.cam.ac.uk/~escudero/ProtoDUNEStitchingExample2.png).

### DIFF
--- a/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
+++ b/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
@@ -557,7 +557,7 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
             continue;
         }
 
-    if ((stitchedLArTPCs.first == nullptr) || (stitchedLArTPCs.second == nullptr))
+        if ((stitchedLArTPCs.first == nullptr) || (stitchedLArTPCs.second == nullptr))
             throw StatusCodeException(STATUS_CODE_FAILURE);
 
         float x0(0.f);
--- a/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
+++ b/larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
@@ -538,7 +538,7 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
         std::sort(pfoVector.begin(), pfoVector.end(), LArPfoHelper::SortByNHits);
 
         PfoVector reducedPfoVector;
-        const ParticleFlowObject *const selectedPfoToEnlarge(this->ReduceToLongestStitch(pfoVector, reducedPfoVector, pPfoToEnlarge, pfoToLArTPCMap));
+        const ParticleFlowObject *const pSelectedPfoToEnlarge(this->ReduceToLongestStitch(pfoVector, pPfoToEnlarge, pfoToLArTPCMap, reducedPfoVector));
 
         LArTPCPair stitchedLArTPCs(nullptr, nullptr);
 
@@ -556,6 +556,9 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
 
             continue;
         }
+
+    if ((stitchedLArTPCs.first == nullptr) || (stitchedLArTPCs.second == nullptr))
+            throw StatusCodeException(STATUS_CODE_FAILURE);
 
         float x0(0.f);
         const float tpcBoundaryCenterX(LArStitchingHelper::GetTPCBoundaryCenterX(*stitchedLArTPCs.first, *stitchedLArTPCs.second));
@@ -595,31 +598,30 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
 
         for (const ParticleFlowObject *const pPfoToDelete : reducedPfoVector)
         {
-            if (selectedPfoToEnlarge == pPfoToDelete)
+            if (pSelectedPfoToEnlarge == pPfoToDelete)
                 continue;
 
-            pAlgorithm->StitchPfos(selectedPfoToEnlarge, pPfoToDelete, pfoToLArTPCMap);
-            stitchedPfosToX0Map.insert(PfoToFloatMap::value_type(selectedPfoToEnlarge, x0));
+            pAlgorithm->StitchPfos(pSelectedPfoToEnlarge, pPfoToDelete, pfoToLArTPCMap);
+            stitchedPfosToX0Map.insert(PfoToFloatMap::value_type(pSelectedPfoToEnlarge, x0));
         }
     }
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-const ParticleFlowObject *StitchingCosmicRayMergingTool::ReduceToLongestStitch(PfoVector &pfoVector, PfoVector &reducedPfoVector,
-    const ParticleFlowObject *const pPfoToEnlarge, const PfoToLArTPCMap &pfoToLArTPCMap) const
+const ParticleFlowObject *StitchingCosmicRayMergingTool::ReduceToLongestStitch(const PfoVector &pfoVector, const ParticleFlowObject *const pPfoToEnlarge,
+    const PfoToLArTPCMap &pfoToLArTPCMap, PfoVector &reducedPfoVector) const
 {
-    if (pfoVector.size() > 2 )
+    if (pfoVector.size() > 2)
     {
-        reducedPfoVector = this->SelectLongestStitch(pfoVector, pfoToLArTPCMap);
+        this->SelectLongestStitch(pfoVector, pfoToLArTPCMap, reducedPfoVector);
         std::sort(reducedPfoVector.begin(), reducedPfoVector.end(), LArPfoHelper::SortByNHits);
 
         return this->GetClosestPfo(pPfoToEnlarge, reducedPfoVector);
     }
     else
     {
-        for (const ParticleFlowObject *const pPfoToShift : pfoVector) reducedPfoVector.push_back(pPfoToShift);
-        std::sort(reducedPfoVector.begin(), reducedPfoVector.end(), LArPfoHelper::SortByNHits);
+        reducedPfoVector = pfoVector;
 
         return pPfoToEnlarge;
     }
@@ -627,36 +629,12 @@ const ParticleFlowObject *StitchingCosmicRayMergingTool::ReduceToLongestStitch(P
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-const ParticleFlowObject *StitchingCosmicRayMergingTool::GetClosestPfo(const ParticleFlowObject *const pPfoToEnlarge, const PfoVector &pfoVector) const
+void StitchingCosmicRayMergingTool::SelectLongestStitch(const PfoVector &pfoVector, const PfoToLArTPCMap &pfoToLArTPCMap, PfoVector &reducedPfoVector) const
 {
-    float minDistance(std::numeric_limits<float>::max());
-    for (const ParticleFlowObject *const pPfoToShift : pfoVector)
-    {
-        if (pPfoToShift == pPfoToEnlarge)
-            return pPfoToEnlarge;
+    //ATTN this is the initialization, in case the rest of logic fails
+    reducedPfoVector = pfoVector;
+    reducedPfoVector.resize(2);
 
-        const float thisDistance(LArPfoHelper::GetTwoDSeparation(pPfoToShift, pPfoToEnlarge));
-        if (thisDistance < minDistance)
-            minDistance = thisDistance;
-    }
-
-    //TODO - better way to avoid this loop? maybe create a map pfo,distance?
-    for (const ParticleFlowObject *const pPfoToShift : pfoVector)
-    {
-        if ((LArPfoHelper::GetTwoDSeparation(pPfoToShift, pPfoToEnlarge) - minDistance) < std::numeric_limits<float>::epsilon())
-            return pPfoToShift;
-    }
-
-    //ATTN: should not arrive here, but if we did, return the first element in pfoVector
-    PfoVector::const_iterator iter = pfoVector.begin();
-    return *iter;
-}
-
-//------------------------------------------------------------------------------------------------------------------------------------------
-
-PfoVector StitchingCosmicRayMergingTool::SelectLongestStitch(PfoVector &pfoVector, const PfoToLArTPCMap &pfoToLArTPCMap) const
-{
-    PfoVector selectedPfos;
     int totalHitsStitched(0);
 
     for (const ParticleFlowObject *const pPfoToShift1 : pfoVector)
@@ -666,7 +644,7 @@ PfoVector StitchingCosmicRayMergingTool::SelectLongestStitch(PfoVector &pfoVecto
             if (pPfoToShift2 == pPfoToShift1)
                 continue;
 
-            const unsigned int twoDHitsPfo1(this->NumberOfTwoDHits(pPfoToShift1)), twoDHitsPfo2(this->NumberOfTwoDHits(pPfoToShift2));
+            const unsigned int twoDHitsPfo1(LArPfoHelper::GetNumberOfTwoDHits(pPfoToShift1)), twoDHitsPfo2(LArPfoHelper::GetNumberOfTwoDHits(pPfoToShift2));
 
             PfoToLArTPCMap::const_iterator iter1(pfoToLArTPCMap.find(pPfoToShift1));
             PfoToLArTPCMap::const_iterator iter2(pfoToLArTPCMap.find(pPfoToShift2));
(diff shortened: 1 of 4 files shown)


larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -537,11 +537,14 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
         for (const ParticleFlowObject *const pPfo : pfoList) pfoVector.push_back(pPfo);
         std::sort(pfoVector.begin(), pfoVector.end(), LArPfoHelper::SortByNHits);
 
+        PfoVector reducedPfoVector;
+        const ParticleFlowObject *const selectedPfoToEnlarge(this->ReduceToLongestStitch(pfoVector, reducedPfoVector, pPfoToEnlarge, pfoToLArTPCMap));
+
```
PandoraPFA: Pandora style guide is all pointer variables start with a small p, e.g. `pSelectedPfoToEnlarge`.

Also, style guide is for const function arguments first in argument list, followed by non-const arguments, e.g. containers that you will fill. `pfoVector` can be passed by const reference here, with only non-const argument being `reducedPfoVector` that you will fill.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -537,11 +537,14 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
         LArTPCPair stitchedLArTPCs(nullptr, nullptr);
 
         try
         {
-            this->FindStitchedLArTPCs(pfoVector, pfoToLArTPCMap, stitchedLArTPCs);
+            this->FindStitchedLArTPCs(reducedPfoVector, pfoToLArTPCMap, stitchedLArTPCs);
```
PandoraPFA: Whilst in this area, can you also address a pending TODO here, and raise an exception if either of the LArTPC pointers in the LArTPCPair remains as nullptr, just before these pointers are dereferenced, so just before `x0` is declared, and `tpcBoundaryCenterX` is calculated. Thanks.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
-	    stitchedPfosToX0Map.insert(PfoToFloatMap::value_type(pPfoToEnlarge, x0));
+            pAlgorithm->StitchPfos(selectedPfoToEnlarge, pPfoToDelete, pfoToLArTPCMap);
+            stitchedPfosToX0Map.insert(PfoToFloatMap::value_type(selectedPfoToEnlarge, x0));
         }
     }
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+const ParticleFlowObject *StitchingCosmicRayMergingTool::ReduceToLongestStitch(PfoVector &pfoVector, PfoVector &reducedPfoVector,
```
PandoraPFA: (See earlier comment about order of arguments and const qualifier for pfoVector reference.)

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+    if (pfoVector.size() > 2 )
+    {
+        reducedPfoVector = this->SelectLongestStitch(pfoVector, pfoToLArTPCMap);
+        std::sort(reducedPfoVector.begin(), reducedPfoVector.end(), LArPfoHelper::SortByNHits);
+
+        return this->GetClosestPfo(pPfoToEnlarge, reducedPfoVector);
+    }
+    else
+    {
+        for (const ParticleFlowObject *const pPfoToShift : pfoVector) reducedPfoVector.push_back(pPfoToShift);
```
PandoraPFA: Cleaner (and probably more efficient) to just assign `pfoVector` to `reducedPfoVector` here. Then sorting is already in place, etc.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
         }
     }
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+const ParticleFlowObject *StitchingCosmicRayMergingTool::ReduceToLongestStitch(PfoVector &pfoVector, PfoVector &reducedPfoVector,
+    const ParticleFlowObject *const pPfoToEnlarge, const PfoToLArTPCMap &pfoToLArTPCMap) const
+{
+    if (pfoVector.size() > 2 )
```
PandoraPFA: Space after 2!

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+
+        return pPfoToEnlarge;
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+const ParticleFlowObject *StitchingCosmicRayMergingTool::GetClosestPfo(const ParticleFlowObject *const pPfoToEnlarge, const PfoVector &pfoVector) const
+{
+    float minDistance(std::numeric_limits<float>::max());
```
PandoraPFA: Usual approach is to declare pointer to a const pfo at start and initialise it to your default return, e.g. `const Pfo *pClosestPfo(pfoVector.begin());` Then have a single loop where you keep track of `minDistance`, and assign to `pClosestPfo` whenever you find a pfo with distance less than min distance. After the single loop, can just return `pClosestPfo`. Can use similar approach for next function too.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+    {
+        for (const ParticleFlowObject *const pPfoToShift : pfoVector) reducedPfoVector.push_back(pPfoToShift);
+        std::sort(reducedPfoVector.begin(), reducedPfoVector.end(), LArPfoHelper::SortByNHits);
+
+        return pPfoToEnlarge;
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
```
PandoraPFA: In Pandora style guide, `SelectLongestStitch` would typically have its implementation before `GetClosestPfo`, because it's called first in `ReduceToLongestStitch`.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+                const LArTPC *const pLArTPC2(iter2->second);
+
+                if (((twoDHitsPfo1 + twoDHitsPfo2) > totalHitsStitched) && LArStitchingHelper::CanTPCsBeStitched(*pLArTPC1, *pLArTPC2))
+                {
+                    totalHitsStitched = twoDHitsPfo1 + twoDHitsPfo2;
+                }
+            }
+        }
+    }
+
```
PandoraPFA: Almost as above, typical approach is to declare a new PfoVector at start and initialise it to your default return (so first two pfos from input pfoVector). Then have a single loop where you keep track of `totalHitsStitched`, and assign to vector whenever you find a stitch involving more hits. After the single loop, can just return the new vector.

Note, we don't often return a vector by value in Pandora, as it's a copy at the end, but you can do this for cleanliness and clarity here if you like, as these will not be big vectors. (Alternative would be to pass by non-const reference, of course).

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+    //ATTN we shouldnt arrive here, but if we did, just resize to 2 (pfoVector is ordered by number of hits anyway)
+    pfoVector.resize(2);
+    return pfoVector;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+//This could be replaced by LArPfoHelper Get2DClusters and then LArClusterHelper GetOrderedCaloHitList, or this below could be moved to LArPfoHelper
+unsigned int StitchingCosmicRayMergingTool::NumberOfTwoDHits(const ParticleFlowObject *pfo) const
```
PandoraPFA: Preferred: `const Pfo *const pPfo`, or `const ParticleFlowObject *const pPfo` - note second const and small p.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+{
+    unsigned int totalHits(0);
+    for (ClusterList::const_iterator iter = pfo->GetClusterList().begin(), iterEnd = pfo->GetClusterList().end(); iter != iterEnd; ++iter)
```
PandoraPFA: Old-style - now prefer range-based for loop, e.g. `for (const Cluster *const pCluster : pfo->GetClusterList())`

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +593,138 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+        }
+    }
+
+    //ATTN we shouldnt arrive here, but if we did, just resize to 2 (pfoVector is ordered by number of hits anyway)
+    pfoVector.resize(2);
+    return pfoVector;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+//This could be replaced by LArPfoHelper Get2DClusters and then LArClusterHelper GetOrderedCaloHitList, or this below could be moved to LArPfoHelper
```
PandoraPFA: Ideal if can find existing helper function to do this, otherwise function like this indeed required! Would normally name as `GetNumberOfTwoDHits` as this is a "get" function.

Even if required, could still use some functions from e.g. `LArPfoHelper` to get 2D clusters and skip the is 3D check. Not big issues at stake here, though, as your function is small and easy to read.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.h
```diff
@@ -199,6 +199,25 @@ class StitchingCosmicRayMergingTool : public StitchingBaseTool
...
+     * @param reducedPfoVector the reduced vector of pfos
+     * @param pPfoToEnlarge the pfo we are enlarging
+     * @param pfoToLArTPCMap the pfo to lar tpc map
+     *
+     * @return particleflow object to enlarge
+     */
+    const pandora::ParticleFlowObject *ReduceToLongestStitch(pandora::PfoVector &pfoVector, pandora::PfoVector &reducedPfoVector,
+        const pandora::ParticleFlowObject *const pPfoToEnlarge, const PfoToLArTPCMap &pfoToLArTPCMap) const;
+
+    // TODO - need doxygen info
```
PandoraPFA: If reorder function implementation, should reorder here too. Doxygen TODO.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: Addressed. The reason why pfoVector was not passed by const reference was because it could not be passed as const to SelectLongestStitch called inside ReduceToLongestStitch, as that one would resize pfoVector to return it if the rest of the logic failed. I agree it is better to have it as constant so I changed the final bit of SelectLongestStitch accordingly.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: I added a check whether one of the two objects in the LArTPCPair is still nullptr and throw a status_code_failure exception here. However, given the logic in FindStitchedLArTPCs, I think the information when a status_code_not_found could be used for that purpose, and inside the catch throw the exception in that case. Nevertheless, extra check outside that try-catch doesn't hurt?

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: thanks, I will add another one! :P

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: yeah, it was easier for me to write it in this way, but forgot to move it afterwards 

PS: Do that style guide actually exist in written form? would be useful to have in documentation...

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: thanks! this comment was very useful, implemented in both

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: implemented, thanks, it looks nicer now :) I have also avoided the return vector by value and instead passed as argument and edit it inside

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
loressa: I think the LArPfoHelper is actually a better place for it to leave, and I have moved it there, changed the name and a few modifications.

loressa:
Hello,

Thanks for the comments, they were very useful! (have not replied to those just "done") 
See updates in commit b2fc14f

Tested visually in the same single gun events and in some protoDUNE MCC12 events, see for example: https://www.hep.phy.cam.ac.uk/~escudero/ProtoDUNEMCC12Event1.png

Thanks,
Lorena

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -590,19 +596,101 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
...
+{
+    //ATTN this is the initialization, in case the rest of logic fails
+    reducedPfoVector = pfoVector;
+    reducedPfoVector.resize(2);
+
+    int totalHitsStitched(0);
+
+    for (const ParticleFlowObject *const pPfoToShift1 : pfoVector)
+    {
+        for (const ParticleFlowObject *const pPfoToShift2 : pfoVector)
```
StevenGreen1: Do these two for loops double count the stitches?  I don't think this will cause a problem because you wipe the reducedPfoVector every time, but it would be more efficient to just loop over all pairs of stitches?

loressa: it is by construction double counted, but only saved once. 
hmm I guess it could be replaced by something like: 
for (iterator it = pfoVector.begin() ; it != pfoVector.end(); ++it) 
   pfo1 = *it;
   for (iterator it2 = it;  ++it2... )

StevenGreen1: Yeah I think you're totally right, it's double counted, but won't change things.  For efficiency purposes I'd suggest doing the change you suggested.  Otherwise the logic looks nice and clean to me and I'm happy with it.  Thanks for doing this!

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
PandoraPFA: It's not strictly necessary now (and this is why the code is as it is), but I still always feel a bit uncomfortable seeing something initialised to nullptr, assigned to in another function and then dereferenced. The only exception, which is common throughout our codebase actually, is where list addresses are returned by reference, and there we always just check the returned status code. But we're there trusting SDK functionality, rather than "just" one function in this tool.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
PandoraPFA: Agreed, the issues at stake are cosmetics and efficiency, with thought about clarity too. I'd have a slight preference for the cosmetic appearance of the initial implementation, but it is of course looping over more combinations of Pfos than is necessary (combinatorics are not big here, though - max one Pfo per volume). What should win is probably the implementation that makes it clearest what you're trying to achieve. Your call!

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.h
```diff
@@ -199,6 +199,38 @@ class StitchingCosmicRayMergingTool : public StitchingBaseTool
...
+
+	/**
+	 * @brief Get the closest pfo to pfoToEnlarge from pfoVector
+	 *
+	 * @param pPfoToEnlarge pfo we search for the closest one to
+	 * @param pfoVector vector of pfos in which to find the closest to pPfoToEnlarge
+	 *
+	 * @return the pfo closest to pPfoToEnlarge from those contained in pfoVector
+	 */
+	const pandora::ParticleFlowObject *GetClosestPfo(const pandora::ParticleFlowObject *const pPfoToEnlarge, const pandora::PfoVector &pfoVector) const;
```
PandoraPFA: Check indentation (might just be display on GitHub, but please check there's no tab characters!). One tab is equivalent to four spaces in our code.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -554,6 +557,9 @@ void StitchingCosmicRayMergingTool::StitchPfos(const MasterAlgorithm *const pAlg
             continue;
         }
 
+    if ((stitchedLArTPCs.first == nullptr) || (stitchedLArTPCs.second == nullptr))
```
PandoraPFA: Check indentation (might just be display on GitHub). Can also have `if (!stitchedLArTPCs.first || !stitchedLArTPCs.second)` if you prefer, but this is fine.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
PandoraPFA: Bottom of slide 16 [here](https://indico.fnal.gov/event/11857/session/11/material/1/0.pdf) is as close as we've come to writing it down formally!

larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -94,6 +94,21 @@ void LArPfoHelper::GetClusters(const ParticleFlowObject *const pPfo, const HitTy
...
+{
+    unsigned int totalHits(0);
+
+    ClusterList clusterList;
+    LArPfoHelper::GetTwoDClusterList(pPfo, clusterList);
+    for (const Cluster *const pCluster : clusterList)
+        totalHits += pCluster->GetNCaloHits();
+
+    return totalHits;
+
```
PandoraPFA: Empty line! :P

PandoraPFA:
Happy to merge when you are ready - many thanks for the feature, which is exactly what we need!

(Next step: update to make multiple stitches, timescale unknown).

Thanks to all involved in the review - this is just fantastic for software development!

loressa:
Thanks! 

I added a commit fixing indentation, etc. I think I will keep the original implementation of the double loop in this occasion. Now going to prepare for large statistics testing. 

Thanks again,
Lorena

StevenGreen1:
Hi Lorena,

Can I ask why you want to keep the double loop? It’s probably not doing anything other than increasing the time it takes for this to run and so would highly recommend removing it unless you know of a good reason to keep it.

Cheers
Steve

loressa:
See latest commit. Visually identical results as of last Friday. Running bigger tests now. 

Lorena

PR #84, opened 2019-07-12
Feature/ProtoDUNE_BDTVertex

StevenGreen1:
Hi, 

These are the necessary changes required to enable the MvaVertexSelectionAlgorithm to work for test beam particles.  I'm current in the process of training/testing this now, but I thought I'd put the PR up so that we resolve any structural issues.  Once the testing is complete I'll put some plots here too showing the performance change.

I thought about making this more complex (e.g. making this a base class and inheriting a neutrino and test beam version), but given the modifications needed to get this to work for a test beam interaction were quite minimal I felt the addition of two new derived classes was overly complicated.  Also it was this format of changes I used for the earlier metrics I showed so thought I'd stick with the simple approach for the sake of pragmatism.  Do let me know if you disagree with that though and if I don't hear anything I'll just post again when the performance plots are ready.

Cheers
Steve

### DIFF
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -143,8 +143,7 @@ void LArMCParticleHelper::GetTrueTestBeamParticles(const MCParticleList *const p
 {
     for (const MCParticle *const pMCParticle : *pMCParticleList)
     {
-        const int nuance(LArMCParticleHelper::GetNuanceCode(pMCParticle));
-        if (LArMCParticleHelper::IsPrimary(pMCParticle) && (nuance == 2001))
+        if (LArMCParticleHelper::IsTriggeredBeamParticle(pMCParticle))
             trueTestBeamParticles.push_back(pMCParticle);
     }
 
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -49,6 +49,14 @@ bool LArMCParticleHelper::IsBeamNeutrinoFinalState(const MCParticle *const pMCPa
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+bool LArMCParticleHelper::IsTriggeredBeamParticle(const MCParticle *const pMCParticle)
+{
+    const int nuance(LArMCParticleHelper::GetNuanceCode(pMCParticle)); 
+    return (LArMCParticleHelper::IsPrimary(pMCParticle) && (nuance == 2001));
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
 bool LArMCParticleHelper::IsBeamParticle(const MCParticle *const pMCParticle)
 {
     const int nuance(LArMCParticleHelper::GetNuanceCode(pMCParticle));
@@ -127,6 +135,20 @@ void LArMCParticleHelper::GetTrueNeutrinos(const MCParticleList *const pMCPartic
     }
 
     std::sort(trueNeutrinos.begin(), trueNeutrinos.end(), LArMCParticleHelper::SortByMomentum);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void LArMCParticleHelper::GetTrueTestBeamParticles(const MCParticleList *const pMCParticleList, MCParticleVector &trueTestBeamParticles)
+{
+    for (const MCParticle *const pMCParticle : *pMCParticleList)
+    {
+        const int nuance(LArMCParticleHelper::GetNuanceCode(pMCParticle));
+        if (LArMCParticleHelper::IsPrimary(pMCParticle) && (nuance == 2001))
+            trueTestBeamParticles.push_back(pMCParticle);
+    }
+
+    std::sort(trueTestBeamParticles.begin(), trueTestBeamParticles.end(), LArMCParticleHelper::SortByMomentum);
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -131,6 +139,20 @@ void LArMCParticleHelper::GetTrueNeutrinos(const MCParticleList *const pMCPartic
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArMCParticleHelper::GetTrueTestBeamParticles(const MCParticleList *const pMCParticleList, MCParticleVector &trueTestBeamParticles)
+{
+    for (const MCParticle *const pMCParticle : *pMCParticleList)
+    {
```
PandoraPFA: This could call `LArMCParticleHelper::IsTriggeredBeamParticle` so that the definition, with its use of a specific nuance value, etc., is implemented in only one function.

larpandoracontent/LArVertex/MvaVertexSelectionAlgorithm.cc
```diff
@@ -614,12 +615,18 @@ std::string MvaVertexSelectionAlgorithm<T>::GetInteractionType() const
     PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::GetList(*this, m_caloHitListName, pCaloHitList));
 
     // ATTN Assumes single neutrino is parent of all neutrino-induced mc particles
-    LArMCParticleHelper::MCContributionMap nuMCParticlesToGoodHitsMap;
+    LArMCParticleHelper::MCContributionMap targetMCParticlesToGoodHitsMap;
     LArMCParticleHelper::SelectReconstructableMCParticles(pMCParticleList, pCaloHitList, LArMCParticleHelper::PrimaryParameters(),
-        LArMCParticleHelper::IsBeamNeutrinoFinalState, nuMCParticlesToGoodHitsMap);
+        LArMCParticleHelper::IsBeamNeutrinoFinalState, targetMCParticlesToGoodHitsMap);
+
+    if (m_testBeamMode)
```
PandoraPFA: Should this be an `if (m_testBeamMode) {} else {}` approach, with the IsBeamNeutrinoFinalState check in the else block, rather than before the if?

larpandoracontent/LArVertex/MvaVertexSelectionAlgorithm.cc
```diff
@@ -680,19 +687,30 @@ void MvaVertexSelectionAlgorithm<T>::GetBestVertex(const VertexVector &vertexVec
     const MCParticleList *pMCParticleList(nullptr);
     PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::GetList(*this, m_mcParticleListName, pMCParticleList));
 
-    // Obtain vector: true neutrinos
-    MCParticleVector mcNeutrinoVector;
-    LArMCParticleHelper::GetTrueNeutrinos(pMCParticleList, mcNeutrinoVector);
+    // Obtain vector: true targets
+    MCParticleVector mcTargetVector;
+    LArMCParticleHelper::GetTrueNeutrinos(pMCParticleList, mcTargetVector);
```
PandoraPFA: Should this be removed, as it's also in the else block, below (which seems to be the correct location)?

larpandoracontent/LArVertex/MvaVertexSelectionAlgorithm.cc
```diff
@@ -680,19 +687,30 @@ void MvaVertexSelectionAlgorithm<T>::GetBestVertex(const VertexVector &vertexVec
...
 
     for (const Vertex *const pVertex : vertexVector)
     {
         float mcVertexDr(std::numeric_limits<float>::max());
-        for (const MCParticle *const pMCNeutrino : mcNeutrinoVector)
+        for (const MCParticle *const pMCTarget : mcTargetVector)
         {
-            const CartesianVector mcNeutrinoPosition(pMCNeutrino->GetEndpoint().GetX() + m_mcVertexXCorrection, pMCNeutrino->GetEndpoint().GetY(),
-                pMCNeutrino->GetEndpoint().GetZ());
+            const CartesianVector mcTargetPosition((m_testBeamMode && std::fabs(pMCTarget->GetParticleId()) == 11) ? pMCTarget->GetVertex() :
```
PandoraPFA: This could be a const reference (relatively small gain here, but probably worthwhile)

larpandoracontent/LArVertex/MvaVertexSelectionAlgorithm.cc
StevenGreen1: Hi, Yeah that approach would work just as well.  In theory it will make no difference as there should only ever be either a triggered beam particle or a neutrino in a given event, but it's probably cleaner to divide the modes up as you say.

larpandoracontent/LArVertex/MvaVertexSelectionAlgorithm.cc
StevenGreen1: That's a bad one!  Good spot.

StevenGreen1:
Hi John, 

Thanks for the feedback.  I've applied the sensible changes you suggested.  I'm still testing the BDT vertex performance, so I think we can hold back on merging this just yet (or indeed until there's a quieter week), but I do expect this to be fine now.  Will add some results once I've got a nicely trained BDT vertex model working. 

Cheers
Steve

PandoraPFA:
Perfect, thank you!

StevenGreen1:
Hi,

Here's some plots showing the performance of the BDT vertex.  Everything looks good and I'll merge the PR now.

Cheers, Steve 

BeamParticleVertexDeltaRProton
BeamParticleVertexDeltaRPositron
BeamParticleVertexDeltaRPiPlus
BeamParticleVertexDeltaRKaonPlus
BeamParticleVertexDeltaRAntiMuon
BeamParticleVertexDeltaR

PR #86, opened 2019-07-18
Shower Overlap Result Comparison

StevenGreen1:
Extending the variables used in the comparison operator of the LArShowerOverlapResult class.  This resolves a reproducibility issue noted in certain DUNEFD and UBooNE events.

### DIFF
--- a/larpandoracontent/LArObjects/LArShowerOverlapResult.cc
+++ b/larpandoracontent/LArObjects/LArShowerOverlapResult.cc
@@ -83,13 +83,13 @@ bool ShowerOverlapResult::operator<(const ShowerOverlapResult &rhs) const
     if (std::fabs(m_matchedFraction - rhs.m_matchedFraction) > std::numeric_limits<float>::epsilon())
         return (m_matchedFraction < rhs.m_matchedFraction);
 
+    if (std::fabs(m_xOverlap.GetXSpanW() - rhs.m_xOverlap.GetXSpanW()) > std::numeric_limits<float>::epsilon())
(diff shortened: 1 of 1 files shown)


PandoraPFA:
Thanks, that looks fine. All we can do without adding new properties to the overlap result or rethinking the overall approach (not on the cards right now). Please feel free to merge, from my perspective.

loressa:
Hello,

I think I would use the span in collection plane (W) before the other induction planes. You can check the sizes of those overlap planes in general for each plane, I suspect W would always give a better discrimination (less detector gaps) 

Thanks
Lorena

PandoraPFA:
This seems to only affect something like one in ten thousand shower overlap results, so I don't have any strong opinions about ordering. This is just to help ensure ordering is always strictly defined, so either (any) ordering of checks works for me. Presumably there's absolutely no performance impact? Could just make the change so we have a consensus, or leave it as is for expediency!

loressa:
Sure, it probably won't affect much, but I am just thinking would be nice to rely more on the view usually gives us more information (counting gaps & geometrically -angle-)

StevenGreen1:
Hi Lorena, John,

Thanks for taking a look at this. I think Lorena makes a good point that flipping the checking to start with W should give a bit more discriminating power as the span should be bigger for particles travelling along z. Therefore I’ll implement that change today. Johns also right in that this will only affect a tiny number of events but it should still be a performance gain so is worth doing. 

If there are no further comments I’ll also merge this in later today so I can continue with the in LArSoft testing. 

Cheers

PR #87, opened 2019-07-23
MC To Pfo Matches

StevenGreen1:
Hi,

The aim of this PR is to ensure the mcToPfoHitSharingMap is populated for all mc primaries, even if no pfos are associated to it.  This resolves an unknown exception being thrown in downstream logic if m_printAllToScreen is true.

This change opts to add in the null entry into mcToPfoHitSharingMap early on rather than check the MCParticle is present in the map downstream because this mirrors what is done in the interpreted matching where a null entry is added if no good match can be made.  

I'm also open to other suggestions for changes instead, just let me know what you think.
Cheers
Steve

### DIFF
--- a/larpandoracontent/LArHelpers/LArMonitoringHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMonitoringHelper.cc
@@ -240,11 +240,13 @@ void LArMonitoringHelper::PrintMatchingTable(const PfoVector &orderedPfoVector, 
     // Make a new row for each MCParticle
     for (unsigned int mcParticleId = 0; mcParticleId < orderedMCParticleVector.size(); ++mcParticleId)
     {
-        LArMCParticleHelper::MCParticleToPfoHitSharingMap::const_iterator it = mcParticleToPfoHitSharingMap.find(orderedMCParticleVector.at(mcParticleId));
-        if (it == mcParticleToPfoHitSharingMap.end())
-            throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+        const MCParticle *const pMCParticle(orderedMCParticleVector.at(mcParticleId));
+        LArMCParticleHelper::MCParticleToPfoHitSharingMap::const_iterator it = mcParticleToPfoHitSharingMap.find(pMCParticle);
+        LArMCParticleHelper::PfoToSharedHitsVector pfoToSharedHitsVector;
 
-        const MCParticle *const pMCParticle(it->first);
+        if (it != mcParticleToPfoHitSharingMap.end())
+            pfoToSharedHitsVector = it->second;
+
         const LArFormattingHelper::Color mcCol(
             LArMCParticleHelper::IsBeamNeutrinoFinalState(pMCParticle) ?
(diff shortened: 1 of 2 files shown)


StevenGreen1:
Hi, 

As the addition of MC primaries to the map is only needed because of the way that the event validation algorithm handles the map downstream, there is no obvious need to add the primaries in by default in the LArMCParticleHelper.  However, adding them in would not be explicitly wrong either.  As, for the case of an MCPrimary with no associated PFOs, I would not expect the MCParticle to appear in these maps, I will leave the addition of the primaries in the event validation algorithm for now, but this can be reviewed at a later date if needed.

Cheers
Steve

loressa:
Hi,

Whilst I understand this is done in this function only to solve a single, localized problem, I got the same feeling than John, that it could be added upstream into LArMCParticleHelper::GetPfoMCParticleHitSharingMaps. 

That would need more work in terms of checking the output map of mc -pfo hit matching is correctly used with the additional entries with no matching pfo, but having had a quick look at its usage in other places, this seems to possibly make sense in general. 

If you see for example how its used in LArMonitoring::PfoValidationAlgorithm: 
https://github.com/PandoraPFA/LArContent/blob/ae636e8a5cf650308de9b56acd0920abb1039c3c/larpandoracontent/LArMonitoring/PfoValidationAlgorithm.cc#L66
the map is then used in LArMonitoringHelper::PrintMatchingTable here: 
https://github.com/PandoraPFA/LArContent/blob/ae636e8a5cf650308de9b56acd0920abb1039c3c/larpandoracontent/LArMonitoring/PfoValidationAlgorithm.cc#L90
when its receiving this map, but a list of orderedMCParticles from LArMonitoringHelper::GetOrderedMCParticleVector, which will contain MC particles with good hits produced independently on whether they have a match or not. I dont know if we use this PrintMatchingTable at all or not, but I have the feeling that for the same reason causing the exception motivating this PR, in this example it will throw an exception here: 
https://github.com/PandoraPFA/LArContent/blob/ae636e8a5cf650308de9b56acd0920abb1039c3c/larpandoracontent/LArHelpers/LArMonitoringHelper.cc#L245

So a change like the one implemented here, subject to testing and checking usage in other places, might solve other exceptions. 

Best,
Lorena

StevenGreen1:
Hi, 

Lorena and I discussed this in person and we noted that the only remaining place inside Pandora where the LArMCParticleHelper::GetPfoMCParticleHitSharingMaps function is called and the mcParticleToPfoHitSharingMap used is inside the LArMonitoringHelper::PrintMatchingTable.  Lorena highlighted in the above comment that if the target MCParticle is not present in the mcParticleToPfoHitSharingMap, this function threw an uncaught exception.  As there is no need for an exception to be thrown in this function, as it simply means the target MCParticle does not have any reconstructed matches, the LArMonitoringHelper::PrintMatchingTable function has been edited to display no matches instead of throwing an exception.  

This change was tested on an event where this problem occurred and it resolved the issue.  Furthermore we discussed the code implementation and agreed it was sound and will be merged into the master branch shortly. 

Cheers,
Steve

StevenGreen1:
Merged into master after rebasing.

PR #91, opened 2019-08-09
Return Type Check

StevenGreen1:
Check return type of GetGlobalFitDirection function in shower asymmetry feature tool in response to Coverity highlighted issue.

If the fit direction cannot be found the loop continues in search of another shower like cluster to base the calculation of the asymmetry score around.  In case of no matching clusters the default asymmetry score, 1.f, is used.  Previously, in the case of an unsuccessful fit, a null fit direction (0,0,0) was used to calculate the asymmetry score, which is not desired behaviour.

I looked into throwing an exception too, but there are no try-catch blocks higher up in the calling hierarchy that would work nicely for that approach.  Therefore, as this tool just calculates a single variable for the ML vertexing approach and it's still possible to proceed with the reconstruction even if this one variables isn't set properly, I decided to go with persisting the default value.

### DIFF
--- a/larpandoracontent/LArVertex/ShowerAsymmetryFeatureTool.cc
+++ b/larpandoracontent/LArVertex/ShowerAsymmetryFeatureTool.cc
@@ -61,7 +61,8 @@ float ShowerAsymmetryFeatureTool::GetShowerAsymmetryForView(const CartesianVecto
             showerFit.GetLocalPosition(vertexPosition2D, rL, rT);
 
             CartesianVector showerDirection(0.f, 0.f, 0.f);
-            showerFit.GetGlobalFitDirection(rL, showerDirection);
+            if (STATUS_CODE_SUCCESS != showerFit.GetGlobalFitDirection(rL, showerDirection))
+                continue;
 
             const float projectedVtxPosition = vertexPosition2D.GetDotProduct(showerDirection);
 


PandoraPFA:
Looks fine, thanks! The only decision really is what to do if a bad status code is returned, and continuing here looks sensible. FYI the sliding linear fit functions all used to raise exceptions, but it turned out that the way it was all used meant that raising frequently, rather than returning frequently was quite a large contribution to the run time! The question now is whether there's any change in behaviour, but the modification should definitely go in. Good to see that the Coverity fix got things up and running again too!

StevenGreen1:
Yeah I haven't tested this extensively, which would be sensible to do.  Any change in behaviour must be due to it now using an asymmetry score of 1.f rather than proceeding with a bogus fit direction, so even if there is a change it's arguable that the new result would be better (or at least well defined!).  I don't think I'll be able to get any testing done in the short term as I'm a bit busy, but I'll add it to the to do list. We can merge this now though if @loressa is happy too as it will remove the highlighted Coverity issue. Cheers.

StevenGreen1:
BeamParticleCompleteness
BeamParticleEfficiencyVsNHits
BeamParticlePurity
CosmicRayCompleteness
CosmicRayEfficiencyVsNHits
CosmicRayPurity

I've tested the impact of this PR and it doesn't seem to have any significant impact on the performance, which is good.  I'll merge this in via the command line shortly. Cheers, Steve

PR #94, opened 2019-09-11
Add pfo variable is test beam hiererachy to tree.

StevenGreen1:
Add a variable to the root output tree indicating whether a best matched pfo is part of the test beam hierarchy.


StevenGreen1:
Thanks for the feedback.  Rebased and merged via command line.

PR #95, opened 2019-09-27
Feature/test beam vertex persistence

StevenGreen1:
These changes set both the interaction and start vertices for the parent test beam pfo.  The alterations to the GetVertex function account for this fact and ensure default behaviour is retained for all other particles.  In the case of the parent test beam pfo, the default vertex returned by GetVertex is the start vertex as this is needed for fitting for directions downstream and this behaviour should be retained.  There is now also a GetTestBeamInteraction vertex function that returns the interaction vertex and this will be used in a modification to LArPandora to save this vertex in a separate collection.

This PR will be tested shortly and results added but the conceptual review can begin immediately.  Cheers, Steve

### DIFF
--- a/larpandoracontent/LArHelpers/LArPfoHelper.cc
+++ b/larpandoracontent/LArHelpers/LArPfoHelper.cc
@@ -477,7 +477,7 @@ const Vertex *LArPfoHelper::GetVertex(const ParticleFlowObject *const pPfo)
     // ATTN : Test beam parent pfos contain an interaction and start vertex
     if (LArPfoHelper::IsTestBeam(pPfo) && pPfo->GetParentPfoList().empty())
     {
-        pVertex = LArPfoHelper::GetVertexType(pPfo->GetVertexList(), VERTEX_START);
+        pVertex = LArPfoHelper::GetVertexWithLabel(pPfo->GetVertexList(), VERTEX_START);
     }
     else
     {
@@ -498,9 +498,9 @@ const Vertex *LArPfoHelper::GetVertex(const ParticleFlowObject *const pPfo)
 const Vertex *LArPfoHelper::GetTestBeamInteractionVertex(const ParticleFlowObject *const pPfo)
 {
     if (pPfo->GetVertexList().empty() || !pPfo->GetParentPfoList().empty() || !LArPfoHelper::IsTestBeam(pPfo))
-        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+        throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
 
-    const Vertex *pInteractionVertex(LArPfoHelper::GetVertexType(pPfo->GetVertexList(), VERTEX_INTERACTION));
+    const Vertex *pInteractionVertex(LArPfoHelper::GetVertexWithLabel(pPfo->GetVertexList(), VERTEX_INTERACTION));
 
     if (VERTEX_3D != pInteractionVertex->GetVertexType())
         throw StatusCodeException(STATUS_CODE_FAILURE);
@@ -510,7 +510,7 @@ const Vertex *LArPfoHelper::GetTestBeamInteractionVertex(const ParticleFlowObjec
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-const Vertex *LArPfoHelper::GetVertexType(const VertexList &vertexList, const VertexLabel &vertexLabel)
+const Vertex *LArPfoHelper::GetVertexWithLabel(const VertexList &vertexList, const VertexLabel vertexLabel)
 {
     const Vertex *pTargetVertex(nullptr);
 
--- a/larpandoracontent/LArHelpers/LArPfoHelper.cc
+++ b/larpandoracontent/LArHelpers/LArPfoHelper.cc
@@ -472,15 +472,67 @@ const Vertex *LArPfoHelper::GetVertex(const ParticleFlowObject *const pPfo)
     if (pPfo->GetVertexList().empty())
         throw StatusCodeException(STATUS_CODE_NOT_FOUND);
 
-    if (pPfo->GetVertexList().size() != 1)
-        throw StatusCodeException(STATUS_CODE_FAILURE);
+    const Vertex *pVertex(nullptr);
 
-    const Vertex *const pVertex = *(pPfo->GetVertexList().begin());
+    // ATTN : Test beam parent pfos contain an interaction and start vertex
+    if (LArPfoHelper::IsTestBeam(pPfo) && pPfo->GetParentPfoList().empty())
+    {
+        pVertex = LArPfoHelper::GetVertexType(pPfo->GetVertexList(), VERTEX_START);
+    }
+    else
+    {
+        if (pPfo->GetVertexList().size() != 1)
+            throw StatusCodeException(STATUS_CODE_FAILURE);
+
+        pVertex = *(pPfo->GetVertexList().begin());
+    }
 
     if (VERTEX_3D != pVertex->GetVertexType())
         throw StatusCodeException(STATUS_CODE_FAILURE);
 
     return pVertex;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+const Vertex *LArPfoHelper::GetTestBeamInteractionVertex(const ParticleFlowObject *const pPfo)
+{
+    if (pPfo->GetVertexList().empty() || !pPfo->GetParentPfoList().empty() || !LArPfoHelper::IsTestBeam(pPfo))
+        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+
+    const Vertex *pInteractionVertex(LArPfoHelper::GetVertexType(pPfo->GetVertexList(), VERTEX_INTERACTION));
+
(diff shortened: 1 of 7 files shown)


larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -485,6 +495,48 @@ const Vertex *LArPfoHelper::GetVertex(const ParticleFlowObject *const pPfo)
...
+
+    if (VERTEX_3D != pInteractionVertex->GetVertexType())
+        throw StatusCodeException(STATUS_CODE_FAILURE);
+
+    return pInteractionVertex;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+const Vertex *LArPfoHelper::GetVertexType(const VertexList &vertexList, const VertexLabel &vertexLabel)
```
PandoraPFA: Better to call this `GetVertexOfLabel` or `GetVertexWithLabel`, or something that communicates that we're accessing via VertexLabel, rather than via VertexType (our two distinguishing enumerated types).

Worry what happens if there's multiple vertices with the same label in the list. Here you'll get the last vertex with that label. Is that OK? No current use-case of course, but should be prepared!

VertexLabel argument should be by value, rather than reference (it's just an enumerated type).

larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -485,6 +495,48 @@ const Vertex *LArPfoHelper::GetVertex(const ParticleFlowObject *const pPfo)
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+const Vertex *LArPfoHelper::GetTestBeamInteractionVertex(const ParticleFlowObject *const pPfo)
+{
+    if (pPfo->GetVertexList().empty() || !pPfo->GetParentPfoList().empty() || !LArPfoHelper::IsTestBeam(pPfo))
+        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
```
PandoraPFA: STATUS_CODE_INVALID_PARAMETER maybe more appropriate?

larpandoracontent/LArHelpers/LArPfoHelper.h
```diff
@@ -311,6 +311,25 @@ class LArPfoHelper
...
+
+    /**
+     *  @brief  Get the vertex with a specific vertex label in a given vertex list
+     *
+     *  @param  vertexList vertex list
+     *  @param  vertexLabel target vertex label type
+     *
+     *  @return address of the desired vertex
+     */
+    static const pandora::Vertex *GetVertexType(const pandora::VertexList &vertexList, const pandora::VertexLabel &vertexLabel);
```
PandoraPFA: See comments above regarding function name and arguments.

larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -472,10 +472,20 @@ const Vertex *LArPfoHelper::GetVertex(const ParticleFlowObject *const pPfo)
...
+    {
+        pVertex = LArPfoHelper::GetVertexType(pPfo->GetVertexList(), VERTEX_START);
+    }
+    else
+    {
+        if (pPfo->GetVertexList().size() != 1)
+            throw StatusCodeException(STATUS_CODE_FAILURE);
+
+        pVertex = *(pPfo->GetVertexList().begin());
+    }
```
PandoraPFA: We could make this cleaner if we risk some restructuring (not clear whether we should do this):

If the parent pfo list is empty, could call the same function, but with different labels: VERTEX_START for test-beam, and VERTEX_INTERACTION for other cases.

If the parent pfo list isn't empty, could still just ask for the first vertex, or know that it's label should be VERTEX_INTERACTION again (I actually think these particles should now have vertices with label VERTEX_START).

Would we still want to enforce the size of vertex list being 1 for everything except test beam particles?

Is this all too fiddly?

larpandoracontent/LArThreeDReco/LArEventBuilding/TestBeamParticleCreationAlgorithm.cc
```diff
@@ -118,38 +110,34 @@ StatusCode TestBeamParticleCreationAlgorithm::SetupTestBeamVertex(const Pfo *con
...
-        const Vertex *pTestBeamStartVertex(nullptr);
-        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::Vertex::Create(*this, parameters, pTestBeamStartVertex));
-
-        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::SaveList<Vertex>(*this, m_parentVertexListName));
-        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ReplaceCurrentList<Vertex>(*this, m_parentVertexListName));
-        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::AddToPfo(*this, pTestBeamPfo, pTestBeamStartVertex));
+        const Vertex *const pNuVertex(LArPfoHelper::GetVertex(pNuPfo));
+        PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::RemoveFromPfo(*this, pNuPfo, pNuVertex));
+        PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::AddToPfo(*this, pTestBeamPfo, pNuVertex));
+        // ATTN This vertex already lives in m_parentVertexListName
```
PandoraPFA: Handy comment!

larpandoracontent/LArThreeDReco/LArEventBuilding/TestBeamParticleCreationAlgorithm.cc
```diff
@@ -118,38 +110,34 @@ StatusCode TestBeamParticleCreationAlgorithm::SetupTestBeamVertex(const Pfo *con
         std::cout << "TestBeamParticleCreationAlgorithm::SetupTestBeamVertex - Test beam particle has no initial vertex" << std::endl;
     }
 
-    if (m_keepStartVertex)
+    // Set start vertex
```
PandoraPFA: This all seems OK, but definitely relying on your testing here too!

larpandoracontent/LArHelpers/LArPfoHelper.cc
StevenGreen1: Hi John,  Good point about the naming scheme.  I will update that.  If there's multiple vertices of the same time in the list passed to this function it throws an exception rather than returning the last two.  I made that decision as I don't think there should ever be a case where a particle has multiple types of the same vertex.  Will also remove the reference too as you're right it doesn't save any overheads.

larpandoracontent/LArHelpers/LArPfoHelper.cc
StevenGreen1: I think you're right.  I didn't particularly like this bit of the function as it is ugly, but it was necessary because we've been assuming all PFOs have a single vertex everywhere else in the reconstruction.  I also agree that VERTEX_START makes more sense too for everything apart form the neutrino/test beam parent.  

From a pragmatic sense I'm tempted to leave things as they are for now to avoid the fiddly-ness and have a review of the use of vertices.  Long term this will be needed when Pandora tried to do the reconstruct about secondary interaction vertices anyway.  With that in mind I wouldn't enforce a fixed vertex list size because the start, end and interaction are all distinct valid vertices that might be of interest.  If you think it's worth doing that sooner rather than later I'm happy to give the review a go, but it might be sensible to make that a separate development from this also.  What do you think it best to do going forward with this one?

larpandoracontent/LArHelpers/LArPfoHelper.cc
StevenGreen1: Sure thing.

larpandoracontent/LArThreeDReco/LArEventBuilding/TestBeamParticleCreationAlgorithm.cc
StevenGreen1: You did write it!  :)

StevenGreen1:
I've just finished testing this PR and is makes no difference to the reconstruction of particles in the hierarchies (the dotted lines all sit on top of the solid ones, it's not that they're missing!):

PrimaryRecoEfficiencyVsNHits
PrimaryCompleteness
PrimaryPurity

There is a difference in the vertexing for the interaction, but that's due to the fact that the interaction vertex previously was being set as the vertex for the last daughter in the list of MCParticles in the hierarchy (an oversight on my part previously).  Having fixed this it doesn't seem to change things too much.  Also for context, there's no space charge correction applied here so the resolution is far away from peaking at 0, but I just wanted some plot to demonstrate that there has been a change.  I should also say that we've never looked at the reco vertex using the test beam hierarchy algorithm before and so we don't need to worry about any plots made previously being wrong.  

Vertex

If you're happy I'll merge this and can tag a release for next week too?
Cheers
Steve

StevenGreen1:
Merged via command line.

PR #96, opened 2019-10-06
Feature/vertex cheating mode

loressa:
Simplified version of feature/VertexLegacy, with only the addition of the new vertex cheating mode, using the best vertex candidate (3D vertex candidate created closer to true position), rather than the true vertex position.


loressa:
I implemented the suggested requests, which will follow in a clean PR, closing this one. 

Lorena

PR #97, opened 2019-10-07
Feature/vertex cheating mode clean

loressa:
New vertex cheating mode, using the best vertex candidate (3D vertex candidate created closer to true position), rather than the true vertex position


StevenGreen1:
Feature included in the vertex refactoring  pull request.

PR #98, opened 2019-10-08
Refactoring the vertex selection algorithms.

StevenGreen1:
Implementation of a restructuring of the vertex selection algorithms within Pandora.  A new vertex training algorithm class has been introduced that inherits from the base and is used to construct the derived MVA vertex selection algorithm.  In addition to this a cheated vertex selection algorithm has been added that inherits from the training algorithm class.  This cheated selection picks the best vertex (if such a vertex exists) from those candidate vertices provided.


StevenGreen1:
First metric produced compares the restructured vertex resolution (delta R, 3D) for the MVA and cheated vertex selection using this feature branch.  The validation jobs are running using the current default vertex selection code and will be used to compare to this new performance once finished.  However, as a first pass this new cheating algorithm appears to be working as expected.

Vertex

StevenGreen1:
Comparing the vertex resolution using the MVA algorithm from the current reconstruction and this PR gives identical performance.

Fig
The samples used in making this plot are 1, 3, 5 and 7 GeV ProtoDUNE-SP beam and cosmic events without space charge.

StevenGreen1:
Merged after final testing completed.

PR #99, opened 2019-10-11
Feature/AdjustVertexPosition

PandoraPFA:
This feature should not be triggered often, but responds to some MicroBooNE events in each of which there was a large cosmic-ray remnant in one view of the slice. The remnants were located a significant distance from the neutrino interaction in the same slice (and, importantly at lower z), but there are sufficient connecting hits (and basically nowhere else for the cosmic remnants to go), so it's understandable that they ended up in the same slice.

The cosmic remnant, plus sufficient random hits in other views, creates vertex candidates at low z, and then one of these candidates is selected (the low z leads to a favourable beam score, and the significant remnant cluster is a major factor in driving a good candidate score).

Now we get to the key part: because the remnant only features in one view, it doesn't get make into a Pfo, but the downstream neutrino interaction does. We therefore end up with a neutrino interaction vertex displaced from the reconstructed daughter Pfos.

This PR involves the NeutrinoHierarchy algorithm. If found that there are no vertex-associated Pfos, the neutrino interaction vertex is relocated, using a simple scheme for now: the closest 3D position in the largest (most hits) daughter Pfo. This fixes all known problem events so far, but sophistication could be improved in the longer term.

E.g. VHits in slice 1, note cosmic remnant present only in one view in this slice (wire position up page; drift position across page):

VHitsSlice1_7004_237_11854

E.g. WHits and old vertex position, with significant offset:

WHitsAndOldVertex_7004_237_11854

E.g. Whits and new vertex position, no issue:

WHitsAndNewVertex_7004_237_11854

Obviously some soak testing will be required before this can be merged too. Expected impact at ProtoDUNE is negligible, but this should also be tested via e.g. observing unchanged vertex resolution plots.

Thanks!

### DIFF
--- a/larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
+++ b/larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
@@ -236,8 +236,9 @@ void NeutrinoHierarchyAlgorithm::AdjustVertexAndPfoInfo(const ParticleFlowObject
     LArPfoHelper::GetThreeDClusterList(candidateDaughterPfoVector.front(), daughterClusterList3D);
     daughterClusterList3D.sort(LArClusterHelper::SortByNHits);
 
+    // If no 3D hits, must leave vertex where it was
     if (daughterClusterList3D.empty())
-        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+        return;
 
     const Vertex *pOldNeutrinoVertex(LArPfoHelper::GetVertex(pNeutrinoPfo));
     const CartesianVector newVertexPosition(LArClusterHelper::GetClosestPosition(pOldNeutrinoVertex->GetPosition(), daughterClusterList3D.front()));
--- a/larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
+++ b/larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
@@ -234,6 +234,7 @@ void NeutrinoHierarchyAlgorithm::AdjustVertexAndPfoInfo(const ParticleFlowObject
 
     ClusterList daughterClusterList3D;
     LArPfoHelper::GetThreeDClusterList(candidateDaughterPfoVector.front(), daughterClusterList3D);
+    daughterClusterList3D.sort(LArClusterHelper::SortByNHits);
 
     if (daughterClusterList3D.empty())
         throw StatusCodeException(STATUS_CODE_NOT_FOUND);
--- a/larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
+++ b/larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
@@ -187,9 +187,8 @@ void NeutrinoHierarchyAlgorithm::ProcessPfoInfoMap(const ParticleFlowObject *con
     }
 
     // Handle exceptional case where vertex selected far from any other particle passing creation thresholds
-    if (0 == callDepth && 0 == pNeutrinoPfo->GetNDaughterPfos())
+    if ((0 == callDepth) && (0 == pNeutrinoPfo->GetNDaughterPfos()))
     {
-std::cout << "No vertex pfo associations. Move the vertex to match that of largest pfo." << std::endl;
         this->AdjustVertexAndPfoInfo(pNeutrinoPfo, candidateDaughterPfoList, pfoInfoMap);
         return this->ProcessPfoInfoMap(pNeutrinoPfo, candidateDaughterPfoList, pfoInfoMap, callDepth + 1);
     }
@@ -229,7 +228,7 @@ void NeutrinoHierarchyAlgorithm::AdjustVertexAndPfoInfo(const ParticleFlowObject
     PfoVector candidateDaughterPfoVector(candidateDaughterPfoList.begin(), candidateDaughterPfoList.end());
     std::sort(candidateDaughterPfoVector.begin(), candidateDaughterPfoVector.end(), LArPfoHelper::SortByNHits);
 
-    // TODO Consider deleting the neutrino pfo if there are no daughter pfo candidates?
+    // TODO Consider deleting the neutrino pfo if there are no daughter pfo candidates
     if (candidateDaughterPfoVector.empty())
         throw StatusCodeException(STATUS_CODE_NOT_FOUND);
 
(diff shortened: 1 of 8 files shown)


StevenGreen1:
Hi John,

Thanks for this PR.  I will review the code now, but I had a conceptual question first.  Is there nothing in the current vertexing that says if a vertex is really far away from hits in any of the views, then it's unlikely to be the actual vertex (working under the assumption that slicing has worked ok and we have a good neutrino interaction completeness in the slice)?  It seems like, if that's not the case, then we might want to add something as it could prove useful.

Cheers
Steve

larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
```diff
@@ -214,6 +222,53 @@ void NeutrinoHierarchyAlgorithm::ProcessPfoInfoMap(const ParticleFlowObject *con
...
+        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+
+    ClusterList daughterClusterList3D;
+    LArPfoHelper::GetThreeDClusterList(candidateDaughterPfoVector.front(), daughterClusterList3D);
+
+    if (daughterClusterList3D.empty())
+        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+
+    const Vertex *pOldNeutrinoVertex(LArPfoHelper::GetVertex(pNeutrinoPfo));
+    const CartesianVector newVertexPosition(LArClusterHelper::GetClosestPosition(pOldNeutrinoVertex->GetPosition(), daughterClusterList3D.front()));
```
StevenGreen1: Am I wrong in thinking that the newVertexPosition is semi-arbitrary because it is set to the closest distance between the old vertex and whichever 3D cluster is at the front of the unsorted daughterClusterList3D list?  Should we compare to all 3D clusters in this list instead?

StevenGreen1: I now see in the ordering that the PFO used to get the 3D clusters is the one with the largest number of PFOs.  Is there any mechanism by which a PFO can have more than one 3D cluster at this stage?  If not then this is fine, but it might be worth adding a comment saying something along the lines of this is the largest 3D cluster in the code as it wasn't immediately obvious when reading through (although your git comment is very nice and clear).

StevenGreen1:
Hi John,

I've gone over the code and other than the one comment I made about possibly adding in a comment saying largest 3D cluster (and the one conceptual comment) I'm happy with this PR.  Adding a plot or two for performance to this PR would be fab and once that's done feel free to merge it whenever you like.

Cheers
Steve

larpandoracontent/LArThreeDReco/LArEventBuilding/NeutrinoHierarchyAlgorithm.cc
PandoraPFA: Hi Steve,

Thanks - you're right that this has ordered the pfos and got the first 3D cluster (which is from a list that holds its order but isn't explicitly sorted), so there's an implicit assumption that there will only be one 3D cluster per pfo.

Right now, I think every pfo will have maximum one cluster of each of the following hit types: u, v, w, 3D. I'm not aware of any that have two of a given hit type (would be due to an error), but it also wouldn't hurt to add a quick sort of the 3D clusters in this function anyhow.

Would this work for you?

StevenGreen1: Adding a sort would make things very clear and help the readability I think, but I'll leave the final decision up to you because a comment would equally help that aspect and save having to call the albeit trivial sort.  But there's also the issue of comments going out of date and so I can see benefits either way.

PandoraPFA:
Hi Steve,

Thanks again. In terms of your initial query. We ensure that all vertex candidates are motivated by the presence of clusters in at least two views. We also demand that all candidates sit on/near a hit or in a gap in all three views.

In the example events, there were sufficient stray hits to ensure all these conditions were met. The selection of the vertex then, of course, uses the features that are designed to help pick out good interaction vertex candidates. The presence of a large remnant in one view, plus the lower z coordinate, seems to ensure that this candidate is selected.

Later, it turns out that no pfo is actually built near the selected candidate, but one/more are built a significant distance away. This is when this modification steps in and does something relatively simple (now) to recover the situation.

What particular performance figures do you think we need here. I have run (but not yet had time to check) the standard uboone assessments without cosmics (so don't expect any impact there really). Let's agree some metrics to say this is OK and if you could check your ProtoDUNE test setup too, that'd be great.

Thinking about ProtoDUNE, it's unlikely there'd be beam particles far downstream of the start of the detector and (a cosmic remnant there). It might be possible for there to be a very big cosmic at higher z in the same slice as the beam particle, but the z-aspect helps prevent choosing the interaction vertex location at the end of the remnant. That's my thinking, anyhow.

Thanks!

StevenGreen1:
Hi John,

Thanks for that info.  Ah so it seems like it'd be nice to have some iterative vertexing that builds the 3D particles and then uses that info to further refine things as that would clear this issue up (hopefully).  However, that's too big a change to the flow to introduce fro now and this change alone works well.  

In terms of plots, I think I'd go for a vertex resolution plot as that's ultimately what is being modified.  I'd also check a reco efficiency/correct event fraction too for completeness.  I can't imagine this change would impact either in a significant way, but given we'll be making both the vertexing and reco eff plots from the event validation metrics, we may as well check both. Does that sound ok to you?  I'll get the metrics going for ProtoDUNE later today.

Cheers
Steve

StevenGreen1:
Also before I get on with the testing, could you rebase this on top of the current master as there have been a few changes to the master since this feature was made (2 of the 5 commits are already in the master history!).

Cheers
Steve

PandoraPFA:
Hi Steve,

Thanks for looking at ProtoDUNE metrics. I'll try to get MicroBooNE metrics during the next couple of days. Thanks for the heads up regarding rebasing too. Will do whatever is needed at the point of merging to make this sit cleanly upon the existing commits, with N commits to a feature branch and a single merge commit (no-ff) back to master, as usual.

Thanks!

StevenGreen1:
Hi John,

No worries and sounds good. Could I be a pain and as you to do the rebasing soon (in the next few days is fine) and the force push to your feature branch? I like to, when validating PRs, to compare the feature branch about to be merged to the current master.  If I do that now with this branch, I have to account for the changes that are in the master, but not in your feature. I think if I rebased this myself for testing it would probably be fine, but it does mean the testing I do relies on the assumption that the rebasing I do for testing will be identical to the rebasing you do for merging and by rebasing now we can remove that assumption (this is very picky and I'll work around it if you want, but it's just how I've been testing all my PRs).

Cheers
Steve

PandoraPFA:
Done (rebasing and sorting of 3D cluster list), thanks! Next step, unless you can think of anything, is to collect some performance figures. Will get on to that at earliest opportunity, but might be a day or two.

PandoraPFA:
Hi Steve,

I've checked through the MicroBooNE metrics and there are essentially no changes. I did find one event where there was a change, from outputting a particle with a displaced vertex to outputting no particles. This is because the event is rather "longitudinal" and one view is missing in a gap.

The vertex is placed at the beginning of the track, but pfos are only formed from hits further down the track. The failure is because there are precisely no 3D hits produced in this event, where it's longitudinal and there's only two views. There's no way to move the vertex after finding there's no vertex-associated Pfos.

We could make a small change to just return (and do what was done before) rather than raise an exception (and bail) if there are no 3D hits. What do you think?

Here's the picture of the relevant event; hopefully it's clear what's going on. Red (u), cyan (v) and purple (w) are hits in pfos; grey are all input hits and the yellow marker is the vertex.

Screenshot 2019-10-15 at 21 35 23

Thanks!

StevenGreen1:
Hi John,

Some of the ProtoDUNE metrics are affected by this change sadly.  The vertex resolution plot shows a non-negligible difference, which I will begin looking into now.

The metrics for the individual particles in the hierearchy are unchanged though so it's just the final vertex that is being affect by this change, as we expect. 
 
**Vertex Resolution**
VertexDeltaR

**Reco Eff, Purity, Comp**
PrimaryCompleteness
PrimaryPurity
PrimaryRecoEfficiencyVsNHits

PandoraPFA:
Hi Steve,

Thanks for this info - are these tests including the most recent commit?
https://github.com/PandoraPFA/LArContent/pull/99/commits/5f767acc48b5fe91e9e525f517c25300b5bbe12c

If not, then we'd expect to lose any particles where the vertex had no associated pfos (wasn't clear whether that should ever happen at ProtoDUNE).

If so, then must mean that moving the vertex to the nearest 3D hit (if there are no vertex-associated pfos) has a noticeable impact. The question then is whether it's making things better or worse.

If all is working right in the implementation, it's difficult to see that we'd ever want to leave a vertex in a position where it isn't associated to any Pfos. So, things might actually be better, OR we might not be moving the vertex to the best 3D position?

Thanks again!

StevenGreen1:
Hi John,

So this testing doesn't include that, but I'll start some new ones going now with it in to double check.  In the mean time I'll also try and pull up a few events to see what's going on as hopefully it'll be clear.

Cheers
Steve

StevenGreen1:
Hi John,

Having scanned a few more ProtoDUNE events, the topology with a vertex far from ay 3D hits is actually semi-common and so the vertex is shifted as expected.  I think the PR changes does still make sense though for these events and the new vertex is more sensible than one floating far away from the interaction.  Therefore, I'm happy to say we should merge this.

Cheers
Steve

A few ProtoDUNE examples: 
Screenshot 2019-10-17 at 16 07 32
Screenshot 2019-10-17 at 16 17 50

PandoraPFA:
Hi Steve,

Thank you very much for the careful testing in the ProtoDUNE environment - much appreciated!

PR #101, opened 2019-11-25
Hook up save event API.

CrossR:
This just hooks in the `PandoraMonitoring` saving API.

For now, I've condensed it into a single option, if the user has displaying set (`DisplayEvent`) and a path to save to set (`SaveEventPath`), we save the images.

Of note, this option only applies to the `LArVisualMonitoring` the option is applied to. That means to save both pre and post reco displays, the `SaveEventPath` option must be set in both parts of the XML. This is both the easiest way to add it, and also gives some flexibility on only saving one or the other, to save space etc.


PandoraPFA:
Hi Ryan, Thanks - this looks good, but I think we should sit on it until we next release a new "pandora" external product to LArSoft. Only then would we want to couple the algorithm library to a newer version of the pandora external. I've been talking about making some small changes to the SDK for release early in the New Year, so the Monitoring update could also go out then, as could changes to LArContent that rely on the updated monitoring. I hope that works for you.

CrossR:
Tidied this up, I think the force pushes to master back in Dec/Jan had caused this to become a bit strange and include some unrelated changes.

Tested again and still works, and I've been using it locally for a while to speed up hand scanning.

PR #104, opened 2020-01-18
Deep Learning Vertexing DUNE

phuncr:


larpandoracontent/LArVertex/DLVertexCreationAlgorithm.cc
```diff
@@ -374,6 +378,62 @@ bool DLVertexCreationAlgorithm::DetectorCheck(const pandora::CartesianVector &po
     return((!volumeCheck) || gapCheck);
 }
 
+//------------------------------------------------------------------------------------------------------------------------------------------
+bool DLVertexCreationAlgorithm::EventViewCheck(const pandora::ClusterList *const pClusterList, unsigned int &allHitsCount) const
```
AndyChappell: Minor point, but I would intuitively expect a function called `EventViewCheck` to return true if there were hits to be processed, rather than if there was nothing there. I would consider either changing the name to something like `IsEventViewEmpty`, or perhaps inverting the return value.

AndyChappell:
Archived.

PR #107, opened 2020-03-11
Feature/TwoViewMatching

PandoraPFA:
Some restructuring of the 2D->3D matching algorithms in order to (ideally at minimal cost) support two-view matching approaches. These would be either for dual-phase LArTPCs, or for two-view "particle recovery" in a single-phase LArTPC.

The ThreeDBase algorithm has been replaced by a MatchingBase algorithm, from which TwoViewMatching and ThreeViewMatching algorithms inherit. The three-view case owns an OverlapTensor<T> (where T is some kind of OverlapResult) and the two-view case own an OverlapMatrix<T>.

One area of repetition is the TwoViewTrackMatching and ThreeViewTrackMatching algorithms. Clever ideas to avoid/reduce repetition would be most welcome here.

The existing (and rather complex) three-view matching algorithms should be rather unaffected, with just a few interface-related changes required in order to aid consistency. Their config is, currently, deliberately unaffected, though this means registering the algorithms under their old names, and not yet reflecting the fact that the class names have changed for clarity.

An example TwoViewTransverseTracks algorithm has been added, with a placeholder OverlapResult, and a placeholder TwoViewClearTracks tool, alongside a TransverseMatrixVisualisation tool (which is probably the most useful starting point).

Thanks!


PandoraPFA:
Will look into Travis CI failure - there have been a number of build failures this week, with some relating to docker. May need to remove commit 5035fb9 from the mix (review at point of merging).

larpandoracontent/LArThreeDReco/LArThreeDBase/ThreeViewTrackMatchingAlgorithm.h
```diff
@@ -1,42 +1,36 @@
...
-
-//------------------------------------------------------------------------------------------------------------------------------------------
-
 /**
- *  @brief  ThreeDTransverseTracksAlgorithm class
+ *  @brief  ThreeViewTrackMatchingAlgorithm class
  */
 template<typename T>
-class ThreeDTracksBaseAlgorithm : public ThreeDBaseAlgorithm<T>
+class ThreeViewTrackMatchingAlgorithm : public ThreeViewMatchingAlgorithm<T>
```
AndyChappell: I think the matching classes can be reduced to a single instance of TrackMatchingAlgorithm<T, U> -> MatchingAlgorithm<T, U> -> MatchingBaseAlgorithm with the addition of the extra type in the template specification. I think a new abstract class, ClusterOverlapContainer (or some better name), could take ownership of the cluster lists, input cluster lists and the overlap tensor/matrix, along with the functions to perform the UpdateForNewCluster, UpdateUponDeletion, GetClusterListName etc. which can handle the specific tasks relating to 2 or 3 views.

Most of the other functions then become essentially identical for the 2 or 3 view case, so you wouldn't need Two and Three view classes for matching, just a single hierarchy where all of the differences relating to the number of views get moved to this new mini class hierarchy.

PR #108, opened 2020-03-11
Feature/TwoDMetricsExample

PandoraPFA:
Added the ability to set all of the parameters used in the MCParticle primary selection via XML in the EventValidationBase algorithm.

This was motivated by a workflow where the existing validation mechanics are used to assess 2D pattern recognition (so need to adjust the number of "good views" used to select an MC primary), but the change stands alone.

[The commit 5035fb9 is present because it's in the LArSoft/larpandoracontent/develop branch (to which we will ultimately submit any PR), but isn't yet in the LArSoft master or PandoraPFA master branches. Its inclusion will be considered/updated before PR is submitted.]


PandoraPFA:
Merged.

PR #110, opened 2020-03-16
Feature/unfolding mc particle hierarchy

imawby:
Some Comments:

1) I had removed the photon and neutron considerations remain but  can easily be switched off by setting m_selectInputHits to false in the xml. 
 2) As per the current implementation the user has to do some of the unfolding or folding on the pfo side themselves. This is because they have to pass the function a list of pfos and they get a pfo->hits map for those pfos. So for the unfolded version they would have to pass in the whole pfo list and for the folded version they would have to pass in the list of primaries. I thought this was a bit annoying but I didn't change it since I think in the case of the test beam the pfos you want to select isn't as simple as those two cases, also one could chose to look at cosmics only etc...
3) I have pretty much done the same thing for the test beam as the neutrino case since the functions have a very similar structure. The only difference is that there is some code in the test beam case that i think says  'if the parent particle of the any of the MCParticles in the mc->hits map is not itself in the map (because it wasn't deemed reconstructable) then add it in but with an empty calo hit list associated to it' I thought this was independent of the folding so left it in.

### DIFF
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -34,7 +34,7 @@ LArMCParticleHelper::PrimaryParameters::PrimaryParameters() :
     m_minPrimaryGoodViews(2),
     m_selectInputHits(true),
     m_maxPhotonPropagation(2.5f),
-    m_minHitSharingFraction(0.9f), 
+    m_minHitSharingFraction(0.9f),
     m_foldBackHierarchy(true)
 {
 }
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -34,11 +34,42 @@ LArMCParticleHelper::PrimaryParameters::PrimaryParameters() :
     m_minPrimaryGoodViews(2),
     m_selectInputHits(true),
     m_maxPhotonPropagation(2.5f),
-    m_minHitSharingFraction(0.9f)
+    m_minHitSharingFraction(0.9f), 
+    m_foldBackHierarchy(true)
 {
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool LArMCParticleHelper::DoesPrimaryMeetCriteria(const MCParticle *const pMCParticle, std::function<bool(const MCParticle *const)> fCriteria)
+{
+    try
+    {
+        const MCParticle *const pPrimaryMCParticle = LArMCParticleHelper::GetPrimaryMCParticle(pMCParticle);
+        return fCriteria(pPrimaryMCParticle);
+    }
+    catch (const StatusCodeException &) {}
+
+    return false;
+    
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool LArMCParticleHelper::DoesLeadingMeetCriteria(const MCParticle *const pMCParticle, std::function<bool(const MCParticle *const)> fCriteria)
+{
+    try
+    {
+        const MCParticle *const pLeadingMCParticle = LArMCParticleHelper::GetLeadingMCParticle(pMCParticle);
+        return fCriteria(pLeadingMCParticle);
+    }
+    catch (const StatusCodeException &) {}
+
+    return false;
+
+}
+
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 bool LArMCParticleHelper::IsBeamNeutrinoFinalState(const MCParticle *const pMCParticle)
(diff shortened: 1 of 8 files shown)


larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -810,7 +865,7 @@ bool LArMCParticleHelper::PassMCParticleChecks(const MCParticle *const pOriginal
     if (NEUTRON == std::abs(pThisMCParticle->GetParticleId()))
         return false;
 
-    if ((PHOTON == pThisMCParticle->GetParticleId()) && (PHOTON != pOriginalPrimary->GetParticleId()) && (E_MINUS != std::abs(pOriginalPrimary->GetParticleId())))
+    if ((PHOTON == pThisMCParticle->GetParticleId()) && (PHOTON != GetPrimaryMCParticle(pThisMCParticle)->GetParticleId()) && (E_MINUS != std::abs(GetPrimaryMCParticle(pThisMCParticle)->GetParticleId())))
```
AndyChappell: I would take the opportunity to assign the result of `GetPrimaryMCParticle(pThisMCParticle)->GetParticleId()` to a variable before its use in the conditional, otherwise you're repeating exactly the same function call twice - assuming the compiler can't optimise this away for you (it probably can't).

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -34,13 +34,44 @@ LArMCParticleHelper::PrimaryParameters::PrimaryParameters() :
...
+    m_foldBackHierarchy(true)
 {
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+bool LArMCParticleHelper::DoesPrimaryMeetCriteria(const MCParticle *const pMCParticle, std::function<bool(const MCParticle *const)> fCriteria)
+{
+    try
```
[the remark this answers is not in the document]
imawby: When testing I found that an error occurs when trying to find the primary MCParticle - from memory I think it was a STATUS_CODE_NOT_FOUND but I could be wrong. When using the helper to create a MCParticle -> hits map you have to provide some criteria that you want the particles to meet, these criteria are things like: 'is it a cosmic ray AND is it a primary'. Obviously when not folding back the hierarchy you want to avoid the latter criteria so for each particle I obtain the primary and apply the criteria to the primary. This means that the particle has to have a primary, so if it doesn't then i just say it doesn't meet the criteria. So there is a hidden criteria that the MCParticle has to have a primary, which I think is something I want to have.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
imawby: so, i have changed this back to pOriginalPrimary and just made sure that pOriginalPrimary is obtained in the calling function by the function GetPrimaryMCParticle instead of from the map (which is now not necessarily) the MCParticle -> primary MCParticle map

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -479,7 +479,14 @@ void LArMCParticleHelper::SelectReconstructableMCParticles(const MCParticleList
...
     // Obtain vector: target mc particles
     MCParticleVector targetMCVector;
-    parameters.m_foldBackHierarchy ? LArMCParticleHelper::GetPrimaryMCParticleList(pMCParticleList, targetMCVector) : (void) std::copy(pMCParticleList->begin(), pMCParticleList->end(), std::back_inserter(targetMCVector));
+    if (parameters.m_foldBackHierarchy)
+    {
+        LArMCParticleHelper::GetPrimaryMCParticleList(pMCParticleList, targetMCVector);
+    }
+    else
+    {
+        (void) std::copy(pMCParticleList->begin(), pMCParticleList->end(), std::back_inserter(targetMCVector));
```
AndyChappell: No need for the preceding (void) here

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -510,7 +517,14 @@ void LArMCParticleHelper::SelectReconstructableTestBeamHierarchyMCParticles(cons
...
     // Obtain vector: target mc particles
     MCParticleVector targetMCVector;
-    parameters.m_foldBackHierarchy ? LArMCParticleHelper::GetLeadingMCParticleList(pMCParticleList, targetMCVector) : (void) std::copy(pMCParticleList->begin(), pMCParticleList->end(), std::back_inserter(targetMCVector));
+    if (parameters.m_foldBackHierarchy)
+    {
+        LArMCParticleHelper::GetLeadingMCParticleList(pMCParticleList, targetMCVector);
+    }
+    else
+    {
+        (void) std::copy(pMCParticleList->begin(), pMCParticleList->end(), std::back_inserter(targetMCVector));
```
AndyChappell: No need for the preceding (void).

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -648,7 +662,14 @@ void LArMCParticleHelper::CollectReconstructable2DHits(const ParticleFlowObject
 
     // If foldBackHierarchy collect all 2D calo hits in pfo hierarchy
     // else collect hits directly belonging to pfo
-    foldBackHierarchy ? LArPfoHelper::GetAllDownstreamPfos(pPfo, pfoList) : pfoList.push_back(pPfo);
+    if (foldBackHierarchy)
+    {
+	LArPfoHelper::GetAllDownstreamPfos(pPfo, pfoList);
```
AndyChappell: Indentation

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -648,7 +662,14 @@ void LArMCParticleHelper::CollectReconstructable2DHits(const ParticleFlowObject
...
+    }
+    else
+    {
+	pfoList.push_back(pPfo);
```
AndyChappell: Indentation

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -813,8 +834,26 @@ void LArMCParticleHelper::SelectParticlesMatchingCriteria(const MCParticleVector
...
+		    continue;
+	    }
+	    else
+	    {
+	        if (!LArMCParticleHelper::DoesPrimaryMeetCriteria(pMCParticle, fCriteria))
+		    continue;
+	    }
+	}
+	
+        selectedParticles.push_back(pMCParticle);
```
AndyChappell: Indentation

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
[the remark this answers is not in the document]
PandoraPFA: Turns out that LArMCParticleHelper::GetPrimaryMCParticle can raise an exception if there's a failure to find said primary, or if the input MC info is in an unexpected form, so that MCParticles can have more than one parent. We're at the mercy of the input MC information, so there are some sanity checks, and it's via testing that these will be picked up.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
PandoraPFA: I wondered about the details behind this change - so that's back as per the original implementation, now?

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
imawby: (this has been fixed in the 'indentation fix' commit)

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
imawby: yes, it i have put back to the original

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -598,31 +653,50 @@ void LArMCParticleHelper::GetPfoMCParticleHitSharingMaps(const PfoContributionMa
...
+    // If foldBackHierarchy collect all 2D calo hits in pfo hierarchy
+    // else collect hits directly belonging to pfo
 
-    // ATTN: Only collect downstream pfos for daughter test beam particles & cosmics
-    if (pPfo->GetParentPfoList().empty() && LArPfoHelper::IsTestBeam(pPfo))
+    if (foldBackHierarchy)
     {
-        pfoList.push_back(pPfo);
+        // ATTN: Only collect downstream pfos for daughter test beam particles & cosmics
+        if (pPfo->GetParentPfoList().empty() && LArPfoHelper::IsTestBeam(pPfo))
```
PandoraPFA: For info: This is fiddly logic (existed before this PR), basically trying to handle the incoming test beam particle in a similar manner to a neutrino case. i.e. it's the chains of particles emerging from the interaction vertex that we want to collect. If we collect starting from the test beam particle itself, we'll only ever get one hierarchy.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -681,12 +755,12 @@ void LArMCParticleHelper::SelectCaloHits(const CaloHitList *const pCaloHitList,
...
 
-            LArMCParticleHelper::MCRelationMap::const_iterator mcIter = mcToPrimaryMCMap.find(pHitParticle);
+            LArMCParticleHelper::MCRelationMap::const_iterator mcIter = mcToTargetMCMap.find(pHitParticle);
 
-            if (mcToPrimaryMCMap.end() == mcIter)
+            if (mcToTargetMCMap.end() == mcIter)
                 continue;
 
-            const MCParticle *const pPrimaryParticle = mcIter->second;
+            const MCParticle *const pPrimaryParticle = LArMCParticleHelper::GetPrimaryMCParticle(pHitParticle);
```
PandoraPFA: Can you explain the reasoning for this change, thanks? If subtle, could add an // ATTN comment.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -754,19 +828,36 @@ void LArMCParticleHelper::SelectGoodCaloHits(const CaloHitList *const pSelectedC
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 void LArMCParticleHelper::SelectParticlesMatchingCriteria(const MCParticleVector &inputMCParticles, std::function<bool(const MCParticle *const)> fCriteria,
-    MCParticleVector &selectedParticles)
+    MCParticleVector &selectedParticles, const bool foldBackHierarchy, bool isTestBeam)
```
PandoraPFA: `const bool isTestBeam`

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -754,19 +828,36 @@ void LArMCParticleHelper::SelectGoodCaloHits(const CaloHitList *const pSelectedC
...
-        if (fCriteria(pMCParticle))
-            selectedParticles.push_back(pMCParticle);
+        if (foldBackHierarchy)
+        {
+            if (!fCriteria(pMCParticle))
+                continue;
+        }
+        else
+        {
+            if(isTestBeam)
```
PandoraPFA: Typically have space after if, e.g. `if (isTestBeam)`

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -754,19 +828,36 @@ void LArMCParticleHelper::SelectGoodCaloHits(const CaloHitList *const pSelectedC
...
+            else
+            {
+                if (!LArMCParticleHelper::DoesPrimaryMeetCriteria(pMCParticle, fCriteria))
+                    continue;
+            }
+        }
+        selectedParticles.push_back(pMCParticle);
     }
-}
+}
```
PandoraPFA: Stray whitespace!

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -64,8 +64,25 @@ class LArMCParticleHelper
         bool          m_selectInputHits;          ///< whether to select input hits
         float         m_maxPhotonPropagation;     ///< the maximum photon propagation length
         float         m_minHitSharingFraction;    ///< the minimum Hit sharing fraction
+	bool          m_foldBackHierarchy;        ///< whether to fold the hierarchy back to the primary (neutrino) or leading particles (test beam)
```
PandoraPFA: Indentation.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -274,9 +299,10 @@ class LArMCParticleHelper
      *  @param  pfoList the input list of Pfos
      *  @param  selectedMCParticleToHitsMap the input mapping from selected reconstructable MCParticles to their hits
      *  @param  pfoToReconstructable2DHitsMap the output mapping from Pfos to their reconstructable 2D hits
+     *  @param  foldToPrimaries whether to fold the particle hierarchy back to primaries
      */
     static void GetPfoToReconstructable2DHitsMap(const pandora::PfoList &pfoList, const MCContributionMap &selectedMCParticleToHitsMap,
-        PfoContributionMap &pfoToReconstructable2DHitsMap);
+	PfoContributionMap &pfoToReconstructable2DHitsMap, bool foldBackHierachy);
```
PandoraPFA: Check indentation (and const bool as discussed in implementation file). Update variable name in doxygen @param above, and in other cases here.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
imawby: In the folding case, for which the code was originally written, the function would be passed the map particle->primary particle. Then the code checks if the particle chain (from primary to particle) passes some checks - if there is a photon that travels beyond a certain distance or a neutron in the chain it does not pass the checks. When we are unfolding we can have the case where this function is instead passed the map particle->particle and so pPrimaryParticle may not be equal to mcIter->second. Therefore, I changed this line to get the PrimaryParticle from a function instead of the map. For half the cases it is a repetition as this has already been done when the user choses to fold back, so maybe I could add a statement that checks whether mcIter->first == mcIter->second and only run this function when this is true? and use mcIter->second when this is not true?

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
imawby: sorry! still getting used to the Pandora style

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -422,59 +461,75 @@ void LArMCParticleHelper::GetMCParticleToCaloHitMatches(const CaloHitList *const
...
     // Ensure the MCParticles have enough "good" hits to be reconstructed
-    LArMCParticleHelper::SelectParticlesByHitCount(candidateTargets, mcToTrueHitListMap, mcToPrimaryMCMap, parameters, selectedMCParticlesToHitsMap);
+    LArMCParticleHelper::SelectParticlesByHitCount(candidateTargets, targetMCToTrueHitListMap, mcToTargetMCMap, parameters, selectedMCParticlesToHitsMap);
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 void LArMCParticleHelper::SelectReconstructableTestBeamHierarchyMCParticles(const MCParticleList *pMCParticleList, const CaloHitList *pCaloHitList, const PrimaryParameters &parameters,
     std::function<bool(const MCParticle *const)> fCriteria, MCContributionMap &selectedMCParticlesToHitsMap)
 {
```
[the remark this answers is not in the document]
imawby: maybe in the future we can add another member variable to the PrimaryParameters and combine the function 'branches' into one?

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
imawby: thank you for the info

PandoraPFA:
Merged - thanks!

PR #112, opened 2020-03-27
Feature/TwoViewMatchingUpdate

PandoraPFA:
Refinements to the TwoViewMatching, leading to more complete sharing of implementation between two-view and three-view cases. The differences between these cases are handled via the Two and ThreeViewMatchingControl classes, which (a little redundantly, but helping to enforce some consistency) inherit an interface defined by the NViewMatchingControl class.

Inheritance structure now is: 
Screenshot 2020-03-27 at 11 34 49


PandoraPFA:
Merged - thanks @AndyChappell

PR #122, opened 2020-05-27
feature/TrackInEMShower

imawby:
The TrackInEMShower algorithm is designed to merge CR muons that have been split by EM behaviour along the the track. The algorithm finds associated clusters, collects hits that fall close to a line which connects the two determined 'merging points' and merges them together. 

There is a remaining problem, which i'm not sure how to handle. Sometimes the clustering follows a delta ray and this is a mistake, but this algorithm attempts to find a 'merging point' by walking back along the cluster from it's endpoint until it finds a point that has a gradient which agrees with the average gradient of the associated cluster. So it would find a point before the clustering has followed the delta ray. It would then find all hits close to the connecting line and any hits that are in the cluster and between the two merging points but do not fall close to the line (i.e. the ones of the delta ray) are removed from the cluster. The problem is that I then do not know what to do with them and I just abandon them. I didn't want to make a single cluster out of them as this wouldn't make sense (we could be merging clusters on opposite sides of the track). Maybe we could have a proximity based clustering. i.e. these are our new clusters is this hit close to any of them, if not make a new cluster. I wondered your thoughts? (I have attached an image of a delta ray problem, as you can see the hits are lost by the end)

BEFORE ALG: The red and blue clusters are those that are associated and the dots their merging points, you can see the merging point is behind the delta ray for the red cluster
ProtoDUNE_Before_2
AFTER ALG: The clusters get merged, and the original green merging mistake is also fixed (yay!) but you see that we have lost the hits from the delta ray
ProtoDUNE_After_2


larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+        if (caloHitIter == extrapolatedCaloHitVector.end())
+        {
+            ++segmentsWithoutHits;
+            if (segmentsWithoutHits > m_maxTrackGaps)
+                return false;
+            continue;
+        }
+        
+        CartesianVector lowerBoundary(innerMergePoint + (trackStep * static_cast<float>(i)));
+        CartesianVector upperBoundary(innerMergePoint + (trackStep * static_cast<float>(i + 1.f)));
```
AndyChappell, line 416: The casts here aren't necessary.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+
+        if (i >= fullSegments - 1)
+        {
+            if (lengthOfTrackRemainder > m_lineSegmentLength * 0.5f)
+            {
+                lowerBoundary = lowerBoundary - (trackStep * 0.5f * (i - fullSegments + 1.f)) + (trackDirection * 0.5f * (i - fullSegments + 1.f) * lengthOfTrackRemainder);
+                upperBoundary = upperBoundary - (trackStep * 0.5f * (i - fullSegments + 2.f)) + (trackDirection * 0.5f * (i - fullSegments + 2.f) * lengthOfTrackRemainder);
```
AndyChappell, line 423: This looks like it is dealing with the final segment, which is not necessarily the length of a track step, but some fraction of a track step. What's the need for all of the half step adjustments here?

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -0,0 +1,353 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+inline pandora::CartesianVector TrackInEMShowerAlgorithm::ClusterAssociation::GetConnectingLineDirection() const
+{
+    return m_connectingLineDirection;
+}
+
+} //namespace lar_content
+
+
```
AndyChappell, line 352: Whitespace

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+#include "larpandoracontent/LArHelpers/LArGeometryHelper.h"
+
+#include "larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+TrackInEMShowerAlgorithm::ClusterAssociation::ClusterAssociation(const Cluster *const pInnerCluster, const Cluster *const pOuterCluster,const CartesianVector &innerMergePoint,
```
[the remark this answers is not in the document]
imawby: So I stole this convention from elsewhere in pandora, an inner cluster is supposed to be a cluster at a lower Z value whilst an outer cluster is one at a higher Z value. Personally, i'm not a fan of the convention either and it took me a while to work out. I do like the upstream/downstream suggestion though and think it is much clearer

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+    m_minDirectionDeviationCosAngle(0.99),
+    m_distanceFromLine(0.35),
+    m_maxTrackGaps(2),
+    m_lineSegmentLength(7.f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+    
+bool TrackInEMShowerAlgorithm::SortByDistanceToLine::operator() (const pandora::CaloHit *const pLhs, const pandora::CaloHit *const pRhs)
```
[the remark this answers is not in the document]
imawby: That is much better, applied.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+    
+    for (const Cluster *const pCluster : clusterVector)
+    {
+        const TwoDSlidingFitResultMap::const_iterator microFitResultIter(microSlidingFitResultMap.find(pCluster));
+        
+        if(microFitResultIter == microSlidingFitResultMap.end())
+        {
+            try
+            {
+                (void) microSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, m_slidingFitWindow, slidingFitPitch)));
```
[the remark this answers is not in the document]
imawby: I see this often when filling sliding fit maps but am happy to change if thats not how its done now

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+TrackInEMShowerAlgorithm::ClusterAssociation::ClusterAssociation() :
+    m_pInnerCluster(nullptr),
+    m_pOuterCluster(nullptr),
+    m_innerMergePoint(CartesianVector(0.f, 0.f, 0.f)),
+    m_innerMergeDirection(CartesianVector(0.f, 0.f, 0.f)),
+    m_outerMergePoint(CartesianVector(0.f, 0.f, 0.f)),
+    m_outerMergeDirection(CartesianVector(0.f, 0.f, 0.f)),
+    m_connectingLineDirection(CartesianVector(0.f, 0.f, 0.f))
+{
+}
```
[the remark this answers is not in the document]
AndyChappell: So I think this one is a little tricky in its current form. The issue here is that you declare/define a ClusterAssociation object ahead of the call to FindBestClusterAssociation, which leads to the default constructor requirement, before this variable is then reassigned to a 'real' value in the function. Ideally you would keep track of the best clusters etc as you go through the loop, and only instantiate the cluster association when there's an actual association to create, but a clean solution here is not trivial due to the fact that it's not guaranteed to find an association at all, which rules out creating the object in the function and returning a const reference. I'll see if I can think of a workable alternative (that also avoids multiple constructor/destructor calls as the best association is updated).

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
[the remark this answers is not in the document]
AndyChappell: I think it's fine to leave it in if it's common for the sliding fit case.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+        return false;
+
+    if ((extrapolatedInnerPoint.GetZ() > outerPoint.GetZ() + m_maxZSeparation) || (extrapolatedInnerPoint.GetZ() < outerPoint.GetZ() - m_maxZSeparation))
+        return false;
+
+    if ((extrapolatedOuterPoint.GetX() > innerPoint.GetX() + m_maxXSeparation) || (extrapolatedOuterPoint.GetX() < innerPoint.GetX() - m_maxXSeparation))
+        return false;
+
+    if ((extrapolatedOuterPoint.GetZ() > innerPoint.GetZ() + m_maxZSeparation) || (extrapolatedOuterPoint.GetZ() < innerPoint.GetZ() - m_maxZSeparation))
+        return false;
```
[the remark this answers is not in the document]
imawby: changed this to a circular separation cut

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
imawby: So, what I do is merge the final two track segments together (i.e. a full segment + track remainder). Then, if the track remainder was originally over half of the full segment length i split the merged segments into two and loop over both of them. If the track remainder was originally under half of the full segment length i keep the merged segments as one. The half step adjustments are so that i get the correct upper and lower limits for these segments

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
imawby: changed

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,624 @@
...
+    TwoDSlidingFitResultMap &macroSlidingFitResultMap) const
+{
+    const float slidingFitPitch(LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+    
+    for (const Cluster *const pCluster : clusterVector)
+    {
+        try
+        {
+            (void) microSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, m_slidingFitWindow, slidingFitPitch)));
+            (void) macroSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, 1000000, slidingFitPitch)));
```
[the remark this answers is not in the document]
imawby: So when I originally done this I found serious performance changes and also found that the number of entries in the map falls, suggesting fits in some cases are no longer made. I wasn't sure if there was a limit on the sliding fit window after which the fit no longer made sense. But I can have another look at this.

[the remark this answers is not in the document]
AndyChappell: I can see why the computational performance would change, as this upper limit appears to be used as a condition for loop termination in the fit, so uint::max implies O(10^9) iteration loops, so that would seem to incentivise the use of the smallest number you can get away with. It's interesting that the fit would have different outcomes for larger values, given that 10^6 seems like a large enough number for any such issue to appear in practice. Certainly worth investigating a sensible number here and assigning it to some named variable/const to make it clear what the value is being used for.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+        CaloHitVector extrapolatedCaloHitVector;
+        this->GetExtrapolatedCaloHits(clusterAssociation, pClusterList, extrapolatedCaloHitVector, clusterToCaloHitListMap);
+        
+        if (extrapolatedCaloHitVector.empty())
+            break;
+        
+        if (!this->IsTrackContinuous(clusterAssociation, extrapolatedCaloHitVector))
+            break;
+        
+        this->RefineTracks(clusterAssociation, microSlidingFitResultMap, macroSlidingFitResultMap, clusterVector, extrapolatedCaloHitVector);
```
johnmarshall80, line 116: Is it possible to call MergeHits first (which internally does all required updates for deletion, etc.), then RefineTrack for the whole track in one go, or does that break something? (RefineTrack could internally call the functions to do required updates for deletion, etc. again.)

larpandoracontent/LArContent.cc
```diff
@@ -146,6 +146,8 @@
 #include "larpandoracontent/LArTwoDReco/LArClusterAssociation/TransverseAssociationAlgorithm.h"
 #include "larpandoracontent/LArTwoDReco/LArClusterAssociation/TransverseExtensionAlgorithm.h"
 
+#include "larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h"
```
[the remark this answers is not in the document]
imawby: Perhaps the LArCosmicRay subdirectory?

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+        
+        this->MergeHits(clusterAssociation, clusterToCaloHitListMap, microSlidingFitResultMap, macroSlidingFitResultMap, clusterVector);
+
+        this->UpdateSlidingFitResultMap(clusterVector, microSlidingFitResultMap, macroSlidingFitResultMap);
+
+        mergeMade = true;
```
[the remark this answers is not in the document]
imawby: This would actually be super beneficial! It would stop merging of tracks in the shower after the main muon track has been merged.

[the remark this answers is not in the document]
imawby: Although perhaps we can safeguard even further by using the main track cluster we had created in the previous loop iteration (if there was one)

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+    } while (mergeMade);
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackInEMShowerAlgorithm::SelectCleanClusters(const ClusterList *pClusterList, ClusterVector &clusterVector) const
```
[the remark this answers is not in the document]
imawby: noted

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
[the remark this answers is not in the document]
imawby: Actually this can already be done! 

So essentially the cluster vector contains the clusters that the algorithm considers when searching for a CR pair. So, in each passing of the merging process in the cleanup stage I remove clusters from the map and cluster vector that are going to be changed (so the main track clusters and all small shower clusters that have hits that belong to the main track). Then when I update for the next iteration I could perhaps only add back in the main track cluster and so in the future only this cluster and others not considered in the previous merge will be considered. But I think i'll test this with metrics

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+    TwoDSlidingFitResultMap &macroSlidingFitResultMap) const
+{
+    const float slidingFitPitch(LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+    const unsigned int macroSlidingFitWindow(1000);
+    
+    for (const Cluster *const pCluster : clusterVector)
+    {
+        try
+        {
+            (void) microSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, m_slidingFitWindow, slidingFitPitch)));
```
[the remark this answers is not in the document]
imawby: This has been on my mind. 

The clusters contained within a cluster association are guaranteed to have an entry in both maps - this is because when forming the cluster association object a cluster pair if skipped if an entry is not found in both maps. Only the fits of the clusters in the cluster association are ever accessed. 

But I think that i'll demand this to be the case here so the map is smaller

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+
+bool TrackInEMShowerAlgorithm::IsTrackContinuous(const ClusterAssociation &clusterAssociation, CaloHitVector &extrapolatedCaloHitVector) const
+{
+    const CartesianVector &upstreamMergePoint(clusterAssociation.GetUpstreamMergePoint()), &downstreamMergePoint(clusterAssociation.GetDownstreamMergePoint());
+    const CartesianVector &trackDirection(clusterAssociation.GetConnectingLineDirection());
+    const CartesianVector trackStep(trackDirection * m_lineSegmentLength);
+
+    // To handle final segment - merge track remainder with preceding segment and if track remainder was more than half trackstep split into two
+    const float trackLength((downstreamMergePoint - upstreamMergePoint).GetMagnitude());
+    const unsigned int fullSegments(floor(trackLength / m_lineSegmentLength));
```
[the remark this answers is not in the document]
imawby: It is not, but i'll add in this check. m_lineSegmentLength is set by the user, and so if they set it to be 0 then this will always fail. If we raised an exception for an invalid parameter then this would be raised for all events, until the user changed the parameter. So is it better to return false here (in which case the alg will do nothing) with a helpful print statement rather than throwing an exception?

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+
+        if (segmentsWithoutHits > m_maxTrackGaps)
+            return false;
+
+        // leave loop early if final merged segment is not split into two
+        if (i == (fullSegments - 1) && !(lengthOfTrackRemainder > m_lineSegmentLength * 0.5))
+            return true;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
imawby: I agree, this loop is quite difficult to follow. I'll have a think about this

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TrackInEMShowerAlgorithm::IsInLineSegment(const CartesianVector &lowerBoundary, const CartesianVector &upperBoundary, const CartesianVector &point) const
+{
+    const float gradient = (-1.0)*(upperBoundary.GetX() - lowerBoundary.GetX()) / (upperBoundary.GetZ() - lowerBoundary.GetZ());
```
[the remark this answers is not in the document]
imawby: It is meant to be the gradient of the line perpendicular to the 'connecting line'. I use it to mark the boundary edge of the segment. I'll make this clearer

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::MergeAndDeleteClusters(*this, pClosestCluster, pClusterToMerge));
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackInEMShowerAlgorithm::UpdateAfterClusterModification(const Cluster *const pCluster, ClusterVector &clusterVector, TwoDSlidingFitResultMap &microSlidingFitResultMap,
+    TwoDSlidingFitResultMap &macroSlidingFitResultMap) const
+{
+    // Remove from SF maps
+    std::vector<TwoDSlidingFitResultMap*> slidingFitResultVector({&microSlidingFitResultMap, &macroSlidingFitResultMap});
```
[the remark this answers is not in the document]
imawby: I can't a typedef to TwoDSlidingFitResultMap pointers, so i'll define one in my header file

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+void TrackInEMShowerAlgorithm::RemoveClusterFromClusterVector(const Cluster *const pCluster, ClusterVector &clusterVector) const
+{
+    ClusterVector::const_iterator clusterToDelete(std::find(clusterVector.begin(), clusterVector.end(), pCluster));
+    if (clusterToDelete != clusterVector.end())
+        clusterVector.erase(clusterToDelete);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackInEMShowerAlgorithm::UpdateSlidingFitResultMap(const ClusterVector &clusterVector, TwoDSlidingFitResultMap &microSlidingFitResultMap,
```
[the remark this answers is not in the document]
imawby: yes!

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,892 @@
...
+{
+    float rL(0.f), rT(0.f);
+    const TwoDSlidingFitResult &microFitResult(microSlidingFitResultMap.at(pCluster));
+    microFitResult.GetLocalPosition(splitPosition, rL, rT);
+
+    const TwoDSlidingFitResult macroFitResult(macroSlidingFitResultMap.at(pCluster));
+    CartesianVector averageDirection(0.f, 0.f, 0.f);
+    macroFitResult.GetGlobalDirection(macroFitResult.GetLayerFitResultMap().begin()->second.GetGradient(), averageDirection);
+
+    const float clusterGradient(averageDirection.GetZ()/averageDirection.GetX());
```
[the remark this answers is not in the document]
imawby: I've changed the code so that if the cluster direction has a zero x component the cluster is marked as vertical. Then in the following the gradient and intercept correspond to a straight line fit in which the gradient corresponds to dx/dz (which is 0 for this case) so instead the hits are split into a left and right cluster

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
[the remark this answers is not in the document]
johnmarshall80: If you add this check in ReadSettings, and print a useful message before returning e.g. STATUS_CODE_INVALID_PARAMETER, then the whole job will actually refuse to start until the parameter is changed. As things stand, there's no other code that uses this member variable, so it would then be guaranteed not to be zero before use in division. But, future developments may change this, so probably good to protect in this function too, just before the division, e.g. raising an exception if negligible difference to zero `(std::fabs(m_lineSegmentLength) < std::numeric_limits<float>::epsilon())`

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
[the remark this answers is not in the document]
johnmarshall80: e.g. `typedef std::vector<TwoDSlidingFitResultMap*> SlidingFitMapVector;` or some better name, balanced between being too long and being too vague!

larpandoracontent/LArContent.cc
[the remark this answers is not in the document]
johnmarshall80: Yes, that does look best.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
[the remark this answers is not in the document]
johnmarshall80: Great, the main aim of changes such as this is to be sure that things will behave in those exceptional cases where things aren't the standard, expected use case. If we think about what might go wrong each step, then hopefully we'll pre-empt most problems and produce something nice and robust.

larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
```diff
@@ -251,11 +251,8 @@ void StitchingCosmicRayMergingTool::CreatePfoMatches(const LArTPC &larTPC1, cons
         return;
     }
 
-    // To skip impact parameters
-    const bool isInGap3D_1(LArGeometryHelper::IsInGap(this->GetPandora(), pointingVertex1.GetPosition(),  TPC_VIEW_W, 0.f));
-    const bool isInGap3D_2(LArGeometryHelper::IsInGap(this->GetPandora(), pointingVertex2.GetPosition(),  TPC_VIEW_W, 0.f));
-
-    if (!isInGap3D_1 || !isInGap3D_2)
+    if (!LArGeometryHelper::IsInGap(this->GetPandora(), pointingVertex1.GetPosition(),  TPC_3D, 0.f) ||
+        !LArGeometryHelper::IsInGap(this->GetPandora(), pointingVertex2.GetPosition(),  TPC_3D, 0.f))
```
johnmarshall80, line 255: (Could consider making gap tolerance configurable, but don't explicitly need to.) Could just drop the explicit 0.f, and the function should default this parameter to 0.f.

larpandoracontent/LArThreeDReco/LArCosmicRay/CosmicRayTrackRecoveryAlgorithm.cc
```diff
@@ -1,722 +0,0 @@
-/**
```
johnmarshall80, line 1: I'm not sure why this file has disappeared with this commit. Was this intended?

larpandoracontent/LArThreeDReco/LArTransverseTrackMatching/TracksCrossingGapsTool.cc
```diff
@@ -214,16 +214,20 @@ bool TracksCrossingGapsTool::PassesGapChecks(ThreeViewTransverseTracksAlgorithm
...
-
-    if ((STATUS_CODE_SUCCESS != statusCodeW) && (!this->IsEndOfCluster(xSample, slidingFitResultW)))
-        return this->CheckXPositionInGap(xSample, slidingFitResultW, slidingFitResultU, slidingFitResultV, gapInW, gapInU, gapInV);
+        if ((STATUS_CODE_SUCCESS != statusCodeV) && (!this->IsEndOfCluster(xSample, slidingFitResultV)))
+            return this->CheckXPositionInGap(xSample, slidingFitResultV, slidingFitResultU, slidingFitResultW, gapInV, gapInU, gapInW);
 
+        if ((STATUS_CODE_SUCCESS != statusCodeW) && (!this->IsEndOfCluster(xSample, slidingFitResultW)))
+            return this->CheckXPositionInGap(xSample, slidingFitResultW, slidingFitResultU, slidingFitResultV, gapInW, gapInU, gapInV);
+    }
+    catch (StatusCodeException &statusCodeException) {}
```
johnmarshall80, line 229: Could make this `const StatusCodeException &statusCodeException`, but not going to be hugely important here (it'd be more for consistency).

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -95,16 +54,19 @@ StatusCode TrackInEMShowerAlgorithm::Run()
     TwoDSlidingFitResultMap macroSlidingFitResultMap;
     this->InitialiseSlidingFitResultMaps(clusterVector, microSlidingFitResultMap, macroSlidingFitResultMap);
 
-    // Find longest length association and conitnue merging until no further merges can be made
+    // ATTN: Find longest length association and conitnue merging until no further merges can be made
```
johnmarshall80, line 57: Typo: continue

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -146,14 +106,16 @@ void TrackInEMShowerAlgorithm::InitialiseSlidingFitResultMaps(const ClusterVecto
...
     const float slidingFitPitch(LArGeometryHelper::GetWireZPitch(this->GetPandora()));
-    const unsigned int macroSlidingFitWindow(1000);
     
     for (const Cluster *const pCluster : clusterVector)
     {
         try
         {
-            (void) microSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, m_slidingFitWindow, slidingFitPitch)));
-            (void) macroSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, macroSlidingFitWindow, slidingFitPitch)));
+            TwoDSlidingFitResult microSlidingFitResult(pCluster, m_slidingFitWindow, slidingFitPitch);
```
johnmarshall80, line 114: const (for each TwoDSlidingFitResult)

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -146,14 +106,16 @@ void TrackInEMShowerAlgorithm::InitialiseSlidingFitResultMaps(const ClusterVecto
...
+            TwoDSlidingFitResult macroSlidingFitResult(pCluster, m_macroSlidingFitWindow, slidingFitPitch);
+
+            (void) microSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, microSlidingFitResult));
+            (void) macroSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, macroSlidingFitResult));
         }
         catch (StatusCodeException &) {}
```
johnmarshall80, line 120: Can, as per a further comment here, make this a const reference (but not a big issue).

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -245,15 +207,17 @@ bool TrackInEMShowerAlgorithm::GetClusterMergingCoordinates(const TwoDSlidingFit
...
+    const int endLayer(isUpstream ? currentMicroFitResult.GetMinLayer() : currentMicroFitResult.GetMaxLayer());
+    const int loopTerminationLayer(endLayer + (isUpstream ? -1 : 1));
     const int step(isUpstream ? -1 : 1);
-    unsigned int goodPositionCount(0);
-    unsigned int gradientStabilityHitWindow(std::ceil(currentMicroFitResult.GetCluster()->GetNCaloHits() * 0.1));
-    
-    for (unsigned int i = startLayer; i != loopTerminationLayer; i += step)
+
+    // ATTN: Search for stable region for which the local layer gradient agrees well with associated cluster global gradient
+    unsigned int gradientStabilityHitWindow(std::ceil(currentMicroFitResult.GetCluster()->GetNCaloHits() * m_stableRegionClusterFraction)); // ISOBEL: SHOULD THIS BE A LAYER SPAN??
```
johnmarshall80, line 217: You have your own TODO remaining here.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -452,19 +413,59 @@ bool TrackInEMShowerAlgorithm::IsInLineSegment(const CartesianVector &lowerBound
...
+
+    // Determine the shower clusters that contain hits which belong to the main track
+    ClusterVector showerClustersToFragment;
+    for (const auto &entry : clusterToCaloHitListMap)
+        showerClustersToFragment.push_back(entry.first);
+
+    std::sort(showerClustersToFragment.begin(), showerClustersToFragment.end(), LArClusterHelper::SortByNHits);
+
+    // Remove hits from inbetween shower
+    if (!showerClustersToFragment.empty())
```
johnmarshall80, line 437: This check not needed, and can just keep the loop below (where nothing will happen if container is empty).

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -516,224 +510,188 @@ const Cluster* TrackInEMShowerAlgorithm::RefineTrack(const Cluster *const pClust
...
-            
-            PandoraContentApi::Cluster::Parameters parameters;
-            parameters.m_caloHitList = *fragmentPair.first;
-            PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::Cluster::Create(*this, parameters, pNewCluster));
-            allCreatedClusters.push_back(pNewCluster);
-            continue;
-        }
+    const Cluster *pAboveCluster(nullptr), *pBelowCluster(nullptr);
+    
+    const float connectingLineGradient(clusterAssociation.GetConnectingLineDirection().GetZ()/clusterAssociation.GetConnectingLineDirection().GetX());
```
johnmarshall80, line 554: Ensure protected against divide by zero. (Spaces around operator too.)

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -516,224 +510,188 @@ const Cluster* TrackInEMShowerAlgorithm::RefineTrack(const Cluster *const pClust
...
+    for (const OrderedCaloHitList::value_type &mapEntry : orderedCaloHitList)
     {
-        this->UpdateAfterClusterModification(pClusterToEnlarge, clusterVector, microSlidingFitResultMap, macroSlidingFitResultMap);
-        return;
-    }
+        for (const CaloHit *const pCaloHit : *mapEntry.second) 
+        {
+            ++count;
+            
+            if (count == 1)
```
johnmarshall80, line 620: This confused me at first, but it's just to avoid comparing hit with itself the first time round. Is it clearer to keep track of `pPreviousCaloHit`, rather than `previousHitPosition`, then can always just skip over cases where `pPreviousCaloHit == pCaloHit`?

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -516,224 +510,188 @@ const Cluster* TrackInEMShowerAlgorithm::RefineTrack(const Cluster *const pClust
...
+            if (separationDistance > maxSeparationDistance)
+                maxSeparationDistance = separationDistance;
+
+            previousHitPosition = pCaloHit->GetPositionVector();
+        }
     }
-        
-    this->UpdateAfterClusterModification(pClusterToEnlarge, clusterVector, microSlidingFitResultMap, macroSlidingFitResultMap);
+
+    if (maxSeparationDistance > m_maxHitSeparationForConnectedCluster)
```
johnmarshall80, line 632: Could presumably bug out of the loop above, and stop making comparisons (i.e. return true from within loop), if any separation distance crosses this threshold?

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -761,88 +719,88 @@ void TrackInEMShowerAlgorithm::AddToNearestCluster(const Cluster *const pCluster
...
-    // Remove from SF maps
-    std::vector<TwoDSlidingFitResultMap*> slidingFitResultVector({&microSlidingFitResultMap, &macroSlidingFitResultMap});
-    this->RemoveClusterFromSlidingFitResultMaps(pCluster, slidingFitResultVector);
-
-    // Remove from CV
-    this->RemoveClusterFromClusterVector(pCluster, clusterVector);
-}
-
-//------------------------------------------------------------------------------------------------------------------------------------------
+    std::vector<TwoDSlidingFitResultMap*> slidingFitResultMapVector({&microSlidingFitResultMap, &macroSlidingFitResultMap});
```
johnmarshall80, line 725: Could typedef this, for clarity.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
```diff
@@ -761,88 +719,88 @@ void TrackInEMShowerAlgorithm::AddToNearestCluster(const Cluster *const pCluster
...
-        
-        if(microFitResultIter == microSlidingFitResultMap.end())
-        {
-            try
-            {
-                (void) microSlidingFitResultMap.insert(TwoDSlidingFitResultMap::value_type(pCluster, TwoDSlidingFitResult(pCluster, m_slidingFitWindow, slidingFitPitch)));
-            }
-            catch (StatusCodeException &) {}
-        }
+        TwoDSlidingFitResult microSlidingFitResult(pMainTrackCluster, m_slidingFitWindow, slidingFitPitch);
```
johnmarshall80, line 750: const

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -48,51 +48,41 @@ class TrackInEMShowerAlgorithm : public pandora::Algorithm
...
          *  @brief  Returns the downstream cluster address
          */
-        const pandora::Cluster* GetDownstreamCluster() const;
+        const pandora::Cluster *GetDownstreamCluster() const;
 
         /**
          *  @brief  Returns the upstream cluster merge point
          */
-        pandora::CartesianVector GetUpstreamMergePoint() const;
+        const pandora::CartesianVector GetUpstreamMergePoint() const;
```
johnmarshall80, line 61: Return const reference, for all cases where the instance actually owns the CartesianVector in question. No problem with giving a const reference to the member away, and is more efficient than copying the object.`const pandora::CartesianVector &GetUpstreamMergePoint() const;`

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -121,17 +108,16 @@ class TrackInEMShowerAlgorithm : public pandora::Algorithm
          *  @param  startPoint the line start point
          *  @param  lineDirection the line direction unit vector
          */
-	    SortByDistanceAlongLine(const pandora::CartesianVector &startPoint, const pandora::CartesianVector &lineDirection) :
-            m_startPoint(startPoint), m_lineDirection(lineDirection.GetUnitVector()) {}
+	    SortByDistanceAlongLine(const pandora::CartesianVector &startPoint, const pandora::CartesianVector &lineDirection);
```
johnmarshall80, line 111: Check indentation. Not a tab character? Can run the removeWhiteSpace.sh script too, to finish up.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -1,12 +1,12 @@
...
+ *  @brief  Header file for the track in em shower algorithm class
  *
  *  $Log: $
  */
-#ifndef LAR_EM_TRACK_ALGORITHM_H
-#define LAR_EM_TRACK_ALGORITHM_H 1
+#ifndef LAR_TRACK_IN_EM_SHOWER_ALGORITHM_H
+#define LAR_TRACK_IN_EM_SHOWER_ALGORITHM_H 1
 
 #include "Pandora/Algorithm.h"
```
johnmarshall80, line 12: You probably need <unordered_map> here now too, below the Pandora and larpandoracontent includes.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -274,8 +232,8 @@ class TrackInEMShowerAlgorithm : public pandora::Algorithm
      *  @param  fragmentAbove whether to fragment the above hits based on their separation
      *  @param  fragmentBelow whether to fragment the below hits based on their separation
      */
-    void FragmentCluster(const pandora::Cluster *const pCluster, const pandora::Cluster *const pClusterToEnlarge, const pandora::CaloHitList &aboveTrackList,
-        const pandora::CaloHitList &belowTrackList, const bool fragmentAbove, const bool fragmentBelow) const;
+    //void FragmentCluster(const pandora::Cluster *const pCluster, const pandora::Cluster *const pClusterToEnlarge, const pandora::CaloHitList &aboveTrackList,
```
johnmarshall80, line 235: Can presumably remove all these commented-out definitions now?

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -307,37 +265,31 @@ class TrackInEMShowerAlgorithm : public pandora::Algorithm
      *
      *  @return bool whether the track is disconnected
      */    
-    bool IsClusterRemnantDisconnected(const pandora::CaloHitList &trackHits) const;
+    bool IsClusterRemnantDisconnected(const pandora::Cluster *const pRemnantCluster) const;
+
+    void FragmentRemnantCluster(const pandora::Cluster *const pRemnantCluster) const;
```
johnmarshall80, line 270: Doxygen comments to be added, here and for other functions below.

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.h
```diff
@@ -348,100 +300,81 @@ class TrackInEMShowerAlgorithm : public pandora::Algorithm
...
-inline void TrackInEMShowerAlgorithm::ClusterAssociation::SetUpstreamCluster(const pandora::Cluster *const pCluster)
-{
-    m_pUpstreamCluster = pCluster;
-}
-
-//------------------------------------------------------------------------------------------------------------------------------------------
-
-inline void TrackInEMShowerAlgorithm::ClusterAssociation::SetDownstreamCluster(const pandora::Cluster *const pCluster)
+inline TrackInEMShowerAlgorithm::SortByDistanceAlongLine::SortByDistanceAlongLine(const pandora::CartesianVector &startPoint, const pandora::CartesianVector &lineDirection) :
+    m_startPoint(startPoint), m_lineDirection(lineDirection.GetUnitVector())
```
johnmarshall80, line 376: Initialisations: one per line.

larpandoracontent/LArThreeDReco/LArCosmicRay/CosmicRayTrackRecoveryAlgorithm.cc
imawby: Yes, I added some visualisations to fo some debugging and I think I must have added the file at some point with whitespace changes

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
imawby: much clearer, thank you!!

larpandoracontent/LArTwoDReco/TrackInEMShowerAlgorithm.cc
imawby: thanks!

larpandoracontent/LArThreeDReco/LArCosmicRay/CosmicRayTrackRecoveryAlgorithm.cc
imawby: I had added some code for debugging some failure and accidentally added it to the PR, i think when trying to remove it i accidentally deleted it. But now adding it back in!

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.h
```diff
@@ -0,0 +1,427 @@
...
+    class SortByDistanceAlongLine
+    {
+    public:
+        /**
+         *  @brief  Constructor
+         *
+         *  @param  startPoint the line start point
+         *  @param  lineDirection the line direction unit vector
+         */
+	    SortByDistanceAlongLine(const pandora::CartesianVector &startPoint, const pandora::CartesianVector &lineDirection);
```
PandoraPFA, line 129: Indentation, and check all changed files with removeWhiteSpace.sh

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+
+#include "larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+TrackInEMShowerAlgorithm::TrackInEMShowerAlgorithm() :
+    m_maxMainLoopIterations(5),
```
PandoraPFA, line 21: Is this number large enough to handle standard usage at e.g. ProtoDUNE?

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+
+bool TrackInEMShowerAlgorithm::IsTrackContinuous(const ClusterAssociation &clusterAssociation, const ClusterToCaloHitListMap &clusterToCaloHitListMap) const
+{
+    // ATTN: Collect extrapolated calo hits and sort by projected distance from the upstreamMergePoint
+    CaloHitVector extrapolatedCaloHitVector;
+    for (const auto &entry : clusterToCaloHitListMap)
+        extrapolatedCaloHitVector.insert(extrapolatedCaloHitVector.begin(), entry.second.begin(), entry.second.end());
+
+    std::sort(extrapolatedCaloHitVector.begin(), extrapolatedCaloHitVector.end(), SortByDistanceAlongLine(clusterAssociation.GetUpstreamMergePoint(),
+        clusterAssociation.GetConnectingLineDirection()));
```
PandoraPFA, line 322: Just have a quick think about whether it's likely you might get hits that have just about floating point identical sorting parameters; then it's not clear whether pairs of hits (which may come in in an arbitrary order) will always be ordered in the same manner. If the properties used to sort are within e.g. std::numeric_limits<float>::epsilon() of each other, it's often useful to have a tie-breaker available, such as the hit energy estimate (just pushes the problem to a higher order!). These are subtle things associated with working with unordered containers. Strictly, we ought to, across the board, work with vectors that hold their order (or are sorted, but will always be in the same way, as starting from defined initial order) and use the maps as look ups. We're doing OK here so long as we think about this carefully.

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+
+    CartesianPointVector trackSegmentBoundaries;
+    this->GetTrackSegmentBoundaries(clusterAssociation, trackSegmentBoundaries);
+
```
PandoraPFA, line 326: Worth having a check on `trackSegmentBoundaries.size()`, as below subtract 1 from size, and use `.at(i)` and `.at(i + 1)`. Can just return true/false as appropriate?

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+    return pMainTrackCluster;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackInEMShowerAlgorithm::AddHitsToMainTrack(const Cluster *const pShowerCluster, const Cluster *const pMainTrackCluster, const CaloHitList &caloHitsToMerge,
+    const ClusterAssociation &clusterAssociation, ClusterList &remnantClusterList) const
+{
+    // To ignore crossing CR muon or test beam tracks
+    if ((static_cast<float>(caloHitsToMerge.size()) / static_cast<float>(pShowerCluster->GetNCaloHits())) < m_minHitFractionForHitRemoval)
```
PandoraPFA, line 535: Any chance can have a shower cluster with no hits incoming? (Can only happen if they cluster has some isolated hits, but no non-isolated hits).

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+
+const Cluster *TrackInEMShowerAlgorithm::RemoveOffAxisHitsFromTrack(const Cluster *const pCluster, const CartesianVector &splitPosition, const bool isUpstream,
+    const ClusterToCaloHitListMap &clusterToCaloHitListMap, ClusterList &remnantClusterList, TwoDSlidingFitResultMap &microSlidingFitResultMap,
+    TwoDSlidingFitResultMap &macroSlidingFitResultMap, ClusterVector &clusterVector) const
+{
+    const auto extrapolatedCaloHitIter(clusterToCaloHitListMap.find(pCluster));
+
+    if (extrapolatedCaloHitIter == clusterToCaloHitListMap.end())
+    {
+        this->UpdateForClusterDeletion(pCluster, clusterVector, microSlidingFitResultMap, macroSlidingFitResultMap);
```
PandoraPFA, line 467: Is this the best place to be running this update for deletion, after checking in the clusterToCaloHitListMap? (The answer to this may well be "yes"!). Wondered whether there might be a better place. Good to have one last check through all the UpdateForClusterDeletion calls, as these are so tricky to make 100% robust.

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+                parameters.m_caloHitList.push_back(pCaloHit);
+                PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::Cluster::Create(*this, parameters, pClusterToModify));
+
+                if (pClusterToModify != pMainTrackCluster)
+                    remnantClusterList.push_back(pClusterToModify);
+            }
+        }
+    }
+
+    this->UpdateForClusterDeletion(pCluster, clusterVector, microSlidingFitResultMap, macroSlidingFitResultMap);
```
PandoraPFA, line 522: (This does seem like the right place for this update, but one more review of this and we should be happy 😄)

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
```diff
@@ -0,0 +1,877 @@
...
+
+bool TrackInEMShowerAlgorithm::SortByDistanceAlongLine::operator() (const pandora::CaloHit *const pLhs, const pandora::CaloHit *const pRhs)
+{
+    const CartesianVector lhsDistanceVector(pLhs->GetPositionVector() - m_startPoint);
+    const CartesianVector rhsDistanceVector(pRhs->GetPositionVector() - m_startPoint);
+
+    const float lhsProjectedDistance(lhsDistanceVector.GetDotProduct(m_lineDirection));
+    const float rhsProjectedDistance(rhsDistanceVector.GetDotProduct(m_lineDirection));
+
+    return (lhsProjectedDistance < rhsProjectedDistance);
```
PandoraPFA, line 808: (See earlier comment about sorting tie-breakers)

PandoraPFA:
(P.S. When we merge in feature/larpandoracontent_v03_17_00, that should resolve the failed CI check flagged above)

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
imawby: I'm pretty sure the answer here is no. The pShowerCluster is always a cluster that has a hit identified as being along the line that connects the split clusters. This tagging is made by looking at the cluster's ordered calo hit list and making a position check on each calo hit. So, as long as the ordered calo hit list (from pCluster->GetOrderedCaloHitList) only contains non-isolated hits then we are good. I think this is true since you mentioned that isolated calo hits do not contribute to the spatial info of the cluster

larpandoracontent/LArTwoDReco/LArCosmicRay/TrackInEMShowerAlgorithm.cc
imawby: These are a real pain, and this comment made me realise that I wasn't handling something the way I wanted to. Essentially this statement is saying 'if the main track split cluster has no hits in the extrapolation vector skip the fragmentation process'. What I was trying to say was 'If there is no clustering errors then skip the fragmentation process'. In reality when you have a clustering error you nearly always find hits in the extrapolation vector but I can imagine that this isn't always true. In these cases, the current implementation means that the clustering error just wouldn't be fixed. So, I now have a 'more correct implementation' and have one cluster deletion update. Which I think is placed in the best place since it is that we can guarantee that the cluster exists.

PR #126, opened 2020-06-15
Allow gcc 7 to compile an unused structured binding

AndyChappell:
Unlike later versions of GCC, in GCC 7.3 the structured binding feature generates a warning if any of the bindings are unused. This single commit allows the older compiler to successfully build with an unused binding.


johnmarshall80:
I might have been tempted to add a space after `(void)` but no strong opinion there! :smirk: Looks fine @AndyChappell - thanks and do feel free to proceed.

PR #127, opened 2020-06-18
Move to C++17 standard

AndyChappell:
This PR changes the default LArContent build to use the C++17 standard.


AndyChappell:
Branch is a bit of a mess after various travis/docker tests, closing for a cleaner final branch

PR #128, opened 2020-06-19
Move to C++17 standard

AndyChappell:
This PR changes the default LArContent build to use the C++17 standard.


johnmarshall80:
This all looks good to me - thanks very much indeed, Andy.

PR #131, opened 2020-07-10
Feature/two view matching tools

MariaBrigida:
This PR adds the TwoViewLongTracksTool (to perform matches in case of obvious ambiguities) and the TwoViewSimpleTracksTool (to select best match) to LArTwoViewMatching.


larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+bool TwoViewSimpleTracksTool::Run(TwoViewTransverseTracksAlgorithm *const pAlgorithm, MatrixType &overlapMatrix)
+{
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+       std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    ProtoParticleVector protoParticleVector;
+    this->FindBestTrack(overlapMatrix, protoParticleVector);
+
+    const bool particlesMade(pAlgorithm->CreateThreeDParticles(protoParticleVector));
+    return particlesMade;
```
lhwhitehead, line 37: Could just return the result from `CreateThreeDParticles` without assignment?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+
+        for (MatrixType::ElementList::const_reverse_iterator iIter = elementList.rbegin(), iIterEnd = elementList.rend(); iIter != iIterEnd; ++iIter)
+        {
+            if(this->PassesElementCuts(iIter))
+            {
+                bestElement = *iIter;
+               if (!bestElement.GetOverlapResult().IsInitialized())
+                    continue;
+
+                if ((NULL == bestElement.GetCluster1()) || (NULL == bestElement.GetCluster2()))
```
PandoraPFA, line 69: (Typically prefer `nullptr`, rather than `NULL` these days)

larpandoracontent/LArObjects/LArTrackTwoViewOverlapResult.h
```diff
@@ -199,6 +229,36 @@ inline unsigned int TwoViewTransverseOverlapResult::GetNSamplingPoints() const
...
+   
+    throw pandora::StatusCodeException(pandora::STATUS_CODE_NOT_INITIALIZED);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+inline unsigned int TwoViewTransverseOverlapResult::GetNReUpsampledSamplingPoints() const
+{
+    if (m_isInitialized)
+        return static_cast<unsigned int>(m_downsamplingFactor*m_nSamplingPoints);
```
absolution1, line 245: static cast the downsampling factor, rather than the entire thing.

larpandoracontent/LArObjects/LArTrackTwoViewOverlapResult.h
```diff
@@ -199,6 +229,36 @@ inline unsigned int TwoViewTransverseOverlapResult::GetNSamplingPoints() const
...
+   
+    throw pandora::StatusCodeException(pandora::STATUS_CODE_NOT_INITIALIZED);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+inline unsigned int TwoViewTransverseOverlapResult::GetNMatchedReUpsampledSamplingPoints() const
+{
+    if (m_isInitialized)
+        return static_cast<unsigned int>(m_downsamplingFactor*m_nMatchedSamplingPoints);
```
absolution1, line 255: static cast the downsampling factor, rather than the entire thing.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+            continue;
+
+        unsigned int n0(0), n1(0);
+        MatrixType::ElementList elementList;
+        overlapMatrix.GetConnectedElements(pKeyCluster, true, elementList, n0, n1);
+
+        if (elementList.empty())
+            continue;
+
+        MatrixType::Element bestElement(elementList.back());
```
PandoraPFA, line 59: Looks like `bestElement` doesn't need to exist outside the loop below any more (but probably did during development). Not doing any harm, though.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
```diff
@@ -0,0 +1,173 @@
...
+            return true;
+    }
+
+    return false;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewLongTracksTool::IsLongerThanDirectConnections(IteratorList::const_iterator iIter, const MatrixType::ElementList &elementList,
+    const unsigned int minMatchedSamplingPointRatio, const pandora::ClusterSet &usedClusters)
```
absolution1, line 46: Looking at where this function is called `minMatchedSamplingPointRatio` is always passed `m_minMatchedSamplingPointRatio`.
`minMatchedSamplingPointRatio` can probably be dropped as an argument and the function can always assess using `m_minMatchedSamplingPointRatio` instead.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+
+        for (MatrixType::ElementList::const_reverse_iterator iIter = elementList.rbegin(), iIterEnd = elementList.rend(); iIter != iIterEnd; ++iIter)
+        {
+            if(this->PassesElementCuts(iIter))
+            {
+                bestElement = *iIter;
+               if (!bestElement.GetOverlapResult().IsInitialized())
```
PandoraPFA, line 66: Could this `IsInitialized` check just be part of `PassesElementCuts`?

(Cosmetic: Indentation)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.h
```diff
@@ -0,0 +1,80 @@
...
+     *  @param  iteratorList to receive a list of iterators to long track-like elements
+     */
+    void SelectLongElements(const MatrixType::ElementList &elementList, const pandora::ClusterSet &usedClusters,
+        IteratorList &iteratorList) const;
+
+    float           m_minMatchedFraction;               ///< The min matched sampling point fraction for particle creation
+    float           m_minMatchingScore;                 ///< The min global matching score for particle creation
+    unsigned int    m_minMatchedSamplingPoints;         ///< The min number of matched sampling points for particle creation
+    float           m_minXOverlapFraction;              ///< The min x overlap fraction (in each view) for particle creation
+    unsigned int    m_minMatchedSamplingPointRatio;     ///< The min ratio between 1st and 2nd highest msps for simple ambiguity resolution
```
absolution1, line 75: I think we should probably make this a float to give us more granularity in this parameter.
It would mean updating wherever in the code `m_minMatchedSamplingPointRatio` to static_cast<float> anything this data member multiplies.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+        unsigned int n0(0), n1(0);
+        MatrixType::ElementList elementList;
+        overlapMatrix.GetConnectedElements(pKeyCluster, true, elementList, n0, n1);
+
+        if (elementList.empty())
+            continue;
+
+        MatrixType::Element bestElement(elementList.back());
+
+        for (MatrixType::ElementList::const_reverse_iterator iIter = elementList.rbegin(), iIterEnd = elementList.rend(); iIter != iIterEnd; ++iIter)
```
PandoraPFA, line 61: (Here's where you're relying on consistency with the OverlapResult `operator<` defined before.)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode TwoViewSimpleTracksTool::ReadSettings(const TiXmlHandle xmlHandle)
+{
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "MinMatchedFraction", m_minMatchedFraction));
+
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "MinMatchingScore", m_minMatchingScore));
+
```
PandoraPFA, line 116: (Cosmetic: Extra empty line)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
```diff
@@ -0,0 +1,173 @@
...
+        if ((*iIter) == eIter)
+            continue;
+
+        if (usedClusters.count(eIter->GetCluster1()) || usedClusters.count(eIter->GetCluster2()))
+            continue;
+
+        if (((*iIter)->GetCluster1() != eIter->GetCluster1()) && ((*iIter)->GetCluster2() != eIter->GetCluster2()))
+            continue;
+
+        if (nMatchedReUpsampledSamplingPoints < minMatchedSamplingPointRatio * eIter->GetOverlapResult().GetNMatchedReUpsampledSamplingPoints())
```
absolution1, line 61: See later comment about changing `minMatchedSamplingPointRatio` to a float.
This line would become a little ugly to look at:
`static_cast<int>(minMatchedSamplingPointRatio * static_cast<float>(eIter->GetOverlapResult().GetNMatchedReUpsampledSamplingPoints())))`.

There might also need to be some additional protection for `minMatchedSamplingPointRatio` accidentally going negative somewhere...

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
```diff
@@ -0,0 +1,173 @@
...
+bool TwoViewLongTracksTool::Run(TwoViewTransverseTracksAlgorithm *const pAlgorithm, MatrixType &overlapMatrix)
+{
+   if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+       std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    ProtoParticleVector protoParticleVector;
+    this->FindLongTracks(overlapMatrix, protoParticleVector);
+
+    const bool particlesMade(pAlgorithm->CreateThreeDParticles(protoParticleVector));
+    return particlesMade;
```
absolution1, line 79: I think Leigh already mentioned this one.
could be `return pAlgorithm->CreateThreeDParticles(protoParticleVector);`

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
```diff
@@ -0,0 +1,173 @@
...
+        IteratorList iteratorList;
+        this->SelectLongElements(elementList, usedClusters, iteratorList);
+
+        // Check that elements are not directly connected and are significantly longer than any other directly connected elements
+        for (IteratorList::const_iterator iIter = iteratorList.begin(), iIterEnd = iteratorList.end(); iIter != iIterEnd; ++iIter)
+        {
+            if (TwoViewLongTracksTool::HasLongDirectConnections(iIter, iteratorList))
+                continue;
+
+            if (!TwoViewLongTracksTool::IsLongerThanDirectConnections(iIter, elementList, m_minMatchedSamplingPointRatio, usedClusters))
```
absolution1, line 108: See above/below comments about dropping passing the sampling points ratio as an argument

absolution1, line 108: I think these function calls should use `this` rather than the namespace `TwoViewLongTracksTool::`

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
```diff
@@ -0,0 +1,173 @@
...
+        if (eIter->GetOverlapResult().GetMatchingScore() < m_minMatchingScore)
+            continue;
+
+        if (eIter->GetOverlapResult().GetNMatchedReUpsampledSamplingPoints() < m_minMatchedSamplingPoints)
+            continue;
+
+        const TwoViewXOverlap &xOverlap(eIter->GetOverlapResult().GetTwoViewXOverlap());
+
+        if ((xOverlap.GetXSpan0() > std::numeric_limits<float>::epsilon()) && (xOverlap.GetXOverlapFraction0() > m_minXOverlapFraction) &&
+            (xOverlap.GetXSpan1() > std::numeric_limits<float>::epsilon()) && (xOverlap.GetXOverlapFraction1() > m_minXOverlapFraction))
```
absolution1, line 144: I think the checks on `xOverlap.GetXSpan0()` and `xOverlap.GetXSpan1()` are redundant now that the fraction functions are used; I think those checks can be dropped from the logic.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.h
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     *  @brief  Whether a long element is significantly longer that other elements with which it shares a cluster
+     *
+     *  @param  iIter specifies the long element under consideration
+     *  @param  elementList the full list of connected matrix elements
+     *  @param  minMatchedSamplingPointRatio the min ratio between 1st and 2nd highest msps for simple ambiguity resolution
+     *  @param  usedClusters the list of clusters already marked as to be added to a pfo
+     */
+    static bool IsLongerThanDirectConnections(IteratorList::const_iterator iIter, const MatrixType::ElementList &elementList,
+        const unsigned int minMatchedSamplingPointRatio, const pandora::ClusterSet &usedClusters);
```
absolution1, line 46: See above comments about dropping `minMatchedSamplingPointRatio` as argument

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
absolution1, line 37: return `particlesMade(pAlgorithm->CreateThreeDParticles(protoParticleVector));`

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
absolution1, line 61: Eyeballing this, it looks like the only place `iTerEnd` is used is in the for loop declaration.
Could this be shortened to
`for (MatrixType::ElementList::const_reverse_iterator iIter = elementList.rbegin(); iIter != elementList.rend(); ++iIter)`

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+                if ((NULL == bestElement.GetCluster1()) || (NULL == bestElement.GetCluster2()))
+                    continue;
+
+                ProtoParticle protoParticle;
+                protoParticle.m_clusterList.push_back(bestElement.GetCluster1());
+                protoParticle.m_clusterList.push_back(bestElement.GetCluster2());
+                protoParticleVector.push_back(protoParticle);
+
+                return;
```
absolution1, line 77: I think this could be slightly optimised by not assigning `bestElement` anywhere and dereferencing this iterator instead e.g. `iIter->GetCluster1()`

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -0,0 +1,127 @@
...
+    if (eIter->GetOverlapResult().GetMatchingScore() < m_minMatchingScore)
+        return false;
+
+    if (eIter->GetOverlapResult().GetNMatchedReUpsampledSamplingPoints() < m_minMatchedSamplingPoints)
+        return false;
+
+    const TwoViewXOverlap &xOverlap(eIter->GetOverlapResult().GetTwoViewXOverlap());
+
+    if ((xOverlap.GetXSpan0() > std::numeric_limits<float>::epsilon()) && (xOverlap.GetXOverlapFraction0() > m_minXOverlapFraction) &&
+        (xOverlap.GetXSpan1() > std::numeric_limits<float>::epsilon()) && (xOverlap.GetXOverlapFraction1() > m_minXOverlapFraction))
```
absolution1, line 99: Similar comment for LongTracks: `xOverlap.GetXSpan0()` and `xOverlap.GetXSpan1()` can be dropped, I think.

absolution1: Completely separate comment and a purely stylistic one.
All of the above checks in this function are 'if not this then return false' which is the opposite way round to this check.
It might be neater to have this check be the same way round and have the final return be `return true`

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.h
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     *  @brief  Default constructor
+     */
+    TwoViewSimpleTracksTool();
+
+    bool Run(TwoViewTransverseTracksAlgorithm *const pAlgorithm, MatrixType &overlapMatrix);
+
+private:
+    /**
+     *  @brief  Find best shower match as a simple way to (try to) resolve ambiguities in the matrix
```
absolution1, line 31: typo: shower -> track

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
PandoraPFA: These ones were static in the original three-view long tracks tool, because they were re-used by at least one other tool. They have followed by being static here (not an issue, as they are "stateless" helper functions) and hence are prefixed in this way.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
MariaBrigida: This function is static, so it cannot simply access `m_minMatchedSamplingPointRatio`. What do you suggest?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: This seems to be the same point as in Leigh's comment above, will address both at once.

MariaBrigida: I think the bool does help readability a bit, so would be inclined to leave as it is, but let me know if you feel strongly about this and I can change it! I am also tagging @absolution1  who made a similar comment below.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
PandoraPFA: I'd suggest leave this as is, especially as there's already a precedent for doing it this way in the three-view case.

I think that this was a member variable originally, but it became an argument when the function was made static (which was to enable reuse by another tool).

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
PandoraPFA: No strong opinions on these returns. Feel free to pick 😄 Leaving them means that no action is required now. Removing them is straightforward, and makes things a little simpler. The compiler will likely produce the same output in either case, for our standard builds.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Done!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Okay, I've changed it as per suggestion :)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
MariaBrigida: Will leave it as it is then, thanks!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Done!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Done!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Seems reasonable! Done!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Removed best element, as not necessary!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: @PandoraPFA, did you have a specific suggestion regarding these lines? Thanks!

PandoraPFA: No, was rather just highlighting that this ordered depends exactly on the ordering you defined in the OverlapResult `operator<`, so any changes there would have an impact here. (Strictly, you're sorting the OverlapElements, but these are just OverlapResults, with the associated clusters, and redirect the sorting to the OverlapResult). So, looks OK to me, with the operator putting the "worst" elements first, then you using a reverse iterator to see the "best" elements first.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: I am not sure what you mean by: `xOverlap.GetXSpan0()` can be dropped?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Okay, thank you! :)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -148,12 +183,34 @@ float TwoViewTransverseTracksAlgorithm::CalculateLocalMatchingFraction(const Dis
...
 
-    const int nComparisons(static_cast<int>(discreteProbabilityVector1.GetSize()) - (static_cast<int>(m_minSamples) - 1));
-    if (1 > nComparisons)
-        throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
+//------------------------------------------------------------------------------------------------------------------------------------------
 
-    return (static_cast<float>(nMatchedComparisons) / static_cast<float>(nComparisons));
+float TwoViewTransverseTracksAlgorithm::GetPrimaryAxisDotDriftAxis(const pandora::Cluster *const pCluster)
+{
+    DotProductMap::const_iterator iter(m_dotProductMap.find(pCluster));
```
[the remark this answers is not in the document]
absolution1: So, there are a few cases where the map would need cleaning or updating
 - Moving to a new event
 - Cluster merging/splitting/other modification

Digging my way through the layers of inheritance and looking at the order of `Run`: https://github.com/MariaBrigida/LArContent/blob/feature/TwoViewMatching_tools/larpandoracontent/LArThreeDReco/LArThreeDBase/MatchingBaseAlgorithm.cc#L132

it seems to be that filling or clearing of the map would need to be done in (or an override of):
 - PrepareInputClusters: https://github.com/MariaBrigida/LArContent/blob/feature/TwoViewMatching_tools/larpandoracontent/LArThreeDReco/LArThreeDBase/NViewTrackMatchingAlgorithm.cc#L222
 - TidyUp: https://github.com/MariaBrigida/LArContent/blob/feature/TwoViewMatching_tools/larpandoracontent/LArThreeDReco/LArThreeDBase/TwoViewMatchingControl.cc#L187
 - UpdateForNewCluster: https://github.com/MariaBrigida/LArContent/blob/feature/TwoViewMatching_tools/larpandoracontent/LArThreeDReco/LArThreeDBase/TwoViewMatchingControl.cc#L49
 - Probably others

This is already shaping up to be a fair few changes that will need a bit of thought.  
IIRC, when developing this, I always recalculated the dot product for each overlap result and I do not remember the processing time being impacted much.  
I will run a quick test now to verify that but, assuming it is indeed quick to calculate the PCA + dot product for every overlap result, I'd like to suggest doing exactly that and dropping the cache for now.  We can figure out how best to implement the cache once we actually start merging/splitting clusters at the 2D->3D stage.

What do you think?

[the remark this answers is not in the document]
PandoraPFA: Yep, totally agree. Drop the cache until there's time for clear thought. We have examples (the sliding fit caches, for instance), but they are fiddly and exactly the kind of thing that goes awry when under time pressure. Thanks!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -130,16 +148,33 @@ float TwoViewTransverseTracksAlgorithm::CalculateLocalMatchingFraction(const Dis
...
-                localValues1, localValues2, randomNumberGenerator, m_nPermutations));
+            float localPValue(0);
+            try
+            {
+                localPValue = LArDiscreteProbabilityHelper::CalculateCorrelationCoefficientPValueFromPermutationTest(
+                    localValues1, localValues2, randomNumberGenerator, m_nPermutations);
+            }
+            catch (StatusCodeException &)
+            {
+                std::cout << "TwoViewTransverseTracksAlgorithm: failed to calculate correlation coefficient p-value for these numbers" << std::endl;;
```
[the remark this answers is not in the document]
absolution1: I think we discussed this one in slack; I think we agreed that it's fine for warnings to be printed out if they are meaningful.  This printout was put in place when we found the code was throwing an exception due to a bunch of adjacent sampling points being of equal value due to a large gap in the cluster.  We decided to catch the exception but now, without the added warning, we'd have no idea if this kind of problem was still happening or not.

It is not too verbose and prints out once or twice for a high multiplicity event.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -108,19 +115,30 @@ pandora::StatusCode TwoViewTransverseTracksAlgorithm::CalculateOverlapResult(con
...
     const float matchingScore(1.f - pvalue);
-    const float locallyMatchedFraction(CalculateLocalMatchingFraction(resampledDiscreteProbabilityVector1, 
+    if (matchingScore < m_minOverallMatchingScore)
+        return STATUS_CODE_NOT_FOUND;
+
+    const unsigned int nLocallyMatchedSamplingPoints(this->CalculateNumberOfLocallyMatchingSamplingPoints(resampledDiscreteProbabilityVector1, 
         resampledDiscreteProbabilityVector2, m_randomNumberGenerator));
+    const int nComparisons(static_cast<int>(resampledDiscreteProbabilityVector1.GetSize()) - (static_cast<int>(m_minSamples) - 1));
+    if (1 > nComparisons)
+        throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
```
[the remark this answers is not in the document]
absolution1: throw here, I think.
If the number of total comparisons is 0 then the logic has failed badly (it shouldn't have failed this badly), so the thing should bail.

The other returned non-successes are handled as they should be I think.  I did test this accidentally when I returned the wrong kind of status and the code unexpectedly threw.  The above returns do indeed return sensibly and the overlap is filled as intended.

The other throws that the alg does not explicitly catch work in the sense that they do indeed throw and the alg fails, which is what was intended.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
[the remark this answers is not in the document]
PandoraPFA: IIRC, dropping it _should_ just be a matter of changing the inheritance structure [here](https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h#L32-L35), but agree that let's go with what's been tested right now.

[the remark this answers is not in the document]
absolution1: Looks like a second or two of impact on an event that already takes ~28 seconds.  I think we can safely drop the cache for now.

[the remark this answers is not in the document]
absolution1: Done

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -148,12 +183,34 @@ float TwoViewTransverseTracksAlgorithm::CalculateLocalMatchingFraction(const Dis
...
+        pandora::CartesianPointVector pointVector;
+        LArClusterHelper::GetCoordinateVector(pCluster, pointVector);
+
+        pandora::CartesianVector centroid(0.f, 0.f, 0.f);
+        LArPcaHelper::EigenVectors eigenVecs;
+        LArPcaHelper::EigenValues eigenValues(0.f, 0.f, 0.f);
+        LArPcaHelper::RunPca(pointVector, centroid, eigenValues, eigenVecs);
+
+        const pandora::CartesianVector primaryAxis(eigenVecs.at(0));
+        const pandora::CartesianVector driftAxis(1.f, 0.f, 0.f);
```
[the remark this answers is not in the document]
absolution1: Yeah, I did honestly try looking in the various geometry lookups but couldn't find anything that would tell me what the drift axis was.  So, I begrudgingly hardcoded.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h
```diff
@@ -54,29 +54,44 @@ class TwoViewTransverseTracksAlgorithm : public NViewTrackMatchingAlgorithm<TwoV
...
-    unsigned int                m_nMaxMatrixToolRepeats;         ///< The maximum number of repeat loops over matrix tools
-    unsigned int                m_downsampleFactor;              ///< The downsampling (hit merging) applied to hits in the overlap region
-    unsigned int                m_minSamples;                    ///< The minimum number of samples needed for comparing charges
-    unsigned int                m_nPermutations;                 ///< The number of permutations for calculating p-values
-    float                       m_localMatchingScoreThreshold;   ///< The minimum score to classify a local region as matching
-    std::mt19937                m_randomNumberGenerator;         ///< The random number generator
+    MatrixToolVector            m_algorithmToolVector;                   ///< The algorithm tool vector
+
+    typedef std::unordered_map<const pandora::Cluster*, float> DotProductMap;
+    DotProductMap               m_dotProductMap;                         ///< Map between cluster and it's primary axis dot product with the dirft axis
```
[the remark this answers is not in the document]
absolution1: Dropped the cache so this has brutishly resolved itself.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -130,16 +148,33 @@ float TwoViewTransverseTracksAlgorithm::CalculateLocalMatchingFraction(const Dis
...
         {
-            const float localPValue(LArDiscreteProbabilityHelper::CalculateCorrelationCoefficientPValueFromPermutationTest(
-                localValues1, localValues2, randomNumberGenerator, m_nPermutations));
+            float localPValue(0);
+            try
+            {
+                localPValue = LArDiscreteProbabilityHelper::CalculateCorrelationCoefficientPValueFromPermutationTest(
+                    localValues1, localValues2, randomNumberGenerator, m_nPermutations);
+            }
+            catch (StatusCodeException &)
```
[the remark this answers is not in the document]
absolution1: Got it

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -130,16 +148,33 @@ float TwoViewTransverseTracksAlgorithm::CalculateLocalMatchingFraction(const Dis
...
+                std::cout << "TwoViewTransverseTracksAlgorithm: failed to calculate correlation coefficient p-value for these numbers" << std::endl;;
+                std::cout << "----view 0: ";
+                for (unsigned int iElement = 0; iElement < localValues1.size(); ++iElement)
+                    std::cout<<localValues1.at(iElement) << " ";
+                std::cout << std::endl;
+                std::cout << "----view 1: ";
+                for (unsigned int iElement = 0; iElement < localValues2.size(); ++iElement)
+                    std::cout<<localValues2.at(iElement) << " ";
+                std::cout << std::endl;
+
```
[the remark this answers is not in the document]
absolution1: Thanks a bunch.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
absolution1: `xOverlap.GetXSpan0() > std::numeric_limits<float>::epsilon()` was initially in place to check that the denominator in the fraction calculation wouldn't be 0.  The new function that's now in place checks this under the hood.  So, I don't think that part of the check is needed anymore.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
```diff
@@ -0,0 +1,173 @@
...
+        if (usedClusters.count(eIter->GetCluster1()) || usedClusters.count(eIter->GetCluster2()))
+            continue;
+
+        if (eIter->GetOverlapResult().GetLocallyMatchedFraction() < m_minMatchedFraction)
+            continue;
+
+        if (eIter->GetOverlapResult().GetMatchingScore() < m_minMatchingScore)
+            continue;
+
+        if (eIter->GetOverlapResult().GetNMatchedReUpsampledSamplingPoints() < m_minMatchedSamplingPoints)
```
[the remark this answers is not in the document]
absolution1: I think the names in the tool were chosen for parity with the three view case.  It's unfortunate that the overlap result data member and function have `ReUpsampled` in the name.  Those should probably be changed.  However, it's very much a style thing and should be deferred to down the line, I think.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
MariaBrigida: Okay, nice. I will drop them, then.

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.cc
```diff
@@ -63,21 +65,56 @@ bool TransverseMatrixVisualizationTool::Run(TwoViewTransverseTracksAlgorithm *co
...
+            try{
+                const MCParticle* particle0(MCParticleHelper::GetMainMCParticle(eIter->GetCluster1())); 
+                const MCParticle* particle1(MCParticleHelper::GetMainMCParticle(eIter->GetCluster2())); 
+                pdg0=(particle0->GetParticleId());
+                isPrimary0=(particle0->IsRootParticle());
+                pdg1=(particle1->GetParticleId());
+                isPrimary1=(particle1->IsRootParticle());
+                sameParticle=(particle0->GetUid() == particle1->GetUid());
+            }
+            catch(...){};
```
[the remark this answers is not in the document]
absolution1: yes and yes!

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.cc
```diff
@@ -55,6 +56,7 @@ bool TransverseMatrixVisualizationTool::Run(TwoViewTransverseTracksAlgorithm *co
         ClusterList allClusterList1, allClusterList2;
         std::cout << " Connections: n1 " << n1 << ", n2 " << n2 << ", nElements " << elementList.size() << std::endl;
 
+
```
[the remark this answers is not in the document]
absolution1: absolutely.

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.cc
```diff
@@ -63,21 +65,56 @@ bool TransverseMatrixVisualizationTool::Run(TwoViewTransverseTracksAlgorithm *co
...
+            int pdg0(0);
+            int pdg1(0);
+            bool isPrimary0(false);
+            bool isPrimary1(false);
+            bool sameParticle(false);
+            try{
+                const MCParticle* particle0(MCParticleHelper::GetMainMCParticle(eIter->GetCluster1())); 
+                const MCParticle* particle1(MCParticleHelper::GetMainMCParticle(eIter->GetCluster2())); 
+                pdg0=(particle0->GetParticleId());
+                isPrimary0=(particle0->IsRootParticle());
```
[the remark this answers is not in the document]
absolution1: Good point.  This will be a bit of extra work so I think we should defer this one, especially as its purely a monitoring thing.

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.cc
```diff
@@ -63,21 +65,56 @@ bool TransverseMatrixVisualizationTool::Run(TwoViewTransverseTracksAlgorithm *co
...
+                pdg1=(particle1->GetParticleId());
+                isPrimary1=(particle1->IsRootParticle());
+                sameParticle=(particle0->GetUid() == particle1->GetUid());
```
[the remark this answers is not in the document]
absolution1: It is.  Call me old-fashioned but checking IDs is engrained.

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.h
```diff
@@ -29,9 +29,11 @@ class TransverseMatrixVisualizationTool : public TransverseMatrixTool
...
     pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
 
-    unsigned int    m_minClusterConnections;        ///< The minimum number of cluster connections for display
-    bool            m_ignoreUnavailableClusters;    ///< Whether to ignore (skip-over) unavailable clusters in the matrix
-    bool            m_showEachIndividualElement;    ///< Whether to draw each individual matrix element
+    unsigned int    m_minClusterConnections;                   ///< The minimum number of cluster connections for display
+    bool            m_ignoreUnavailableClusters;               ///< Whether to ignore (skip-over) unavailable clusters in the matrix
+    bool            m_showEachIndividualElement;               ///< Whether to draw each individual matrix element
+    bool            m_showOnlyTrueMatchIndividualElements;     ///< Whether to draw only truly matching individual matrix elements
+
```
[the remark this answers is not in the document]
absolution1: Got it!

larpandoracontent/LArObjects/LArTrackTwoViewOverlapResult.h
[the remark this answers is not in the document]
absolution1: Yes! Got it.

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.cc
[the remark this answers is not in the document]
absolution1: Got em.  Thanks @absolution1

larpandoracontent/LArMonitoring/TransverseMatrixVisualizationTool.cc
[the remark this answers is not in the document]
absolution1: Done.

larpandoracontent/LArObjects/LArTrackTwoViewOverlapResult.h
absolution1: Actually don't do that, but static cast `m_nMatchedSamplingPoints` to a float and then keep the unsigned int static cast for the entire thing.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewClearTracksTool.cc
```diff
@@ -19,7 +19,8 @@ namespace lar_content
 
 TwoViewClearTracksTool::TwoViewClearTracksTool() :
     m_minXOverlapFraction(0.1f),
-    m_minLocallyMatchedFraction(0.5f)
+    m_minMatchingScore(0.95),
```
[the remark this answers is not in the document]
absolution1: Got it!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.h
MariaBrigida: See my answers above! We decided to leave this argument in.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
MariaBrigida: Will leave as they are then!

MariaBrigida: Leaving this as it is

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
MariaBrigida: Done!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
```diff
@@ -95,13 +88,16 @@ bool TwoViewSimpleTracksTool::PassesElementCuts(MatrixType::ElementList::const_r
...
-        (xOverlap.GetXSpan1() > std::numeric_limits<float>::epsilon()) && (xOverlap.GetXOverlapFraction1() > m_minXOverlapFraction))
+    if (!((xOverlap.GetXOverlapFraction0() > m_minXOverlapFraction) &&
+         (xOverlap.GetXOverlapFraction1() > m_minXOverlapFraction)))
     {
-        return true;
+        return false;
     }
 
-    return false;
+    if (!eIter->GetOverlapResult().IsInitialized())
```
PandoraPFA, line 97: This check should go first, at the top of this function, otherwise I think some of the function calls above, e.g. GetXOverlapFraction0(), etc. might raise an exception.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h
```diff
@@ -80,8 +80,8 @@ class TwoViewTransverseTracksAlgorithm : public NViewTrackMatchingAlgorithm<TwoV
     typedef std::vector<TransverseMatrixTool*> MatrixToolVector;
     MatrixToolVector            m_algorithmToolVector;                   ///< The algorithm tool vector
 
-    typedef std::unordered_map<const pandora::Cluster*, float> DotProductMap;
-    DotProductMap               m_dotProductMap;                         ///< Map between cluster and it's primary axis dot product with the dirft axis
+    //typedef std::unordered_map<const pandora::Cluster*, float> DotProductMap;
+    //DotProductMap               m_dotProductMap;                         ///< Map between cluster and it's primary axis dot product with the dirft axis
```
PandoraPFA, line 84: Might as well delete these.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.h
MariaBrigida: This will be addressed in a future PR!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewLongTracksTool.cc
MariaBrigida: This will be addressed in a future PR!

PandoraPFA:
Merged via commandline - thanks all!

PR #132, opened 2020-08-14
Feature/dp cr pcheck

echardon:
The DP team added a sanity check verifying that only 2D clusters that have hits, at least partially, in a common daughter volume should be matched. 

We added some Geometry and Cluster helper functions to perform the check. 
It is done in two algorithms : 1) TwoViewTransverseTrack 2)CosmicRayTrackRecovery(this one actually touches Single-Phase)

A small addition was also made in the MasterAlgorithm.cc in order to initialize properly the new LArCaloHit object that now have a new member variable called `m_daughterVolumeId`.

A parallel pull request has been issued for larpandora repository (https://github.com/PandoraPFA/larpandora/pull/5) regarding the implementation of daughter volumes for DP and the addition of the new member variable in LArCaloHit.


larpandoracontent/LArHelpers/LArClusterHelper.cc
```diff
@@ -642,6 +643,27 @@ void LArClusterHelper::GetCaloHitListInBoundingBox(const pandora::Cluster *const
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+std::set<unsigned int> LArClusterHelper::GetDaughterVolumeIDs(const Cluster *const pCluster)
+{
+    std::set<unsigned int> daughterVolumeIds;
```
AndyChappell, line 648: Prefer `pandora::UIntSet` to `std::set<unsigned int>`.

larpandoracontent/LArHelpers/LArClusterHelper.cc
```diff
@@ -642,6 +643,27 @@ void LArClusterHelper::GetCaloHitListInBoundingBox(const pandora::Cluster *const
...
+
+    for (OrderedCaloHitList::const_iterator ochIter = orderedCaloHitList.begin(), ochIterEnd = orderedCaloHitList.end(); ochIter != ochIterEnd; ++ochIter)
+    {
+        for (CaloHitList::const_iterator hIter = ochIter->second->begin(), hIterEnd = ochIter->second->end(); hIter != hIterEnd; ++hIter)
+        {
+            const CaloHit *const pCaloHit = *hIter;
+            const LArCaloHit *const pLArCaloHit(dynamic_cast<const LArCaloHit*>(pCaloHit));
+
+            if (pLArCaloHit) 
+                daughterVolumeIds.insert(pLArCaloHit->GetDaughterVolumeId());
```
AndyChappell, line 659: If the dynamic cast fails, I guess this implies a significant issue with the hits that probably warrants an exception?

larpandoracontent/LArHelpers/LArClusterHelper.h
```diff
@@ -274,6 +274,15 @@ class LArClusterHelper
...
         const pandora::CartesianVector &upperBound, pandora::CaloHitList &caloHitList);
 
+    /**
+     *  @brief  Get the set of the daughter volumes that contains the cluster
+     *
+     *  @param  pCluster1 address of the cluster
+     *
+     *  @return the set of daughter volumes
+     */
+    static std::set<unsigned int> GetDaughterVolumeIDs(const pandora::Cluster *const pCluster);
```
AndyChappell, line 284: UIntSet

larpandoracontent/LArHelpers/LArClusterHelper.h
```diff
@@ -274,6 +274,15 @@ class LArClusterHelper
     static void GetCaloHitListInBoundingBox(const pandora::Cluster *const pCluster, const pandora::CartesianVector &lowerBound, 
         const pandora::CartesianVector &upperBound, pandora::CaloHitList &caloHitList);
 
+    /**
+     *  @brief  Get the set of the daughter volumes that contains the cluster
+     *
+     *  @param  pCluster1 address of the cluster
```
AndyChappell, line 280: pCluster rather than pCluster1

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -400,6 +400,21 @@ CartesianVector LArGeometryHelper::GetWireAxis(const Pandora &pandora, const Hit
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+std::set<unsigned int> LArGeometryHelper::GetCommonDaughterVolumes(const Cluster *const pCluster1, const Cluster *const pCluster2)
+{
+    std::set<unsigned int> intersect;
+
+    const std::set<unsigned int> daughterVolumeIds1(LArClusterHelper::GetDaughterVolumeIDs(pCluster1)),
```
AndyChappell, line 407: UIntSet

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -400,6 +400,21 @@ CartesianVector LArGeometryHelper::GetWireAxis(const Pandora &pandora, const Hit
...
+{
+    std::set<unsigned int> intersect;
+
+    const std::set<unsigned int> daughterVolumeIds1(LArClusterHelper::GetDaughterVolumeIDs(pCluster1)), 
+        daughterVolumeIds2(LArClusterHelper::GetDaughterVolumeIDs(pCluster2));
+
+    std::set_intersection(daughterVolumeIds1.begin(), daughterVolumeIds1.end(), daughterVolumeIds2.begin(), daughterVolumeIds2.end(), 
+                          std::inserter(intersect, intersect.begin())); 
+
+    return intersect;
```
AndyChappell, line 413: Might be preferable to pass in the set by reference and avoid the copy constructor calls for the return value.

larpandoracontent/LArHelpers/LArGeometryHelper.h
```diff
@@ -197,6 +197,14 @@ class LArGeometryHelper
      */
     static pandora::CartesianVector GetWireAxis(const pandora::Pandora &pandora, const pandora::HitType view);
 
+    /**
+     *  @brief  Return the set of common daughter volumes between two 2D clusters
+     *
+     *  @param  Cluster 1
+     *  @param  Cluster 2
```
AndyChappell, line 204: Missing variable names

larpandoracontent/LArHelpers/LArGeometryHelper.h
```diff
@@ -197,6 +197,14 @@ class LArGeometryHelper
...
+     */
+    static std::set<unsigned int> GetCommonDaughterVolumes (const pandora::Cluster *const pCluster1, const pandora::Cluster *const pCluster2);
```
AndyChappell, line 206: UIntSet

ups/product_deps
```diff
@@ -13,7 +13,7 @@ fcldir  -
 # Add the dependent product and version
 product		version
 eigen		v3_3_5
-pandora		v03_11_01g
+pandora		v03_15_00
 cetbuildtools	v7_14_00	-	only_for_build
 end_product_list
```
AndyChappell, line 18: Revert product_deps changes.

larpandoracontent/LArHelpers/LArClusterHelper.cc
MariaBrigida: Etienne will comment here, but it looks like checks of this type are performed in MasterAlgorithm.cc as well.
As a separate note, I was wondering if we care to avoid storing the same DaughterVolumeId more than once?

AndyChappell: The check should definitely be performed, but in the MasterAlgorithm, if this check fails (aside from one specific case), exceptions are thrown, because it usually implies a fundamental problem if you weren't able to cast the CaloHit to a LArCaloHit, and so shouldn't attempt further processing.

larpandoracontent/LArThreeDReco/LArCosmicRay/CosmicRayTrackRecoveryAlgorithm.cc
```diff
@@ -196,6 +196,11 @@ void CosmicRayTrackRecoveryAlgorithm::MatchClusters(const Cluster* const pSeedCl
     {
         const Cluster *const pTargetCluster = *tIter;
 
+        const std::set daughterVolumeIntersection(LArGeometryHelper::GetCommonDaughterVolumes(pSeedCluster, pTargetCluster));
+        if (daughterVolumeIntersection.empty())
+            continue;
```
MariaBrigida, line 201: We could consider throwing an exception if an intersection is not found here.

AndyChappell: I think in this case `continue` appears to be the right approach, as the code appears to be deciding whether or not to consider the seed and target clusters for matching. An exception here would terminate the function for all subsequent targets, which presumably is not the desired behaviour?

MariaBrigida: I was looking at what happens in the rest of the code - where exceptions are thrown just a few lines below. 

(Regardless of whether we leave this check as it is, I would perhaps move it down after the check at lines 204/205!)

larpandoracontent/LArHelpers/LArClusterHelper.cc
MariaBrigida: Yep, I can see an exception is thrown in the copy constructor in the Master Algorithm, and I think it's definitely necessary there, perhaps less so here - but of course happy for us to add it if that's the consensus!

MariaBrigida: Yep, I can see an exception is thrown in the copy constructor in the Master Algorithm, and I think it's definitely necessary there, perhaps less so here - but of course happy for us to add it if that's the consensus!

echardon:
Issue with the commit history, a new clean branch has been made and a new pull request has been opened : https://github.com/PandoraPFA/LArContent/pull/142

PR #133, opened 2020-08-17
Feature/bdt pfo characterisation rebase

MoteyMousam:
New PR against v03_20_00 for bdt pfo characterisation approach after rebasing


larpandoracontent/LArTrackShowerId/MvaPfoCharacterisationAlgorithm.cc
```diff
@@ -311,6 +318,14 @@ StatusCode MvaPfoCharacterisationAlgorithm<T>::ReadSettings(const TiXmlHandle xm
...
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "SvmName", m_mvaName));
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "MvaName", m_mvaName));
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "WriteToTree", m_writeToTree)); // added by Mousam
+
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "OutputTree", m_treeName)); // added by Mousam
+    
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "OutputFile", m_fileName)); //added by Mousam
```
AndyChappell, line 328: If WriteToTree is set should probably add a conditional block that requires the presence of the output tree and filenames, as in this form, if these are missing in the XML file then they will default to being empty without any complaint from ReadSettings.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -469,9 +470,16 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
         try
         {
             vertexDistance = (pVertexList->front()->GetPosition() - LArPfoHelper::GetVertex(pInputPfo)->GetPosition()).GetMagnitude();
+            error = false;
         }
         catch (const StatusCodeException &) {}
     }
+    if (error == true)
```
AndyChappell, line 477: Can dispense with the == true here.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -469,9 +470,16 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
...
+    {
+        CaloHitList threeDCaloHitList;
+        LArPfoHelper::GetCaloHits(pInputPfo, TPC_3D, threeDCaloHitList);
+        vertexDistance = (pVertexList->front()->GetPosition() - (threeDCaloHitList.front())->GetPositionVector()).GetMagnitude(); // if n3dHits == 1, can't calculate vertex postion of input pfos hence default to 10.0 based on this variable's distribution to make it unbiased towards tracks or showers
```
AndyChappell, line 481: It's not obvious to me how this calculation corresponds to the comment, can you comment further on this?

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -653,11 +659,18 @@ void ThreeDPCAFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featureVector, co
         {
             LArPcaHelper::RunPca(threeDCaloHitList, centroid, eigenValues, eigenVecs);
             const float principalEigenvalue(eigenValues.GetX()), secondaryEigenvalue(eigenValues.GetY()), tertiaryEigenvalue(eigenValues.GetZ());
+
             if (principalEigenvalue > std::numeric_limits<float>::epsilon())
             {
                 pca1 = secondaryEigenvalue/principalEigenvalue;
                 pca2 = tertiaryEigenvalue/principalEigenvalue;
             }
+
```
AndyChappell, line 668: Remove whitespace. May want to run the remove whitespace script to remove some trailing spaces too (recommend isolating this to a single commit).

CMakeLists.txt
```diff
@@ -95,6 +95,11 @@ else()
     find_package(Eigen3 3.3 REQUIRED NO_MODULE)
     include_directories(SYSTEM ${EIGEN3_INCLUDE_DIRS})
 
+    find_package(ROOT 5.26.00 REQUIRED COMPONENTS TMVA)
```
AndyChappell, line 98: This should use ROOT 6.18.04

larpandoracontent/LArTrackShowerId/MvaPfoCharacterisationAlgorithm.cc
MariaBrigida: Only cosmetic, but I'd probably remove the white lines too between the various PANDORA_RETURN_RESULT_IF_AND_IF, for consistency.

MoteyMousam: @AndyChappell Dont think that is require as this can be changed in the neutrino settings file, right? If you dont want to produce root file then you can set <WriteToTree>to "false". I thought that was enough. If you do that, then there shouldn't be any errors.

MoteyMousam: @MariaBrigida Done!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
MoteyMousam: Done!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
MoteyMousam: Done!

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
MoteyMousam: This basically says that if you have a PFO with 1 3D hit (which is suprisingly possible), for some reason, it fails to execute line 472: `LArPfoHelper::GetVertex(pInputPfo)->GetPosition()).GetMagnitude()`.  This ends up throwing an error. To bypass this, I just asked for the position of the 1 3d hit instead and then set vertexDistance to be the magnitude of the difference between the interaction vertex and the hit's position.

larpandoracontent/LArTrackShowerId/MvaPfoCharacterisationAlgorithm.cc
AndyChappell: You can set the flag to not produce the root file, but my concern here is that if you do set the flag to produce the root file, you could in principle forget to set the tree name and filename in XML, and as these don't appear to be set anywhere by default, I think you would run into trouble when trying to write the output file - in which case I think it's worth enforcing the presence of the name variables in the event that the output flag is set to true.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
AndyChappell: Thanks for the confirmation; though I think that makes the comment alongside line 481 slightly outdated - presumably the result of this calculation is not always 10?

MoteyMousam: This is correct. My bad. I will fix it soon.

larpandoracontent/LArTrackShowerId/MvaPfoCharacterisationAlgorithm.cc
MoteyMousam: Maybe I am not getting what you are saying. I was under the assumption that <OutputTree> and <OutputFile> flags were available in the neutrino settings file too (https://github.com/MoteyMousam/LArReco/blob/0d59898780377b86dc604981f63cd58c21e4e96f/settings/PandoraSettings_Neutrino_DUNEFD.xml#L371). I could add something similar (i think) in line 48 of the algorithm (https://github.com/MoteyMousam/LArContent/blob/ce0f96db61c65d962f3169d04717cdac6785de58/larpandoracontent/LArTrackShowerId/MvaPfoCharacterisationAlgorithm.cc#L48). Would it not be more cleaner to use the neutrino settings xml file instead of hard coding the requirement into the alg?

MoteyMousam: @AndyChappell done

PR #136, opened 2020-09-02
feature/pdsp_performance

AndyChappell:
This PR contains computational performance improvements related to the calculation of cluster spans. In particular it caches the result of GetClusterSpanX to avoid needless re-computations (the same approach cannot be applied to GetClusterSpanZ, though this function is also moved for better coherence). This PR depends upon changes to the Cluster class in the SDK.


PandoraPFA (commented):
Many thanks - this is all fine. May just need to update if the GetClusterSpanZ interface changes in the associated PandoraSDK PR.

PR #140, opened 2020-10-06
Decorate MC with 3D info

CrossR:
I'm going through and doing some final tidy up to my 3D hit work, and as part of that splitting it up if possible.

The most obvious part to split off to start with is this, which populates the 3D positions and momentas for the `LArMCParticle`. Whilst not needed for the running of my code (barring the metrics I was making), this seems broadly useful so worth keeping around. Its mostly 2 commits from Steve, then a tidy up commit from me.

The one thing I wasn't sure on that we should probably change is that as far as I can see, this is always on. That is, if we merge this as is, it would mean every PNDR/XML file will contain 3D MC info, which seems excessive. Should I look into stopping that, to keep the files slightly leaner? It wasn't immediately obvious to me the best way to do that, which is why I left it out for now.

A partner PR over in LArPandora can follow, once we've decided on that (its as trivial as you'd guess, make a vector, fill it, assign it to the `mcParticleParameters` in `LArPandoraInput.cxx`).


larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -27,7 +27,9 @@ namespace lar_content
 class LArMCParticleParameters : public object_creation::MCParticle::Parameters
 {
 public:
-    pandora::InputInt   m_nuanceCode;               ///< The nuance code
+    pandora::InputInt                          m_nuanceCode;        ///< The nuance code
+    std::vector<pandora::InputCartesianVector> m_mcStepPositions;   ///< The positions of the geant4 steps
+    std::vector<pandora::InputCartesianVector> m_mcStepMomentas;    ///< The momenta of the geant4 steps
```
CrossR, line 32: Steve's original code uses these `InputCartesianVector`s, which I've left as is, but wasn't sure if/what benefit they had over the `CartesianVector` in this case.

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +253,55 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
     if (!pLArMCParticle)
         return pandora::STATUS_CODE_INVALID_PARAMETER;
 
+    const std::vector<pandora::CartesianVector> &mcStepPositions = pLArMCParticle->GetMCStepPositions();
+    int nMCStepPositions = mcStepPositions.size();
+    std::cout << "LArMCParticleFactory::Write - nMCStepPositions = " << nMCStepPositions << std::endl;
```
CrossR, line 258: Can drop the debug calls, if required. I'm actually not sure what the precedent is around this for writing PNDR/XML files.

absolution1:
Hello @CrossR 
I'll start taking a look now.
If there needs to be an accompanying PR for larpandora then one option for controlling whether the 3D info is saved could be a FCL switch.  I think there's already a FCL switch for enabling/disabling saving the MC particle info....

larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
```diff
@@ -150,6 +151,20 @@ void VisualMonitoringAlgorithm::VisualizeMCParticleList(const std::string &listN
         }
     }
 
+    for (const MCParticle *pMCParticle : *pMCParticleList)
+    {
+        const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle*>(pMCParticle));
+        std::vector<pandora::CartesianVector> steps(pLArMCParticle->GetMCStepPositions());
+        for (unsigned int step = 0; step < steps.size() - 1; step++)
```
absolution1, line 158: I think the usual style is `++step`

larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
```diff
@@ -150,6 +151,20 @@ void VisualMonitoringAlgorithm::VisualizeMCParticleList(const std::string &listN
         }
     }
 
+    for (const MCParticle *pMCParticle : *pMCParticleList)
+    {
+        const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle*>(pMCParticle));
+        std::vector<pandora::CartesianVector> steps(pLArMCParticle->GetMCStepPositions());
```
absolution1, line 157: could this be const?

larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
```diff
@@ -150,6 +151,20 @@ void VisualMonitoringAlgorithm::VisualizeMCParticleList(const std::string &listN
...
+        for (unsigned int step = 0; step < steps.size() - 1; step++)
+        {
+            pandora::CartesianVector start(steps.at(step));
+            pandora::CartesianVector end(steps.at(step + 1));
```
absolution1, line 161: Could these be consts (or even const ref)?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -52,8 +54,24 @@ class LArMCParticle : public object_creation::MCParticle::Object
...
+     *  @return vector of the positions
+     */
+    std::vector<pandora::CartesianVector> GetMCStepPositions() const;
+
+    /**
+     *  @brief  Get the momenta of the geant4 steps
+     *
+     *  @return vector of the momenta
+     */
+    std::vector<pandora::CartesianVector> GetMCStepMomentas() const;
```
absolution1, line 69: I'm wondering about the feasibility of these returning a const ref rather than returning by value (to avoid the potentially heavy cost of copying).  Unfortunately, things like RVO don't help here, I think.
Maybe ponder this one for a bit.

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -103,6 +121,15 @@ inline LArMCParticle::LArMCParticle(const LArMCParticleParameters &parameters) :
     object_creation::MCParticle::Object(parameters),
     m_nuanceCode(parameters.m_nuanceCode.Get())
 {
+    for (auto const &mcStepPosition : parameters.m_mcStepPositions)
+    {
```
absolution1, line 125: Because there's only one nested line in each of the two for loops, you can probably drop the braces

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -103,6 +121,15 @@ inline LArMCParticle::LArMCParticle(const LArMCParticleParameters &parameters) :
...
+        m_mcStepPositions.push_back(mcStepPosition.Get());
```
absolution1, line 126: Would `emplace_back` be a bit faster here?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -103,6 +121,15 @@ inline LArMCParticle::LArMCParticle(const LArMCParticleParameters &parameters) :
...
+    }
+
+    for (auto const &mcStepMomenta : parameters.m_mcStepMomentas)
+    {
```
absolution1, line 130: You can probably drop the braces here.

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -103,6 +121,15 @@ inline LArMCParticle::LArMCParticle(const LArMCParticleParameters &parameters) :
...
+        m_mcStepMomentas.push_back(mcStepMomenta.Get());
```
absolution1, line 131: Would `emplace_back` be more efficient here as well?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
     // ATTN: To receive this call-back must have already set file reader mc particle factory to this factory
     int nuanceCode(0);
 
+    int nMCStepPositions(0);
+    std::vector<pandora::InputCartesianVector> mcStepPositions;
+
+    int nMCStepMomentas(0);
```
absolution1, line 184: Could these ints be unsigned?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
     {
         pandora::BinaryFileReader &binaryFileReader(dynamic_cast<pandora::BinaryFileReader&>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nuanceCode));
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nMCStepPositions));
+        for (int i = 0; i < nMCStepPositions; ++i)
+        {
+            pandora::CartesianVector mcStepPosition(0.0,0.0,0.0);
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(mcStepPosition));
+            mcStepPositions.push_back(pandora::InputCartesianVector(mcStepPosition));
```
absolution1, line 197: emplace_back for efficiency?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
+        }
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nMCStepMomentas));
+        for (int i = 0; i < nMCStepMomentas; ++i)
+        {
+            pandora::CartesianVector mcStepMomenta(0.0,0.0,0.0);
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(mcStepMomenta));
+            mcStepMomentas.push_back(pandora::InputCartesianVector(mcStepMomenta));
```
absolution1, line 205: emplace_back?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
+    int nMCStepMomentas(0);
+    std::vector<pandora::InputCartesianVector> mcStepMomentas;
+
     if (pandora::BINARY == fileReader.GetFileType())
     {
         pandora::BinaryFileReader &binaryFileReader(dynamic_cast<pandora::BinaryFileReader&>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nuanceCode));
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nMCStepPositions));
+        for (int i = 0; i < nMCStepPositions; ++i)
```
absolution1, line 193: I think the index is usually descriptive, although the loop is short here.

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("NuanceCode", nuanceCode));
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("NumberOfMCStepPositions", nMCStepPositions));
+        for (int i = 0; i < nMCStepPositions; ++i)
+        {
+            pandora::CartesianVector mcStepPosition(0.0,0.0,0.0);
+            std::stringstream mcStepPositionName;
+            mcStepPositionName << "MCStepPosition" << i;
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable(mcStepPositionName.str(), mcStepPosition));
+            mcStepPositions.push_back(pandora::InputCartesianVector(mcStepPosition));
```
absolution1, line 220: `emplace_back`?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
+        }
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("NumberOfMCStepMomentas", nMCStepMomentas));
+        for (int i = 0; i < nMCStepMomentas; ++i)
+        {
+            pandora::CartesianVector mcStepMomenta(0.0,0.0,0.0);
+            std::stringstream mcStepMomentaName;
+            mcStepMomentaName << "MCStepMomenta" << i;
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable(mcStepMomentaName.str(), mcStepMomenta));
+            mcStepMomentas.push_back(pandora::InputCartesianVector(mcStepMomenta));
```
absolution1, line 230: e_b?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +253,55 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
     if (!pLArMCParticle)
         return pandora::STATUS_CODE_INVALID_PARAMETER;
 
+    const std::vector<pandora::CartesianVector> &mcStepPositions = pLArMCParticle->GetMCStepPositions();
```
absolution1, line 256: Could a constructor, rather than assignment, be used here?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +253,55 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
...
+    int nMCStepPositions = mcStepPositions.size();
```
absolution1, line 257: construct (and const)?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +253,55 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
...
+    std::cout << "LArMCParticleFactory::Write - nMCStepPositions = " << nMCStepPositions << std::endl;
+
+    const std::vector<pandora::CartesianVector> &mcStepMomentas = pLArMCParticle->GetMCStepMomentas();
```
absolution1, line 260: construct?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +253,55 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
...
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable("NumberOfMCStepMomentas", nMCStepMomentas));
+        int mcStepMomentasCounter(0);
+        for (auto const &mcStepMomenta : mcStepMomentas)
+        {
+            std::stringstream mcStepMomentaName;
+            mcStepMomentaName << "MCStepMomenta" << mcStepMomentasCounter;
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable(mcStepMomentaName.str(), mcStepMomenta));
+            ++mcStepMomentasCounter;
+        }
```
absolution1, line 304: If a counter is needed then would an old fashioned loop work here?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -103,6 +121,15 @@ inline LArMCParticle::LArMCParticle(const LArMCParticleParameters &parameters) :
     object_creation::MCParticle::Object(parameters),
     m_nuanceCode(parameters.m_nuanceCode.Get())
 {
+    for (auto const &mcStepPosition : parameters.m_mcStepPositions)
```
absolution1, line 124: Could these two vectors be initialised in the initialiser list? e.g.
`m_mcStepPositions(parameters.m_mcStepPositions)`

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +178,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
+
     if (pandora::BINARY == fileReader.GetFileType())
     {
         pandora::BinaryFileReader &binaryFileReader(dynamic_cast<pandora::BinaryFileReader&>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nuanceCode));
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nMCStepPositions));
+        for (int i = 0; i < nMCStepPositions; ++i)
+        {
+            pandora::CartesianVector mcStepPosition(0.0,0.0,0.0);
```
absolution1, line 195: Purely cosmetic but you could probably rely on the default constructor here?

absolution1 (commented):
A nice, factorised-out, PR @CrossR 
I've added some minor comments for you to peruse (and optionally ignore) at your leisure.
Most of the comments are "would emplace_back() be more efficient here?"

larpandoracontent/LArObjects/LArMCParticle.h
CrossR: No default constructor for `CartesianVector`s unfortunately.

larpandoracontent/LArObjects/LArMCParticle.h
AndyChappell: This is largely about providing a wrapper that can check the initialisation status for values that come from external sources.

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +249,51 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
...
+        for (auto const &mcStepPosition : mcStepPositions)
+        {
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(mcStepPosition));
+        }
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(nMCStepMomentas));
+        for (auto const &mcStepMomenta : mcStepMomentas)
+        {
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(mcStepMomenta));
+        }
```
AndyChappell, line 273: I'd drop the braces for the for loops here, given that they're one liners.

AndyChappell (commented):
Aside from a couple of trivial comments, this all looks fine to me, thanks Ryan. I won't explicitly approve it yet, as this will no doubt need rebasing and moving to a suitable release branch once the related PRs are ready as well.

AndyChappell (commented):
Updates look good.

CrossR:
I think I still need to do a check pass over this, since right now it will enforce that 3D hits are always added, and always read in, which I'm not convinced is desirable.

Since (to my knowledge) there is no way to easily check if a file contains this and react accordingly due to the binary file format we use, I think I'll just need to add a config option, to explicity say "do/don't read 3D MC".

larpandoracontent/LArObjects/LArMCParticle.h
AndyChappell: I think a const reference makes sense here. The returned variable is a member variable, so there's no risk of returning a temporary and introducing compiler dependency for RVO, so the only question would be whether the returned vector might be edited at some point, at which time an explicit copy could be made as needed?

larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
```diff
@@ -150,6 +151,23 @@ void VisualMonitoringAlgorithm::VisualizeMCParticleList(const std::string &listN
         }
     }
 
+    for (const MCParticle *pMCParticle : *pMCParticleList)
+    {
+        const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle*>(pMCParticle));
+        const std::vector<pandora::CartesianVector> steps(pLArMCParticle->GetMCStepPositions());
```
PandoraPFA, line 157: Suggest const reference. There's definitely a typedef for `std::vector<pandora::CartesianVector>` if you'd like to use that (`CartesianPointList`?), and I think this is all `using namespace pandora`.

larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
```diff
@@ -150,6 +151,23 @@ void VisualMonitoringAlgorithm::VisualizeMCParticleList(const std::string &listN
...
+
+        for (unsigned int step = 0; step < steps.size() - 1; ++step)
```
PandoraPFA, line 159: What happens here if `steps` is empty? Do we try to take 1 away from 0 and assign it to an unsigned int?

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -168,15 +249,47 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
...
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable("NumberOfMCStepPositions", nMCStepPositions));
+        for (int step = 0; step < nMCStepPositions; ++step)
+        {
+            const pandora::CartesianVector &position(mcStepPositions[step]);
+            std::stringstream mcStepPositionName;
+            mcStepPositionName << "MCStepPosition" << step;
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable(mcStepPositionName.str(), position));
+        }
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable("NumberOfMCStepMomentas", nMCStepMomentas));
```
PandoraPFA, line 285: Prefer `NumberOfMCStepMomenta` here and throughout all related usage. Apologies if you have gigabytes of XML files with this name persisted within!

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -27,7 +27,9 @@ namespace lar_content
 class LArMCParticleParameters : public object_creation::MCParticle::Parameters
 {
 public:
-    pandora::InputInt   m_nuanceCode;               ///< The nuance code
+    pandora::InputInt                          m_nuanceCode;        ///< The nuance code
+    std::vector<pandora::InputCartesianVector> m_mcStepPositions;   ///< The positions of the geant4 steps
```
PandoraPFA, line 31: I added a +1 to @AndyChappell's related comments, but expand within this review. Basically, we prefer pandora::InputTypes (e.g. pandora::InputCartesianVector) in the provision of input parameters (although there is an overhead) because this automatically guards against uninitialised values and INFs and NaNs. Within the target, initialised object, though, prefer the actual raw types (e.g. pandora::CartesianVector).

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -137,15 +174,57 @@ inline pandora::StatusCode LArMCParticleFactory::Read(Parameters &parameters, pa
...
+
     if (pandora::BINARY == fileReader.GetFileType())
     {
         pandora::BinaryFileReader &binaryFileReader(dynamic_cast<pandora::BinaryFileReader&>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nuanceCode));
+
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(nMCStepPositions));
+        for (int step = 0; step < nMCStepPositions; ++step)
+        {
+            pandora::CartesianVector mcStepPosition(0.0, 0.0, 0.0);
```
PandoraPFA, line 191: Usually use `0.f` to make explicit that it's a single floating point literal that we want to assign to each element. Here and below.

PandoraPFA (commented):
Thanks very much @CrossR - I just added in a few thoughts, and think that @AndyChappell has already flagged up the key issues. I think it's correct to say that this should allow the Pandora app to choose to include/exclude the additional trajectory points? I think it should be an option in e.g. LArPandora to decide whether to include these, then LArContent/Pandora can respond dynamically. Finally, might want to check the part of the MasterAlg where MCParticles are passed from a Pandora master instance, to worker instances, to copy across any trajectory points too.

PR #141, opened 2020-10-07
Update Chi2 to ensure 3D hits are contained.

CrossR:
Continuing pulling out some bits from my work to prevent a giant PR.

This is just the small change to the 3D hit chi-squared term to update it to include a containment term. If a hit is outside of the detector, it will be picked up and potentially cause it to fail the existing chi-squared cut. I think at its current strength Andy believed it would allow hits a few cms outside the detector (3cm or so), which I believe is desirable due to the fuzzy edges of these detectors.

<details><summary>Comparison Event Display</summary>
<p>

## Before

without

## After

with

</p>
</details>

(Ignore the rest of the issues in the example reconstruction....this fix is just for the containment!)


larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -78,6 +80,7 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType1, const HitTyp
     // TODO Input better uncertainties into this method (sigmaHit, sigmaFit, sigmaX)
     const CaloHit *const pCaloHit2D(protoHit.GetParentCaloHit2D());
     const HitType hitType(pCaloHit2D->GetHitType());
+    const LArTPCMap &larTPCMap(this->GetPandora().GetGeometry()->GetLArTPCMap());
```
AndyChappell, line 83: I think this could be initialised immediately prior to the loop in which it is used

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -113,6 +129,7 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType, const Cartesi
     // TODO Input better uncertainties into this method (sigmaHit, sigmaFit, sigmaX)
     const CaloHit *const pCaloHit2D(protoHit.GetParentCaloHit2D());
     const double sigmaFit(LArGeometryHelper::GetSigmaUVW(this->GetPandora()));
+    const LArTPCMap &larTPCMap(this->GetPandora().GetGeometry()->GetLArTPCMap());
```
AndyChappell, line 132: As above, would move this immediately before its usage loop

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -122,11 +139,31 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType, const Cartesi
...
+
+    for (const LArTPCMap::value_type &mapEntry : larTPCMap)
+    {
+        const LArTPC* currentTPC = mapEntry.second;
+        // Check if the given hit is contained in the detector.
+        // If it is outside, then we get a positive value which
+        // will set the distanceToEdge value.
+        //
+        // We want to know the largest distance outside of the detector,
+        // so that we can make these 3D hits less favourable.
```
AndyChappell, line 154: I like having the comments here, but Pandora's standard line length is 140 chars, so would take advantage of the extra width to compress the comments a bit

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -122,11 +139,31 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType, const Cartesi
...
+        distanceToEdge = std::max(distanceToEdge, (currentTPC->GetCenterY() - 0.5f * currentTPC->GetWidthY()) - bestY);
+        distanceToEdge = std::max(distanceToEdge, bestY - (currentTPC->GetCenterY() + 0.5f * currentTPC->GetWidthY()));
+        distanceToEdge = std::max(distanceToEdge, (currentTPC->GetCenterZ() - 0.5f * currentTPC->GetWidthZ()) - bestZ);
+        distanceToEdge = std::max(distanceToEdge, bestZ - (currentTPC->GetCenterZ() + 0.5f * currentTPC->GetWidthZ()));
```
AndyChappell, line 158: This is probably worth refactoring out to a separate function so that the duplicate case above can also benefit from the accompanying comment.

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -122,11 +139,31 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType, const Cartesi
...
+    }
+
     // ATTN Replace chi2 from LArGeometryHelper for consistency with three-view treatment (purely a measure of delta x)
     const double deltaX(pCaloHit2D->GetPositionVector().GetX() - fitPosition.GetX());
     const double chi2X((deltaX * deltaX) / m_sigmaX2);
+    const double chi2YZ((distanceToEdge * distanceToEdge) / m_sigmaYZ2);
 
-    protoHit.SetPosition3D(position3D, chi2X);
+    protoHit.SetPosition3D(position3D, chi2 + chi2X + chi2YZ);
```
AndyChappell, line 166: Are we happy that the ATTN comment here is no longer applicable with the inclusion of YZ? If it's no longer relevant, perhaps that comment can be removed. If there is still some relevance here, also need to consider the equivalent setting of the overall chi squared value in the function above as well.

AndyChappell (commented):
As with the previous PR, mainly minor comments and one clarification relating to the final chi square calculation, otherwise this looks good to me and can be merged with the related PRs once the final one has been reviewed/approved.

AndyChappell (commented):
Updates look good.

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -72,6 +74,30 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType1, const HitTyp
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+double HitCreationBaseTool::GetDistanceToDetectorEdge(const LArTPCMap &larTPCMap, const CartesianVector &position3D) const
+{
+    double distanceToEdge = 0;
```
absolution1, line 79: Initialise instead of assign

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -72,6 +74,30 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType1, const HitTyp
...
+    const double bestY(position3D.GetY());
+    const double bestZ(position3D.GetZ());
+
+    for (const LArTPCMap::value_type &mapEntry : larTPCMap)
+    {
+        const LArTPC* currentTPC = mapEntry.second;
```
absolution1, line 85: initialise instead of assign

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -72,6 +74,30 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType1, const HitTyp
...
+        // Check if the given hit is contained in the detector. If it is outside, then we get a positive value which
+        // will set the distanceToEdge value.
+        //
+        // We want to know the largest distance outside of the detector, so that we can make these 3D hits less favourable.
+        distanceToEdge = std::max(distanceToEdge, (currentTPC->GetCenterY() - 0.5f * currentTPC->GetWidthY()) - bestY);
+        distanceToEdge = std::max(distanceToEdge, bestY - (currentTPC->GetCenterY() + 0.5f * currentTPC->GetWidthY()));
+        distanceToEdge = std::max(distanceToEdge, (currentTPC->GetCenterZ() - 0.5f * currentTPC->GetWidthZ()) - bestZ);
+        distanceToEdge = std::max(distanceToEdge, bestZ - (currentTPC->GetCenterZ() + 0.5f * currentTPC->GetWidthZ()));
```
absolution1, line 93: I think there are implicit conversions between double and floats here?

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -97,11 +123,15 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType1, const HitTyp
     this->GetPandora().GetPlugins()->GetLArTransformationPlugin()->GetMinChiSquaredYZ(u, v, w, sigmaU, sigmaV, sigmaW, bestY, bestZ, chi2);
     position3D.SetValues(pCaloHit2D->GetPositionVector().GetX(), static_cast<float>(bestY), static_cast<float>(bestZ));
 
+    const LArTPCMap &larTPCMap(this->GetPandora().GetGeometry()->GetLArTPCMap());
+    double distanceToEdge = this->GetDistanceToDetectorEdge(larTPCMap, position3D);
```
absolution1, line 127: initialise instead of assign

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -122,11 +152,15 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType, const Cartesi
     LArGeometryHelper::MergeTwoPositions3D(this->GetPandora(), pCaloHit2D->GetHitType(), hitType, pCaloHit2D->GetPositionVector(),
         fitPosition, position3D, chi2);
 
+    const LArTPCMap &larTPCMap(this->GetPandora().GetGeometry()->GetLArTPCMap());
+    double distanceToEdge = this->GetDistanceToDetectorEdge(larTPCMap, position3D);
```
absolution1, line 156: initialise instead of assign

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
```diff
@@ -135,18 +169,28 @@ void HitCreationBaseTool::GetBestPosition3D(const HitType hitType, const Cartesi
...
     double sigmaX(std::sqrt(m_sigmaX2));
+    double sigmaYZ(std::sqrt(m_sigmaYZ2));
 
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
         "SigmaX", sigmaX));
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "SigmaYZ", sigmaYZ));
 
     m_sigmaX2 = sigmaX * sigmaX;
+    m_sigmaYZ2 = sigmaYZ * sigmaYZ;
```
absolution1, line 181: This section confuses me quite a bit.
 - `sigmaX` is constructed from a data member (`m_sigmaX2`) that is constructed in the initialised list.
 - `SigmaX` is then read in from an XML parameter.
 - `m_sigmaX2` is then set equal to the square of `sigmaX`

Couldn't `sigmaX` be the data member or could `sigmaX2` be read in from XML?
I realise the code you've added didn't set the precedent here but I feel like this should be discussed...

absolution1 (approved):
Looks nice 🎆 
I've added minimal comments + a discussion question

larpandoracontent/LArThreeDReco/LArHitCreation/HitCreationBaseTool.cc
AndyChappell: I think there's a case for adding an m_sigmaX (etc) member variable to clean this up (I think we're stuck with the SigmaX XML parameter given existing usage). Doing it this way allows us to avoid the ugly sqrt before checking to see if an optional parameter has been provided and then recalculating the square from the input value. This was presumably done on efficiency grounds previously. So our options here would be to continue to have a squared member variable alongside a new m_sigmaX member, or dispense with the "efficiency" member variable, given that we actually have examples of products that could follow the same practise, but don't, elsewhere in this file.

PR #143, opened 2020-10-09
3D Spacepoint Improvements

CrossR:
Okay, this is the actual bulk of the work.

I've got it split out from the other 2 PRs, and streamlined the history (its not the most truthful account of the order of how stuff was added, but otherwise this would be a 250 commit PR or something).

There is quite a few places where I've still got in progress `TODO`s, but almost all of them are for trivial things. Most of them are I need to check the magic number makes sense and or tune it and decide if it benefits from being user tunable so its needs a config option. The other `TODO` is usually just a "Maybe early return here", which I just need to check.

Since neither of those are massive show stoppers, and there is large parts of this that are standalone new files, it makes sense to get the ball rolling since I expect this will take a while.

If I've pre-empted a question, I'll stick a comment with why I made a change there. Alternatively, if I know something is dodgy, I'll stick a comment there so you know its an area I'm working on still.

There will also need to be some changes to the file structure. I basically put them the first place I think made a bit of sense, I didn't want to put too much work into deciding on a place, since I figured someone (John/Andy/Dom) would have a better opinion of that.


larpandoracontent/LArHelpers/LArRandomHelper.cc
```diff
@@ -0,0 +1,21 @@
...
+ *
+ *  $Log: $
+ */
+
+#include <random>
+
+namespace lar_content
+{
+
+int GetIntsInRange(const int low, const int high, std::mt19937 &eng)
```
CrossR, line 14: Header for this should explain why I had to write my own, rather than use the built in.

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -286,13 +507,14 @@ void ThreeDHitCreationAlgorithm::RefineHitPositions(const ThreeDSlidingFitResult
...
-            throw StatusCodeException(STATUS_CODE_FAILURE);
+            // std::cout << "ThreeDHitCreationAlgorithm::IterativeTreatment - Unexpected number of trajectory samples" << std::endl;
+            // throw StatusCodeException(STATUS_CODE_FAILURE);
+            continue; // TODO: Fix this, check why it happens and if it is specific to the changes I've made.
         }
 
         double bestY(std::numeric_limits<double>::max()), bestZ(std::numeric_limits<double>::max());
         PandoraContentApi::GetPlugins(*this)->GetLArTransformationPlugin()->GetMinChiSquaredYZ(u, v, w, sigmaU, sigmaV, sigmaW, uFit, vFit, wFit, sigma3DFit, bestY, bestZ, chi2);
-        position3D.SetValues(pCaloHit2D->GetPositionVector().GetX(), static_cast<float>(bestY), static_cast<float>(bestZ));
+        position3D.SetValues(protoHit.GetPosition3D().GetX(), static_cast<float>(bestY), static_cast<float>(bestZ));
```
CrossR, line 517: Maybe a slightly odd/worrying change, since it changes the underlying 3D reco, rather than just being my own standalone stuff.

However, for all non-interpolated spacepoints, the calo hit and the proto hit have the exact same X coordinate, so this change should not change anything for non-interpolated spacepoints.

For an interpolated spacepoint, its entirely possible for the calo hit and the proto hit to have different X coordinates....the proto hit is interpolated, so it can have a slightly different X. Normally, not an issue, but when running John's iterative treatment on interpolated spacepoints, without this change every interpolated spacepoint would reset its X coord to some random other X value it didn't start with, causing all the interpolated points to end up in odd positions.

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -286,13 +507,14 @@ void ThreeDHitCreationAlgorithm::RefineHitPositions(const ThreeDSlidingFitResult
         }
         else
         {
-            std::cout << "ThreeDHitCreationAlgorithm::IterativeTreatment - Unexpected number of trajectory samples" << std::endl;
-            throw StatusCodeException(STATUS_CODE_FAILURE);
+            // std::cout << "ThreeDHitCreationAlgorithm::IterativeTreatment - Unexpected number of trajectory samples" << std::endl;
+            // throw StatusCodeException(STATUS_CODE_FAILURE);
+            continue; // TODO: Fix this, check why it happens and if it is specific to the changes I've made.
```
CrossR, line 512: Looking into, as mentioned in the TODO...no idea if seeing this warning is my fault (I can't see how, I always give a trajectory sample). But it was getting in the way of producing metrics consistently by crashing out too early, so I dropped the error for now. I'll verify if its my doing, or just an oddity I was seeing with empty events that should be expected.

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.h
```diff
@@ -328,7 +423,7 @@ inline ThreeDHitCreationAlgorithm::ProtoHit::ProtoHit(const pandora::CaloHit *co
 
 inline const pandora::CaloHit *ThreeDHitCreationAlgorithm::ProtoHit::GetParentCaloHit2D() const
 {
-    return m_pParentCaloHit2D;
+    return m_pParentCaloHit2D; // TODO: I've now basically changed this by allowing nullptr. Should check its usage.
 }
```
CrossR, line 427: Will check if the default empty constructor is still needed...I can't remember which part needed this, and if I can remove it again which would be preferable.

larpandoracontent/LArUtility/RANSAC/AbstractModel.h
```diff
@@ -0,0 +1,42 @@
+// Code is from https://github.com/drsrinathsridhar/GRANSAC
+// Original license file can be found in larpandoracontent/LArUtility/RANSAC/LICENSE.
```
CrossR, line 2: Location of these files is fairly random, so input welcomed.

I've included the original license since I did start off with some MIT licensed code...but its been pulled apart and been rewritten a few times, so 🤷 .

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
```diff
@@ -0,0 +1,115 @@
+/**
+ *  @file   larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
+ *
+ *  @brief  Implementation file for the PlaneModel, to be used in RANSAC.
+ *
+ *  $Log: $
+ */
+
```
CrossR, line 8: Probably doesn't make sense to be with RANSAC? Not 100%, its a RANSAC model, but its also specific to 3D hits since it uses protohits.

larpandoracontent/LArHelpers/LArGeometryHelper.h
```diff
@@ -223,6 +223,24 @@ class LArGeometryHelper
     static bool IsInGap3D(const pandora::Pandora &pandora, const pandora::CartesianVector &testPoint3D, const pandora::HitType hitType,
         const float gapTolerance = 0.f);
 
+    /**
+     *  @brief  If a 3D test point lies in a registered gap, return a point
+     *          for the opposite edge of the gap. This in ran recursively to
+     *          deal with cases where a gap can be followed by other gaps, such
+     *          to the furthest edge of any chained gaps. Opposite side is
+     *          determined based on the given direction.
```
AndyChappell, line 231: Pandora standard line length is 140 characters

larpandoracontent/LArHelpers/LArGeometryHelper.h
```diff
@@ -223,6 +223,24 @@ class LArGeometryHelper
...
+     *
+     *  @param  pandora the associated pandora instance
+     *  @param  testPoint the test point
+     *  @param  testDirection the test direction, to look that way for a gap.
+     *  @param  gapTolerance the gap tolerance
+     *  @param  recurseLimit the maximum amount of gaps to consider, if multiple are present.
+     *
+     *  @return CartesianVector
```
AndyChappell, line 239: Now returning a float

larpandoracontent/LArHelpers/LArRandomHelper.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <random>
+
+namespace lar_content
+{
+
+/**
+ *  @brief  Specific implementation of std::uniform_int_distribution.
+ *          The default can vary between compilers, which would lead
+ *          to inconsistent randomness between gcc and clang.
```
AndyChappell, line 20: Line length

larpandoracontent/LArHelpers/LArRandomHelper.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ *  @param  low  Lowest number in range to pick.
+ *  @param  high Highest number in range to pick.
+ *  @param  eng The MT19937 to give a random number to scale.
+ */
```
AndyChappell, line 25: Needs description of return value

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.h
```diff
@@ -0,0 +1,275 @@
+/**
```
AndyChappell, line 1: General note on comments. Prefer @return comments to describe what the return value is conceptually, rather than its type.

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.h
```diff
@@ -0,0 +1,275 @@
...
+    virtual pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+
+    /**
+     *  @brief  Given a RANSAC model, run over it and produce the full 3D model
+     *  by extending and improving the initial model.
+     *
+     *  @param  ransac          The complete RANSAC object, containing both the models.
+     *  @param  run             Enum describing which run to consider (Best or Second).
+     *  @param  hitsToUse       The hits to be used in the fit extending.
+     *  @param  protoHitVector  Vector to store the final complete model in.
```
AndyChappell, line 145: Argument name mismatch

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.h
```diff
@@ -0,0 +1,275 @@
...
+     */
+    int RunOverRANSACOutput(RANSAC<PlaneModel, 3> &ransac, RANSACResult run, RANSACHitVector &hitsToUse, RANSACHitVector &finalHits);
+
+
```
AndyChappell, line 149: Excess whitespace

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.h
```diff
@@ -87,13 +100,27 @@ class ThreeDHitCreationAlgorithm : public pandora::Algorithm
          */
         const pandora::CaloHit *GetParentCaloHit2D() const;
 
+        /**
+         *  @brief  Whether the proto hit has been initialised
+         *
+         *  @return boolean
```
AndyChappell, line 106: Usual note about return value descriptions here and elsewhere

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.h
```diff
@@ -145,22 +172,37 @@ class ThreeDHitCreationAlgorithm : public pandora::Algorithm
          */
         void SetPosition3D(const pandora::CartesianVector &position3D, const double chi2);
 
+        /**
+         *  @brief  Set the protohit as an interpolated hit or not.
+         *
+         *  @param  if the hit is interpolated
```
AndyChappell, line 178: Missing the argument name

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.h
```diff
@@ -145,22 +172,37 @@ class ThreeDHitCreationAlgorithm : public pandora::Algorithm
...
+
         /**
          *  @brief  Add a trajectory sample
          *
          *  @param  the trajectory sample
          */
         void AddTrajectorySample(const TrajectorySample &trajectorySample);
 
+        /**
+         * @brief  Equality operator for a ProtoHit, compares position and parent hit.
```
AndyChappell, line 190: Missing parameter description

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.h
```diff
@@ -197,6 +239,43 @@ class ThreeDHitCreationAlgorithm : public pandora::Algorithm
...
+     *  @param  protoHitVector An empty protoHitVector, to be filled with the current state of the 3D hit construction.
+     */
+    void ConsolidatedMethod(const pandora::ParticleFlowObject *const pPfo, ProtoHitVectorMap &protoHitVectorMap,
+            ProtoHitVector &protoHitVector);
+
+    /**
+     *  @brief  Project a ProtoHit into the given view.
+     *
+     *  @param  hit The ProtoHit to project.
+     *  @param  view The view to project the hit into.
```
AndyChappell, line 257: Missing argument description

larpandoracontent/LArUtility/RANSAC/AbstractModel.h
```diff
@@ -0,0 +1,42 @@
...
+
```
AndyChappell, line 3: Modify comments to doxygen style

larpandoracontent/LArUtility/RANSAC/PlaneModel.h
```diff
@@ -0,0 +1,56 @@
...
+
+    PlaneModel(ParameterVector inputParams) { Initialize(inputParams); };
+    virtual ~PlaneModel() {};
+
+    pandora::CartesianVector GetDirection() const;
+    pandora::CartesianVector GetOrigin() const;
+
+    virtual void Initialize(const ParameterVector &inputParams) override;
+    virtual std::pair<double, ParameterVector> Evaluate(const ParameterVector &paramsToEval, double threshold) override;
+    void operator=(PlaneModel &other);
```
AndyChappell, line 51: Add doxygen comments

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -422,6 +422,58 @@ bool LArGeometryHelper::IsInGap3D(const Pandora &pandora, const CartesianVector
...
+        //       If the Z is used for detector gaps, can cause infs.
+        const float gapWidth = std::fabs(lineGap->GetLineStartX() - lineGap->GetLineEndX());
+
+        const float distanceToStart = std::fabs(testPoint.GetX() - lineGap->GetLineStartX());
+        const float distanceToEnd = std::fabs(testPoint.GetX() - lineGap->GetLineEndX());
+        const float middleOfGap = std::fabs((lineGap->GetLineStartX() + lineGap->GetLineEndX()) / 2);
+        const CartesianVector gapPoint(middleOfGap, 0, testPoint.GetZ());
+
+        const float distanceToGap = std::min(distanceToStart, distanceToEnd);
+        const float totalGapSize = distanceToGap + gapWidth;
```
AndyChappell, line 450: What's actually being computed here? This seems to be taking the minimum distance between a point inside the gap and the gap extrema, and adding it to the size of the gap, which seems to have some degree of double-counting.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -422,6 +422,58 @@ bool LArGeometryHelper::IsInGap3D(const Pandora &pandora, const CartesianVector
...
+
+        const CartesianVector pointToGap(gapPoint - testPoint);
+
+        if (!(testDirection.GetX() < 0) == (pointToGap.GetX() < 0))
+            continue;
+
+        const float cosTheta = (pointToGap).GetCosOpeningAngle(testDirection);
+        const float distanceToProject = totalGapSize / cosTheta;
+
+        if (distanceToProject < bestGapSize)
```
AndyChappell, line 460: I think we want an absolute value comparison here as well

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
```diff
@@ -0,0 +1,115 @@
...
+{
+    const auto currentPoint = std::dynamic_pointer_cast<RANSACHit>(param);
+    if(currentPoint == nullptr)
+        throw std::runtime_error("PlaneModel::ComputeDistanceMeasure() - Passed parameter are not of type RANSACHit.");
+
+    const RANSACHit hit = *currentPoint;
+    const Eigen::Vector3f currentPos = hit.GetVector() - m_origin;
+
+    const Eigen::Vector3f b = currentPos.dot(m_direction) * m_direction;
+    const double distance = (currentPos - b).norm();
```
AndyChappell, line 32: Prefer `()` for initialisation over `=`.

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
```diff
@@ -0,0 +1,115 @@
...
+
+    for (unsigned int i = 0; i < inputParams.size(); ++i)
+    {
+        auto currentPoint = *std::dynamic_pointer_cast<RANSACHit>(inputParams[i]);
+        Eigen::Vector3f shiftedPoint = Eigen::Vector3f(currentPoint.GetVector() - m_origin);
+        m.row(i) = shiftedPoint;
+    }
+
+    const Eigen::JacobiSVD<Eigen::MatrixXf> svd(m, Eigen::ComputeThinV);
+    m_direction = svd.matrixV().col(0); // TODO: Check this, returns matrix not vector.
```
AndyChappell, line 81: Can this TODO be removed now?

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
```diff
@@ -0,0 +1,115 @@
...
+    const float totalParams = paramsToEval.size();
+
+    for(auto& param : paramsToEval)
+    {
+        if(ComputeDistanceMeasure(param) < threshold)
+            inliers.push_back(param);
+    }
+
+    // TODO: This could actually take into account the tool -> i.e. weight against the iffy tools.
+    const double inlierFraction = inliers.size() / totalParams;
```
AndyChappell, line 98: Want divide by zero protection here

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+namespace lar_content
+{
+
+template <class T, int t_numParams>
+class RANSAC
+{
+
+public:
+
+    RANSAC(double threshold, int numIterations = 1000)
```
AndyChappell, line 30: Needs comments

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+    {
+        this->Reset();
+
+        m_threshold = threshold;
+        m_numIterations = numIterations;
```
AndyChappell, line 35: Move to initialisation list

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+     */
+    void Reset() { m_data.clear(); };
+
+    virtual ~RANSAC() {};
+
+    std::shared_ptr<T> GetBestModel() { return m_bestModel; };
+    ParameterVector& GetBestInliers() { return m_bestInliers; };
+
+    std::shared_ptr<T> GetSecondBestModel() { return m_secondBestModel; };
+    ParameterVector& GetSecondBestInliers() { return m_secondBestInliers; };
```
AndyChappell, line 49: Comments

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+    /**
+     *  @brief  Given a vector of data, get the best and second best model for it.
+     *
+     *  @param  data The data to use for this RANSAC run.
+     */
```
AndyChappell, line 57: Describe return value

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+
+        m_data = data;
+        this->GenerateSamples();
+
+        std::vector<double> inlierFrac(m_numIterations, 0.0);
+        std::vector<ParameterVector> inliers(m_numIterations);
+        std::vector<std::shared_ptr<T>> sampledModels(m_numIterations);
+
+        RANSAC::CheckModel(inlierFrac, inliers, sampledModels);
+        double bestModelScore = -1;
```
AndyChappell, line 71: Prefer () over =

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+
+            if (inlierFrac[i] > bestModelScore)
+            {
+                bestModelScore = inlierFrac[i];
+                m_bestModel = sampledModels[i];
+                m_bestInliers = inliers[i];
+            }
+        }
+
+        double secondModelScore = -1;
```
AndyChappell, line 86: Prefer () over =

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
+/**
+ *  @file   larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.h
```
AndyChappell, line 2: .cc

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+void LArRANSACMethod::Run(RANSACHitVector &consistentHits, ProtoHitVector &protoHitVector)
+{
+    if (consistentHits.size() < 3)
+        return; // TODO: Here we should default to the old behaviour.
+
+    ParameterVector candidatePoints;
+    this->GetCandidatePoints(consistentHits, candidatePoints);
+
+    const float RANSAC_THRESHOLD = 2.5;
+    const int RANSAC_ITERS = 100; // TODO: Should either be dynamic, or a config option.
```
AndyChappell, line 40: Prefer () over = here and elsewhere

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+            consistentHits, primaryResult);
+
+    RANSACHitVector secondaryResult;
+    const int secondModelCount = this->RunOverRANSACOutput(estimator, RANSACResult::Second,
+            consistentHits, secondaryResult);
+
+    float primaryFavoured = 0;
+    float secondaryFavoured = 0;
+
+    for (auto hit : primaryResult) {
```
AndyChappell, line 55: New line for {

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+        if (hit.IsFavourable())
+            ++primaryFavoured;
+    }
+
+    for (auto hit : secondaryResult) {
```
AndyChappell, line 60: New line for {

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void LArRANSACMethod::GetCandidatePoints(RANSACHitVector &allHits, ParameterVector &candidatePoints)
+{
+    if (allHits.size() < 1000)
+    {
+       for (auto hit : allHits)
+           candidatePoints.push_back(std::make_shared<RANSACHit>(hit));
+
+       return;
```
AndyChappell, line 87: Indentation isn't quite right here

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+
+        candidatePoints.push_back(std::make_shared<RANSACHit>(allHits[i]));
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+int LArRANSACMethod::RunOverRANSACOutput(RANSAC<PlaneModel, 3> &ransac, RANSACResult run,
+        RANSACHitVector &hitsToUse, RANSACHitVector &finalHits
+)
```
AndyChappell, line 109: Excess indentation and move closing parenthesis to end of line

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+{
+    std::map<const CaloHit*, RANSACHit> inlyingHitMap;
+    const float RANSAC_THRESHOLD = 2.5; // TODO: Consolidate to config option.
+    const float EXTEND_THRESHOLD = 6.25; // TODO: Config.
```
AndyChappell, line 113: Outstanding TODOs

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+
+    const ParameterVector currentInliers = run == RANSACResult::Best ? ransac.GetBestInliers() : ransac.GetSecondBestInliers();
+
+    if (currentInliers.size() == 0 || hitsToUse.size() == 0)
+    {
+        for (auto const& caloProtoPair : inlyingHitMap)
```
AndyChappell, line 119: Where is this map populated?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+        return 0;
+    }
+
+    const auto bestModel = ransac.GetBestModel();
+    auto secondModel = ransac.GetSecondBestModel();
+
+    // ATTN: The second model can technically be NULL, so set to first model in
+    //       that case, as that won't cause any issues.
+    if (!secondModel)
+        secondModel = bestModel;
```
AndyChappell, line 131: Is bestModel guaranteed to be non-null here?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+
+    std::sort(sortedHits.begin(), sortedHits.end(), sortByModelDisplacement);
+
+    // TODO: If RANSAC was good enough (i.e. it got everything) skip the sorting and similar.
+    std::list<RANSACHit> currentPoints3D;
+
+    for (auto hit : sortedHits)
+        currentPoints3D.push_back(hit);
+
+    const int FIT_ITERATIONS = 1000; // TODO: Config?
```
AndyChappell, line 191: Any decision on the TODO?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+    int coherentHitCount = 0;
+
+    // INFO: For each iteration, try and extend the fit. Fitting is run
+    //       multiple times per iteration, stopping after the first iteration
+    //       that adds hits.
+    //
+    //       If hits are added, add them to the total hit map and make the new
+    //       hits available for further iterations of the fit extending.
+    //
+    //       When no hits remain, repeat from other end of fit, then stop.
```
AndyChappell, line 209: Compress to fewer lines

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+    for (unsigned int iter = 0; iter < FIT_ITERATIONS; ++iter)
+    {
+        hitsToAdd.clear();
+
+        for (float fits = 1; fits < 4.0; ++fits)
+        {
+            LArRANSACMethod::ExtendFit(
+                nextHits, hitsToUseForFit, hitsToAdd,
+                (EXTEND_THRESHOLD * fits), extendDirection
+            );
```
AndyChappell, line 219: Probably fits on one line

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+
+        for (auto hit : hitsToAdd)
+        {
+            const bool hitAdded = LArRANSACMethod::AddToHitMap(hit, inlyingHitMap);
+
+            if (hitAdded)
+                hitsToUseForFit.push_back(hit);
+
+            // TODO: Do we always want to add the hits? Or only at start/end, not the middle?
+            //       Should this be apart of the other check above?
```
AndyChappell, line 233: Any conclusion on this?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+            if (currentPoints3D.size() == 0)
+                currentPoints3D.push_back(hit);
+        }
+
+        const bool continueFitting =
+            LArRANSACMethod::GetHitsForFit(currentPoints3D, hitsToUseForFit,
+            hitsToAdd.size(), smallIterCount);
```
AndyChappell, line 240: Probably fits on two lines

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+        if (!continueFitting && extendDirection == ExtendDirection::Forward)
+        {
+            extendDirection = ExtendDirection::Backward;
+            hitsToUseForFit.clear();
+            currentPoints3D.clear();
+            smallIterCount = 0;
+
+            auto it = sortedHits.begin();
+            // TODO: Randomly chosen "at least 5 fits worth", evaluate.
+            while (currentPoints3D.size() < (5 * 80) && it != sortedHits.end())
```
AndyChappell, line 255: What do the constant values here represent?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool LArRANSACMethod::GetHitsForFit(
+        std::list<RANSACHit> &currentPoints3D,
+        RANSACHitVector &hitsToUseForFit,
+        const int addedHitCount,
+        int smallAdditionCount
+)
```
AndyChappell, line 284: Compress to fewer lines

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+{
+    const int HITS_TO_KEEP = 80; // TODO: Config and consolidate (reverse bit).
+    const int FINISHED_THRESHOLD = 10; // TODO: Config
```
AndyChappell, line 287: Outstanding TODOs

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+                hitsToUseForFit.erase(hitsToUseForFit.begin());
+
+            ++i;
+        }
+    }
+    else if (addedHitCount > 0)
+    {
+        auto it = hitsToUseForFit.begin();
+        while (hitsToUseForFit.size() > HITS_TO_KEEP)
+            it = hitsToUseForFit.erase(it);
```
AndyChappell, line 319: Given you know how many elements need to be erased, I would think you could erase a set of hits in one call, rather than a loop?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+
+    return false;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool hitIsCloseToEnd(RANSACHit &hit, const CartesianVector &fitEnd,
+        const CartesianVector &fitDirection, float threshold)
```
AndyChappell, line 361: One line

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void LArRANSACMethod::ExtendFit(
+    std::list<RANSACHit> &hitsToTestAgainst,
+    RANSACHitVector &hitsToUseForFit,
+    std::vector<RANSACHit> &hitsToAdd,
+    const float distanceToFitThreshold,
+    const ExtendDirection extendDirection
+)
```
AndyChappell, line 403: Compress to fewer lines

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+{
+    if (hitsToUseForFit.size() == 0)
+        return;
+
+    float distanceToEndThreshold = 10; // TODO: Config?
```
AndyChappell, line 408: Do you want this to be configurable?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+    const unsigned int layerWindow(100);
+    const float pitch(LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+    const ThreeDSlidingFitResult slidingFit(&fitPoints, layerWindow, pitch);
+
+    CartesianVector fitDirection = slidingFit.GetGlobalMaxLayerDirection();
+    CartesianVector fitEnd = slidingFit.GetGlobalMaxLayerPosition();
+
+    if (extendDirection == ExtendDirection::Backward)
+    {
+        // TODO: This seems a bit iffy...does this work as I expect?
```
AndyChappell, line 423: Has this be confirmed?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+        fitDirection = slidingFit.GetGlobalMinLayerDirection();
+        fitEnd = slidingFit.GetGlobalMinLayerPosition();
+        fitDirection = fitDirection * -1.0;
+    }
+
+    // INFO: Check if there is a detector gap, and if there is, extend the fit over it.
+    auto distanceToProjectOverGap = LArGeometryHelper::ProjectAcrossGap3D(this->GetPandora(), fitEnd, fitDirection, 2.0, 10);
+    if (distanceToProjectOverGap > 0)
+        distanceToEndThreshold += std::min(distanceToProjectOverGap, 50.0f);
```
AndyChappell, line 432: Has any tuning been done on this cap? If not, what's the general thinking behind 50?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+
+    // ATTN: This is done in 3 stages to split up the 3 different qualities of
+    //       hits:
```
AndyChappell, line 435: One line

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,490 @@
...
+        hitDisplacement(hit, fitEnd, fitDirection);
+        if (hit.GetDisplacement() < threshold && hit.IsFavourable())
+            return true;
+        return false;
+    };
+    auto unfavourableTest = [] (RANSACHit &hit, float threshold) {
+        if (hit.GetDisplacement() < threshold)
+            return true;
+        return false;
+    };
```
AndyChappell, line 455: Can make these a little more concise by just returning the result of the conditional directly.

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -78,11 +84,44 @@ StatusCode ThreeDHitCreationAlgorithm::Run()
                 break;
 
             pHitCreationTool->Run(this, pPfo, remainingTwoDHits, protoHitVector);
+
+            if (m_useRANSACMethod && LArPfoHelper::IsTrack(pPfo))
+            {
+                // TODO: Replace 10 with a configuration controlled number.
```
AndyChappell, line 90: TODO

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void ThreeDHitCreationAlgorithm::Project3DHit(const ProtoHit &hit, const HitType view, ProtoHit &projectedHit)
+{
+    projectedHit.SetPosition3D(
+        LArGeometryHelper::ProjectPosition(this->GetPandora(), hit.GetPosition3D(), view),
+        hit.GetChi2()
+    );
```
AndyChappell, line 222: Compress

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+
+    std::sort(first.begin(), first.end(), compareFunction);
+    std::sort(second.begin(), second.end(), compareFunction);
+
+    std::set_intersection(
+        first.begin(), first.end(),
+        second.begin(), second.end(),
+        std::inserter(result, result.begin()),
+        compareFunction
+    );
```
AndyChappell, line 241: Compress

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ThreeDHitCreationAlgorithm::ConsolidatedMethod(const ParticleFlowObject *const pPfo, ProtoHitVectorMap &allProtoHitVectors,
+        ProtoHitVector &protoHitVector)
```
AndyChappell, line 247: Excess indentation

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+{
+    if (allProtoHitVectors.size() == 0)
+        return;
+
+    const float DISTANCE_THRESHOLD = 0.05; // TODO: Move to config option.
```
AndyChappell, line 252: TODO

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+    const std::vector<HitType> views = {TPC_VIEW_U, TPC_VIEW_V, TPC_VIEW_W};
+
+    std::map<HitType, RANSACHitVector> goodHits;
+
```
AndyChappell, line 256: Remove new lines here

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+    const std::vector<std::string> toolsToAvoid = {"Tool0039", "Tool0043"}; // TODO: Config option?
```
AndyChappell, line 257: I think this will need to be configurable. Presumably will also need to think about mapping tool names onto the ids, as I assume those numbers are necessarily stable.

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+        const auto avoidedIt = std::find(toolsToAvoid.begin(), toolsToAvoid.end(), toolVectorPair.first);
+        const bool goodTool = avoidedIt == toolsToAvoid.end();
+
+        // INFO: Project every 3D hit into all 2D views, so how well they match
+        // can be compared. This is only really a concern for the tools which
+        // can explode, however, they do provide useful inputs so need to
+        // still be considered.
+        for (const auto &hit : toolVectorPair.second)
+        {
+            const CaloHit* twoDHit = hit.GetParentCaloHit2D();
```
AndyChappell, line 273: Technically I think this can be null, so might want to check this to avoid creating an initialised ProtoHit with a null parent calohit.

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+        return;
+
+    CaloHitVector remainingTwoDHits;
+    this->SeparateTwoDHits(pfo, protoHitVector, remainingTwoDHits);
+
+    if (remainingTwoDHits.empty())
+        return;
+
+    const float layerPitch(LArGeometryHelper::GetWireZPitch(this->GetPandora()));
+    const unsigned int layerWindow(100); // TODO: Check if this should be the same or different.
```
AndyChappell, line 321: Conclusion?

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+
+    double originalChi2(0.);
+    CartesianPointVector currentPoints3D;
+    this->ExtractResults(protoHitVector, originalChi2, currentPoints3D);
+
+    if (currentPoints3D.size() <= 1)
+        return;
+
+    const ThreeDSlidingFitResult slidingFitResult(&currentPoints3D, layerWindow, layerPitch);
+    // CartesianVector fitDirection = slidingFitResult.GetGlobalMaxLayerDirection();
```
AndyChappell, line 331: Presumably this can be deleted?

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+
+        // TODO: Tune cut off.
+        // Now we have a position for the interpolated hit, we need to make a
+        // protoHit out of it.  This includes the calculation of a chi2 value,
+        // for how good the interpolation is.
+        // const float displacement = projectedPosition.GetCrossProduct(fitDirection).GetMagnitude();
+        // const float otherDisp = projectedPosition.GetCrossProduct(projectedDirection).GetMagnitude();
+
+        // if (otherDisp > 150)
+        //     continue;
```
AndyChappell, line 357: What's the status of this block of code?

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+        ProtoHit interpolatedHit(currentCaloHit);
+
+        const CartesianVector projectedHit = LArGeometryHelper::ProjectPosition(
+                this->GetPandora(),
+                projectedPosition,
+                currentCaloHit->GetHitType()
+        );
+        const double distanceBetweenHitsSqrd = (
+                (currentCaloHit->GetPositionVector() - projectedHit).GetMagnitudeSquared()
+        );
```
AndyChappell, line 368: Compress

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+
+        const double sigmaUVW(LArGeometryHelper::GetSigmaUVW(this->GetPandora()));
+        const double sigma3DFit(sigmaUVW * m_sigma3DFitMultiplier);
+        const double interpolatedChi2 = (distanceBetweenHitsSqrd) / (sigma3DFit * sigma3DFit);
+
+        interpolatedHit.SetPosition3D(projectedPosition, interpolatedChi2);
+        interpolatedHit.SetInterpolated(true);
+        interpolatedHit.AddTrajectorySample(
+                TrajectorySample(projectedPosition, currentCaloHit->GetHitType(), sigmaUVW)
+        );
```
AndyChappell, line 378: Probably fits on one line

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +214,183 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+
+        protoHitVector.push_back(interpolatedHit);
+    }
+
+    // ATTN: If we've interpolated at least 80% of this particle, don't use it.
+    //
+    // TODO: Swap to option?
+    // TODO: This ideally would be earlier on, and wouldn't clear, but just drop the interpolated.
```
AndyChappell, line 386: TODO status?

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
AndyChappell: As a minimum I think this should be an option; if using RANSAC then use your new version, if not using RANSAC retain the existing SetValues call to avoid any possibility these values could differ in the underlying reco.

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+            const auto p1 = candidateInliers[i];
+
+            const unsigned int maxDiffSize = diff.size() + (candidateInliers.size() - i);
+
+            // ATTN: Early return.
+            if (maxDiffSize < m_secondUniqueParamCount)
+                return;
+
+            // INFO: If current parameter is in the best model, skip it.
+            if (m_bestModel->ComputeDistanceMeasure(p1) < m_threshold)
```
AndyChappell, line 155: Is m_bestModel guaranteed to be non-null here?

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,213 @@
...
+     *  @param  sampledModels Vector of the actual models that were generated and evaluated.
+     */
+    void CheckModel(std::vector<double> &inlierFrac, std::vector<ParameterVector> &inliers,
+            std::vector<std::shared_ptr<T>> &sampledModels)
+    {
+        for (unsigned int i = 0; i < m_numIterations; ++i)
+        {
+            // Evaluate the current model, so that its performance can be checked later.
+            const std::shared_ptr<T> randomModel = std::make_shared<T>(m_samples[i]);
+            const std::pair<double, ParameterVector> evalPair = randomModel->Evaluate(m_data, m_threshold);
```
AndyChappell, line 199: Confirm that randomModel was correctly allocated before dereferencing

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
```diff
@@ -0,0 +1,115 @@
...
+            throw std::runtime_error("PlaneModel - inputParams type mismatch. It is not a RANSACHit.");
+
+        totals += (*currentPoint).GetVector();
+    }
+
+    m_origin = totals / inputParams.size();
+
+    for (unsigned int i = 0; i < inputParams.size(); ++i)
+    {
+        auto currentPoint = *std::dynamic_pointer_cast<RANSACHit>(inputParams[i]);
```
AndyChappell, line 75: Check the cast worked before dereferencing

AndyChappell (commented):
Most of the comments here are stylistic - for things like () versus = initialisation I haven't identified every instance, so please have a quick pass through to identify the outstanding cases.

Beyond that, there are a few TODO statements that probably warrant addressing ahead of any merge (some are probably ok to leave as future optimisations). There are also a number of your own comments that still need to be addressed.

The critical checks/fixes relate to potential nullptr dereferencing, and another important issue to address will be tool avoidance in a stable, reproducible fashion.

Overall however, it seems like this is pretty close to being release-ready. Thanks Ryan!

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
CrossR: A few lines before this, I loop over the same data and do the same cast and check then, so it *should* be okay here. Can add if you'd like though.

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
CrossR: No tuning, it was just a value that felt right to me and Andy. 50 should be 50cm here, which felt about right for a steep track going into a gap (including a tolerance on either side, so it ends up smaller than 50 for the gap size).

CrossR:
I think I've addressed all the comments now, except a remaining TODO that I'm running a job to check.

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+        if (hit == nullptr)
+            throw std::runtime_error("Inlying hit was not of type RANSACHit");
+
+        const bool addedHit(RANSACMethodTool::AddToHitMap(*hit, inlyingHitMap));
+
+        if (addedHit)
+            sortedHits.push_back(*hit);
+    }
+
+    if (sortedHits.size() < 3 || hitsToUse.size() == 0)
```
AndyChappell, line 149: hitsToUse.empty() preferable here

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+        }
+
+        for (auto hit : hitsToAdd)
+        {
+            const bool hitAdded = RANSACMethodTool::AddToHitMap(hit, inlyingHitMap);
+
+            if (hitAdded)
+                hitsToUseForFit.push_back(hit);
+
+            if (currentPoints3D.size() == 0)
```
AndyChappell, line 216: .empty() as above

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+
+bool RANSACMethodTool::GetHitsForFit(std::list<RANSACHit> &currentPoints3D, RANSACHitVector &hitsToUseForFit, const int addedHitCount,
+    int smallAdditionCount)
+{
+    // ATTN: Four options:
+    //  Added no hits at the end: Stop.
+    //  Added small number of hits at end repeatedly: Stop.
+    //  Added small number of hits inside fit: Clear and add hits.
+    //  Added lots of hits of hits: Don't add hits, just trim vector.
+    if (addedHitCount == 0 && currentPoints3D.size() == 0)
```
AndyChappell, line 268: .empty()

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+        {
+            auto it = hitsToUseForFit.begin();
+            std::advance(it, m_fitSize);
+            hitsToUseForFit.erase(it, hitsToUseForFit.end());
+        }
+    }
+
+    // ATTN: This keeps track of the number of iterations that add only a small
+    // number of hits. Resets if a larger iteration is reached.
+    if (addedHitCount < 5 && currentPoints3D.size() == 0)
```
AndyChappell, line 302: .empty()

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+
+    hit.SetDisplacement(displacement);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void RANSACMethodTool::ExtendFit(std::list<RANSACHit> &hitsToTestAgainst, RANSACHitVector &hitsToUseForFit, std::vector<RANSACHit> &hitsToAdd,
+    const float distanceToFitThreshold, const ExtendDirection extendDirection)
+{
+    if (hitsToUseForFit.size() == 0)
```
AndyChappell, line 377: .empty()

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+
+    std::vector<std::list<RANSACHit>::iterator> hitsToCheck;
+    for (auto it = hitsToTestAgainst.begin(); it != hitsToTestAgainst.end(); ++it)
+    {
+        if (hitIsCloseToEnd((*it), fitEnd, fitDirection, distanceToEndThreshold))
+            hitsToCheck.push_back(it);
+    }
+
+    unsigned int currentTest(0);
+    while (hitsToAdd.size() == 0 && currentTest < tests.size())
```
AndyChappell, line 431: .empty()

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +217,152 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+    std::sort(second.begin(), second.end(), compareFunction);
+
+    std::set_intersection(first.begin(), first.end(), second.begin(), second.end(), std::inserter(result, result.begin()), compareFunction);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ThreeDHitCreationAlgorithm::ConsolidatedMethod(const ParticleFlowObject *const pPfo, ProtoHitVectorMap &allProtoHitVectors, ProtoHitVector &protoHitVector)
+{
+    if (allProtoHitVectors.size() == 0)
```
AndyChappell, line 243: .empty()

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -175,6 +217,152 @@ void ThreeDHitCreationAlgorithm::IterativeTreatment(ProtoHitVector &protoHitVect
...
+        return;
+
+    const std::vector<HitType> views = {TPC_VIEW_U, TPC_VIEW_V, TPC_VIEW_W};
+    std::map<HitType, RANSACHitVector> goodHits;
+
+    for (auto toolVectorPair : allProtoHitVectors)
+    {
+        if (toolVectorPair.second.size() == 0)
```
AndyChappell, line 251: .empty()

larpandoracontent/LArUtility/RANSAC/PlaneModel.cc
```diff
@@ -0,0 +1,117 @@
...
+    ParameterVector inliers;
+    const float totalParams(paramsToEval.size());
+
+    for(auto& param : paramsToEval)
+    {
+        if(ComputeDistanceMeasure(param) < threshold)
+            inliers.push_back(param);
+    }
+
+    if (inliers.size() == 0)
```
AndyChappell, line 97: .empty()

AndyChappell (commented):
Changes in response to review comments look good to me, just some minor updates relating to use of empty over size == 0. Thanks, Ryan.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
MariaBrigida, line 450: It is not clear to me if this is the best X range over which to project, in case the point is inside the gap...

larpandoracontent/LArHelpers/LArRandomHelper.h
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *  @return int An integer between the limits, inclusive.
+ */
+int GetIntsInRange(const int low, const int high, std::mt19937 &eng);
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+} // namespace lar_content
+
+#endif // #ifndef LAR_GEOMETRY_HELPER_H
```
MariaBrigida, line 34: In the comment here it should be LAR_RANDOM_HELPER_H

larpandoracontent/LArThreeDReco/LArHitCreation/ThreeDHitCreationAlgorithm.cc
```diff
@@ -78,11 +88,43 @@ StatusCode ThreeDHitCreationAlgorithm::Run()
...
+                }
+
+                this->IterativeTreatment(protoHitVector);
+                allProtoHitVectors.insert(ProtoHitVectorMap::value_type(pHitCreationTool->GetType(), protoHitVector));
+                protoHitVector.clear();
+            }
         }
 
-        if ((m_iterateTrackHits && LArPfoHelper::IsTrack(pPfo)) || (m_iterateShowerHits && LArPfoHelper::IsShower(pPfo)))
+        if (numberOfFailedAlgorithms == m_algorithmToolVector.size())
```
AndyChappell, line 110: This value does not appear to be altered anywhere.

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+    {
+        const auto hit = std::dynamic_pointer_cast<RANSACHit>(inlier);
+
+        if (hit == nullptr)
+            throw std::runtime_error("Inlying hit was not of type RANSACHit");
+
+        const bool addedHit(RANSACMethodTool::AddToHitMap(*hit, inlyingHitMap));
+
+        if (addedHit)
+            sortedHits.push_back(*hit);
```
AndyChappell, line 146: I'm having a hard time understanding the relationship between some of the hits here. If this protohit is deemed a better candidate than the existing protohit for some parent calo hit, it seems that this and any preceding protohits all get add to the list of sorted hits, which suggests we, in principle at least, could end up with multiple protohits representing the same parent calo hit in one set of sorted hits? Is that right? The final hits list seems to avoid this problem. So what'a the goal of this sorted list?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+
+    // INFO: For each iteration, try and extend the fit. Fitting is run multiple times per iteration, stopping after the first iteration
+    //       that adds hits.
+    //       If hits are added, add them to the total hit map and make the new hits available for further iterations of the fit extending.
+    //       When no hits remain, repeat from other end of fit, then stop.
+    for (unsigned int iter = 0; iter < m_fitIterations; ++iter)
+    {
+        hitsToAdd.clear();
+
+        for (float fits = 1; fits < 4.0; ++fits)
```
AndyChappell, line 201: Does this need to be a float?

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
```diff
@@ -0,0 +1,476 @@
...
+    //          - Unfavourable hits.
+    auto projectedDisplacementTest = [&slidingFit] (RANSACHit &hit, float threshold) {
+        projectedHitDisplacement(hit, slidingFit);
+        return hit.GetDisplacement() < threshold && hit.IsFavourable();
+    };
+    auto displacementTest = [&fitEnd, &fitDirection] (RANSACHit &hit, float threshold) {
+        hitDisplacement(hit, fitEnd, fitDirection);
+        return hit.GetDisplacement() < threshold && hit.IsFavourable();
+    };
+    auto unfavourableTest = [] (RANSACHit &hit, float threshold) {
```
AndyChappell, line 417: Is this the best name for this function?

AndyChappell (commented):
Aside from one part of the code that I'm struggling to understand I'm pretty happy with the overall method here. Beyond this I think the main outstanding change I would request is that all exceptions be of type StatusCodeException, as that's generally what the Pandora codebase looks for in its exception handling.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
CrossR: I've updated this now, to calculate the distance from the current point to each gap edge, and then pick the most appropriate edge based on the given direction.

I did a quick local test for the 6 cases (left / inside / right of gap, going left/right) and got the expected answers out each time (1/2 gap width when in the middle of the gap, full gap plus threshold either side and pointing towards it, 0 when pointing away from the gap).

I'll do a full scale check of this on my sample when I've implemented the other comments (as well as inside the CI) to verify everything is working as expected though.

larpandoracontent/LArHelpers/LArGeometryHelper.cc
```diff
@@ -422,6 +422,74 @@ bool LArGeometryHelper::IsInGap3D(const Pandora &pandora, const CartesianVector
...
+        const CartesianVector startPoint(lineGap->GetLineStartX(), testPoint.GetY(), testPoint.GetZ());
+        const float distanceToStart = std::fabs(testPoint.GetX() - lineGap->GetLineStartX());
+        const float angleToStart = startPoint.GetCosOpeningAngle(testDirection);
+
+        const CartesianVector endPoint(lineGap->GetLineEndX(), testPoint.GetY(), testPoint.GetZ());
+        const float distanceToEnd = std::fabs(testPoint.GetX() - lineGap->GetLineEndX());
+        const float angleToEnd = endPoint.GetCosOpeningAngle(testDirection);
+
+        // INFO: Choose the gap infront of the current point.
+        const bool shouldUseGapStart = angleToStart > angleToEnd ? true : false;
```
MariaBrigida, line 453: Why not using directly the distance from the edges, to determine whether to use the start or the end of the gap? I may be missing something, but it seems to me that angleToStart and angleToEnd do not depend on the testPoint x coordinate...

larpandoracontent/LArUtility/RANSAC/RANSAC.h
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        double secondModelScore(-1);
+
+        for (unsigned int i = 0; i < sampledModels.size(); ++i)
+        {
+            if (inlierFrac[i] == 0.0)
+                continue;
+
+            if (inlierFrac[i] > (secondModelScore * 0.9))
```
MariaBrigida, line 119: Would it make sense to put 0.9 into a variable, or do you anticipate this will never be changed?

CrossR: (Copying a comment over from Slack for context)

As for making it a variable, I wasn’t too sure. I think the set I ended up on was ones that would be somewhat easily tuned by someone, whereas I think that 0.9 requires a lot of internal knowledge about how the method works, so in reality wouldn’t be easily tuned.

(That is compared to something like “how far in front of fit do you look”, this is a lot more abstract of a variable to think about)

larpandoracontent/LArThreeDReco/LArHitCreation/RANSACMethod.cc
MariaBrigida: I second that! I think it would really help to have a few lines of comment to help clarify the flow of this function!

PR #146, opened 2020-10-16
Bug fix where no vertex exists

absolution1:
Figured it might be easier to discuss this in a PR than on slack


larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -460,8 +460,6 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
     if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
         std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
 
-    LArMvaHelper::MvaFeature vertexDistance;
```
PandoraPFA, line 463: Looks like you do want this, actually, as it's a LArMvaHelper::MvaFeature, which is a double wrapped up with a Boolean to describe whether it has been initialized, as per [here](https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArObjects/LArMvaInterface.h#L82)

At end of function, want to place one of these in the feature vector, rather than some default value.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -460,8 +460,6 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
...
-    bool error = true;
```
PandoraPFA, line 464: Instead of this error flag, can ask the LArMvaHelper::MvaFeature whether it `IsInitialized`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -477,20 +475,24 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
...
-            vertexDistance = (pVertexList->front()->GetPosition() - LArPfoHelper::GetVertex(pInputPfo)->GetPosition()).GetMagnitude();
-            error = false;
+            featureVector.push_back((pVertexList->front()->GetPosition() - LArPfoHelper::GetVertex(pInputPfo)->GetPosition()).GetMagnitude());
+            return;
         }
         catch (const StatusCodeException &) {}
     }
 
-    if (error)
+    else if (!pVertexList->empty())
```
PandoraPFA, line 484: So e.g. `if (!vertexDistance.IsInitialized() && .. && etc...)`

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -477,20 +475,24 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
...
     {
         CaloHitList threeDCaloHitList;
         LArPfoHelper::GetCaloHits(pInputPfo, TPC_3D, threeDCaloHitList);
-        vertexDistance = (pVertexList->front()->GetPosition() - (threeDCaloHitList.front())->GetPositionVector()).GetMagnitude(); // if n3dHits == 1, can't calculate vertex postion of input pfos hence ask for the position of the single 3D hit instead and set vertexDistance to be the magnitude of the difference between the interaction vertex and the hit's position.
```
PandoraPFA, line 490: Suggest lose this comment, and add a commit to correct any whitespace/indentation in the file, thank you.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -477,20 +475,24 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
...
+        if (!threeDCaloHitList.empty())
+        {
+            featureVector.push_back((pVertexList->front()->GetPosition() - (threeDCaloHitList.front())->GetPositionVector()).GetMagnitude()); // if n3dHits == 1, can't calculate vertex postion of input pfos hence ask for the position of the single 3D hit instead and set vertexDistance to be the magnitude of the difference between the interaction vertex and the hit's position.
+            return;
+        }
     }
 
-    featureVector.push_back(vertexDistance);
+    featureVector.push_back(std::numeric_limits<LArMvaHelper::MvaFeature>::lowest());
```
PandoraPFA, line 495: So an uninitialised vertexDistance should be what is wanted here, but obviously tests required.

PandoraPFA (commented):
Many thanks @absolution1 - main comment about the LArMvaHelper::MvaFeature.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -460,6 +460,7 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
     if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
         std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
 
+    LArMvaHelper::MvaFeature vertexDistance;
     const VertexList *pVertexList(nullptr);
     (void) PandoraContentApi::GetCurrentList(*pAlgorithm, pVertexList);
```
PandoraPFA, line 466: What do you think about i) checking whether pVertexLists is not nullptr here, and ii) checking to see whether the list isn't empty, up front, here? Because can't calculate vertexDistance below if there's no vertex. Could then remove some vertex checks below, but that might actually make implementation less robust against future changes.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
absolution1: I've reinstated the var

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
absolution1: Included

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
absolution1: Good shout, doing the sanity checks earlier is more in keeping with the other pandora code snippets I've seen.
I've added this in.
One point, I've slightly broken the style of the function ie. the vertex list sanity check fills the feature vector and returns.  However, the more complicated checks still only set `vertexDistance` and then wait for the final `push_back` at the end of the function.

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
```diff
@@ -461,10 +461,14 @@ void ThreeDVertexDistanceFeatureTool::Run(LArMvaHelper::MvaFeatureVector &featur
         std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
 
     LArMvaHelper::MvaFeature vertexDistance;
-    bool error = true;
     const VertexList *pVertexList(nullptr);
     (void) PandoraContentApi::GetCurrentList(*pAlgorithm, pVertexList);
 
+    if (!pVertexList || pVertexList->empty()){
```
PandoraPFA, line 467: Typical style is to have the brace `{` on the next line 😄

PandoraPFA (approved):
This looks good to me now (one small cosmetic comment). We should't merge this until we've discussed with @AndyChappell (who's looking after the big picture right now), but if you want this for rapid testing, can push to a nicely-named feature branch?

absolution1:
I've rebased my commits so the comments are probably going to be a bit mangled...

larpandoracontent/LArTrackShowerId/TrackShowerIdFeatureTool.cc
AndyChappell: Currently, in the case of an empty vertex list the returned distance to the vertex is zero, based on discussion elsewhere I think the conclusion was that a std::numeric_limits::max variant was preferred, is that still the case?

AndyChappell: In the update above to get the MvaFeature type consistency for the vertex distance, the value used has been reset to zero in the case where the vertex list is empty, which, from discussion elsewhere, I think is not what is wanted.

absolution1: H @AndyChappell 
My intention was to return a default-initialised `vertexDistance` when the actual vertex distance couldn't be calculated, which I think is what the original code tried to do.  Is that not the case?

absolution1: Ah I saw the email notification for your other comment (but cannot find it here bizarrely).
The very original plan was to return an extreme value for a default but that was before I knew that the type was wrapped in an initialisation bool.  So, I opted to preserve how the original code handled defaults.

AndyChappell (approved):
Happy with the fix

PR #142, opened 2020-10-07
PR_CRP_CheckForDP

echardon:
The DP team added a sanity check verifying that only 2D clusters that have hits, at least partially, in a common daughter volume should be matched.

We added some Geometry and Cluster helper functions to perform the check.

It is done in three algorithms : 1) TwoViewTransverseTrack 2)CosmicRayTrackRecovery**(this one actually touches Single-Phase)** 3)CosmicRayBaseMatching **(this one also touches Single-Phase)**

A small addition was also made in the MasterAlgorithm.cc in order to initialize properly the new LArCaloHit object that now have a new member variable called m_daughterVolumeId.

A parallel pull request https://github.com/PandoraPFA/larpandora/pull/9 has been issued for larpandora repository regarding the implementation of daughter volumes for DP and the addition of the new member variable in LArCaloHit.

PR #151, opened 2020-12-01
Feature/l ar calo hit factory with version rebased

MariaBrigida:
This PR introduces a larCaloHit version number to LArCaloHitFactories and to the Master, EventReading and EventWriting algorithms.This allows flexibility to handle new larCaloHit parameters, such as a newly introduced daughter volume ID, to be used to ensure that two-view calorimetric matching is not performed for clusters entirely contained in separate volumes.

This PR needs to be carefully integrated with the existing PR https://github.com/PandoraPFA/LArContent/pull/142 where the daughter volume ID is also introduced.

### DIFF
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -670,7 +670,7 @@ StatusCode MasterAlgorithm::Copy(const Pandora *const pPandora, const CaloHit *c
     // ATTN Parent of calo hit in worker is corresponding calo hit in master
     parameters.m_pParentAddress = static_cast<const void*>(pCaloHit);
     parameters.m_larTPCVolumeId = pLArCaloHit->GetLArTPCVolumeId();
-    parameters.m_daughterVolumeId = (m_larCaloHitVersion>1) ? pLArCaloHit->GetDaughterVolumeId() : 0;
+    parameters.m_daughterVolumeId = (m_larCaloHitVersion > 1) ? pLArCaloHit->GetDaughterVolumeId() : 0;
 
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraApi::CaloHit::Create(*pPandora, parameters, m_larCaloHitFactory));
 
@@ -1083,7 +1083,6 @@ StatusCode MasterAlgorithm::RegisterCustomContent(const Pandora *const /*pPandor
 
 StatusCode MasterAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
 {
-    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ReadValue(xmlHandle, "LArCaloHitVersion", m_larCaloHitVersion));
 
     ExternalSteeringParameters *pExternalParameters(nullptr);
 
@@ -1189,6 +1188,8 @@ StatusCode MasterAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
 
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
         "VisualizeOverallRecoStatus", m_visualizeOverallRecoStatus));
+
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ReadValue(xmlHandle, "LArCaloHitVersion", m_larCaloHitVersion));
 
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
         "ShouldRemoveOutOfTimeHits", m_shouldRemoveOutOfTimeHits));
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -670,8 +670,7 @@ StatusCode MasterAlgorithm::Copy(const Pandora *const pPandora, const CaloHit *c
     // ATTN Parent of calo hit in worker is corresponding calo hit in master
     parameters.m_pParentAddress = static_cast<const void*>(pCaloHit);
     parameters.m_larTPCVolumeId = pLArCaloHit->GetLArTPCVolumeId();
-    if(m_larCaloHitVersion>1) parameters.m_daughterVolumeId = pLArCaloHit->GetDaughterVolumeId();
-    else parameters.m_daughterVolumeId=0;
+    parameters.m_daughterVolumeId = (m_larCaloHitVersion>1) ? pLArCaloHit->GetDaughterVolumeId() : 0;
 
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraApi::CaloHit::Create(*pPandora, parameters, m_larCaloHitFactory));
 
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -669,6 +669,7 @@ StatusCode MasterAlgorithm::Copy(const Pandora *const pPandora, const CaloHit *c
     // ATTN Parent of calo hit in worker is corresponding calo hit in master
     parameters.m_pParentAddress = static_cast<const void*>(pCaloHit);
     parameters.m_larTPCVolumeId = pLArCaloHit->GetLArTPCVolumeId();
+    parameters.m_daughterVolumeId = pLArCaloHit->GetDaughterVolumeId();
 
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraApi::CaloHit::Create(*pPandora, parameters, m_larCaloHitFactory));
 
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -34,6 +34,7 @@ namespace lar_content
 
 MasterAlgorithm::MasterAlgorithm() :
     m_workerInstancesInitialized(false),
+    m_larCaloHitVersion(1),
     m_shouldRunAllHitsCosmicReco(true),
     m_shouldRunStitching(true),
     m_shouldRunCosmicHitRemoval(true),
@@ -669,7 +670,8 @@ StatusCode MasterAlgorithm::Copy(const Pandora *const pPandora, const CaloHit *c
     // ATTN Parent of calo hit in worker is corresponding calo hit in master
     parameters.m_pParentAddress = static_cast<const void*>(pCaloHit);
     parameters.m_larTPCVolumeId = pLArCaloHit->GetLArTPCVolumeId();
-    parameters.m_daughterVolumeId = pLArCaloHit->GetDaughterVolumeId();
+    if(m_larCaloHitVersion>1) parameters.m_daughterVolumeId = pLArCaloHit->GetDaughterVolumeId();
+    else parameters.m_daughterVolumeId=0;
 
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraApi::CaloHit::Create(*pPandora, parameters, m_larCaloHitFactory));
 
@@ -1082,6 +1084,8 @@ StatusCode MasterAlgorithm::RegisterCustomContent(const Pandora *const /*pPandor
 
 StatusCode MasterAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
 {
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ReadValue(xmlHandle, "LArCaloHitVersion", m_larCaloHitVersion));
+
     ExternalSteeringParameters *pExternalParameters(nullptr);
 
     if (this->ExternalParametersPresent())
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -1083,7 +1083,6 @@ StatusCode MasterAlgorithm::RegisterCustomContent(const Pandora *const /*pPandor
 
 StatusCode MasterAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
 {
-
     ExternalSteeringParameters *pExternalParameters(nullptr);
 
     if (this->ExternalParametersPresent())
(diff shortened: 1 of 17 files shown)


larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -1081,6 +1083,8 @@ StatusCode MasterAlgorithm::RegisterCustomContent(const Pandora *const /*pPandor
 
 StatusCode MasterAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
 {
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ReadValue(xmlHandle, "LArCaloHitVersion", m_larCaloHitVersion));
```
PandoraPFA, line 1086: Suggest move this down to around line ~1193, so that ReadSettings implementation still starts by focusing on  the fiddly external settings work.

larpandoracontent/LArControlFlow/MasterAlgorithm.h
```diff
@@ -307,6 +307,8 @@ class MasterAlgorithm : public pandora::ExternallyConfiguredAlgorithm
 
     bool                        m_workerInstancesInitialized;       ///< Whether all worker instances have been initialized
 
+    unsigned int                m_larCaloHitVersion;		    ///< The LarCaloHit version for LArCaloHitFactory
```
PandoraPFA, line 310: Ensure declared in same order as initialisation in any constructors, and assignment in the ReadSettings callback.

Check indentation of comment, and capitalisation of LArCaloHit.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -82,8 +90,10 @@ class LArCaloHit : public object_creation::CaloHit::Object
 
 private:
     unsigned int            m_larTPCVolumeId;       ///< The lar tpc volume id
+    unsigned int            m_daughterVolumeId;     ///< The daughter volume id
     pandora::InputFloat     m_pTrack;               ///< The probability that the hit is track-like
     pandora::InputFloat     m_pShower;              ///< The probability that the hit is shower-like
+
```
PandoraPFA, line 96: Empty line.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -94,6 +104,19 @@ class LArCaloHit : public object_creation::CaloHit::Object
...
+     *
+     */
+    LArCaloHitFactory();
+
+    /**
+     *  @brief  Constructor
+     *
+     *  @param  version the LArCaloHit version
+     */
+    LArCaloHitFactory(const unsigned int version);
```
PandoraPFA, line 118: Is it worth having a `...(const unsigned int version = 1);` here, or is the default (explicitly zero argument) constructor required still?

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -200,16 +248,27 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
 {
     // ATTN: To receive this call-back must have already set file reader mc particle factory to this factory
     unsigned int larTPCVolumeId(std::numeric_limits<unsigned int>::max());
+    unsigned int daughterVolumeId(std::numeric_limits<unsigned int>::max());
 
     if (pandora::BINARY == fileReader.GetFileType())
     {
         pandora::BinaryFileReader &binaryFileReader(dynamic_cast<pandora::BinaryFileReader&>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(larTPCVolumeId));
+        if(m_version>1)
```
PandoraPFA, line 257: Cosmetics. Indentation of parentheses, and quite a big splash in implementation for operation. Maybe initialise daughterVolumeId to zero, and have a single line `if (m_version > 1) PANDORA...`

Note spaces around operator>

Here's where we start to actually enforce the versioning scheme, of course. Here we check > 1, so that suggests new volume id in version 2 onwards. We should just check what we want to set as a precedent for versioning.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -200,16 +248,27 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
...
+	else
+        {
+            daughterVolumeId=0;
+        }
     }
     else if (pandora::XML == fileReader.GetFileType())
     {
         pandora::XmlFileReader &xmlFileReader(dynamic_cast<pandora::XmlFileReader&>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("LArTPCVolumeId", larTPCVolumeId));
+        if(m_version>1) {PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("DaughterVolumeId", daughterVolumeId));}
```
PandoraPFA, line 270: See above comments, for binary file reader.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -236,11 +296,13 @@ inline pandora::StatusCode LArCaloHitFactory::Write(const Object *const pObject,
     {
         pandora::BinaryFileWriter &binaryFileWriter(dynamic_cast<pandora::BinaryFileWriter&>(fileWriter));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(pLArCaloHit->GetLArTPCVolumeId()));
+        if(m_version>1)PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(pLArCaloHit->GetDaughterVolumeId()));
```
PandoraPFA, line 299: spaces after `if`, either side of `>`, and after closing `)`. And again below.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -236,11 +296,13 @@ inline pandora::StatusCode LArCaloHitFactory::Write(const Object *const pObject,
...
     }
     else if (pandora::XML == fileWriter.GetFileType())
     {
         pandora::XmlFileWriter &xmlFileWriter(dynamic_cast<pandora::XmlFileWriter&>(fileWriter));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable("LArTPCVolumeId", pLArCaloHit->GetLArTPCVolumeId()));
+        if(m_version>1)PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileWriter.WriteVariable("DaughterVolumeId", pLArCaloHit->GetDaughterVolumeId()));
```
PandoraPFA, line 305: As above.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -34,6 +34,7 @@ namespace lar_content
 
 MasterAlgorithm::MasterAlgorithm() :
     m_workerInstancesInitialized(false),
+    m_larCaloHitVersion(1),
```
PandoraPFA, line 37: Is `1` the current LArCaloHit then? Just need to settle on whether that was 0 or 1, I suppose. Matters most in that it sets a precedent for future such changes.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -82,8 +90,10 @@ class LArCaloHit : public object_creation::CaloHit::Object
 
 private:
     unsigned int            m_larTPCVolumeId;       ///< The lar tpc volume id
+    unsigned int            m_daughterVolumeId;     ///< The daughter volume id
```
PandoraPFA, line 93: Probably makes sense to do it this way. Just putting it out there, that this could be a pandora::InputType (pandora::InputUInt, exists, I think?), so that it could be explicitly uninitialised if the information isn't there.

The advantage of that is that any code accessing this volume id will bail, if it wasn't provided.

The disadvantage is that it's more complicated, there's a small overhead, and we might not want everything to bail, but just proceed with the view that everything has the same daughter volume id.

So, probably fine, but just putting this down for the record.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -669,6 +670,7 @@ StatusCode MasterAlgorithm::Copy(const Pandora *const pPandora, const CaloHit *c
     // ATTN Parent of calo hit in worker is corresponding calo hit in master
     parameters.m_pParentAddress = static_cast<const void*>(pCaloHit);
     parameters.m_larTPCVolumeId = pLArCaloHit->GetLArTPCVolumeId();
+    parameters.m_daughterVolumeId = (m_larCaloHitVersion>1) ? pLArCaloHit->GetDaughterVolumeId() : 0;
```
PandoraPFA, line 673: Spaces around `>`.

See later comment about default value for daughter volume id.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -124,14 +147,18 @@ class LArCaloHitFactory : public pandora::ObjectFactory<object_creation::CaloHit
      *  @param  pObject to receive the address of the object created
      */
     pandora::StatusCode Create(const Parameters &parameters, const Object *&pObject) const;
+
+private:
+     unsigned int	m_version;
```
PandoraPFA, line 152: Nice to add a simple comment/label for doxygen to pick up.

PandoraPFA (commented):
Thanks! Just a few issues that I've flagged for some discussion, so that we can make an explicit, upfront decision about them, then move onwards.

-Versioning scheme. Should we expect the pre-daughter-volume id to be v0, and daughter volume ids to appear in v1, or should these ids appear in v2? Need to set a precedent for future such changes to other objects.

-Default initialisation state. Probably makes sense for all LArCaloHits to have a defined daughter volume id, even if it's just 0 for all. But, putting it out there for discussion that we could leave them explicitly uninitialised. The infrastructure is there.

Nice to see this is only a small set of changes! Thanks again!

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
AndyChappell: I think for versioning my preference is for the first "released" version to be version 1, rather than 0, though no strong opinion

larpandoracontent/LArObjects/LArCaloHit.h
AndyChappell: I think this approach is consistent with the handling of m_larTPCVolumeId? i.e. the equivalent variable in LArCaloHitParameters is defined as an InputUInt, and then here it is a primitive uint.

larpandoracontent/LArObjects/LArCaloHit.h
AndyChappell: Agreed, I think the defaulted version is to be preferred here assuming there's no compiler shenanigans to worry about (I can't think of anything)

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -124,14 +147,18 @@ class LArCaloHitFactory : public pandora::ObjectFactory<object_creation::CaloHit
...
 };
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 inline LArCaloHit::LArCaloHit(const LArCaloHitParameters &parameters) :
     object_creation::CaloHit::Object(parameters),
-    m_larTPCVolumeId(parameters.m_larTPCVolumeId.Get())
+    m_larTPCVolumeId(parameters.m_larTPCVolumeId.Get()),
+    m_daughterVolumeId(parameters.m_daughterVolumeId.IsInitialized()? parameters.m_daughterVolumeId.Get() : 0)
```
AndyChappell, line 161: Space before '?'

AndyChappell (commented):
Looks good overall to me, Maria. A few minor tweaks, plus any further discussion relating to some of John's comments and I think this one is almost ready.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
MariaBrigida: I don't have a strong feeling on this; we could maybe add a line of comment to explain the convention, whichever it is we settle on.

MariaBrigida: I don't have a strong feeling on this; we could maybe add a line of comment to explain the convention, whichever it is we settle on.

larpandoracontent/LArControlFlow/MasterAlgorithm.h
MariaBrigida: Thanks, I've fixed the comment indentation! The capitalisation is the same as for m_larCaloHitFactories.
The variable is declared in same order as constructor initialisation, but the order in ReadSettings is different, as discussed [here](https://github.com/PandoraPFA/LArContent/pull/151/files#r533328123).

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
PandoraPFA: I also don't mind. I guess I had an expectation that our existing unversioned LArCaloHit might be henceforth referred to as the original/zeroth version, but it could equally just be the "first" version. Let's stick with original being v1, and new (first addition of daughter volume id) being v2, then, as implemented?

larpandoracontent/LArObjects/LArCaloHit.h
PandoraPFA: Yes, I think it's good that it's treated in the same way as the main "lar tpc" volume id. The only difference I can think of is that the reconstruction strategy now mandates that all hits have a volume id (otherwise you can't work in the context of the master algorithm, and need to go back to processing all input hits in one go). The daughter volume id, meanwhile, has no algs using it yet, and the strategy might be just to have one or two algs that look at the id. So optional is currently possible, but consistency with an existing id is a strong motivation, and probably wins.

MariaBrigida: Yes, this is done in the same way as for m_larTPCVolumeId. Do we want to change the behaviour for both, or leave as is?

PandoraPFA: Hi, I don't think we should change the implementation of the mandatory volume id, so the question is how we handle the optional daughter volume id. It's looking like it's going to remain optional from the app (or input pndr/xml) side. We could keep it optional in the LArCaloHits too, but I think we're leaning towards this implementation, where it's a simple unsigned int member variable, defaulted to zero.

The only thing we lose by doing that is knowledge of when our algs are using/accessing the daughter volume id when one hasn't been explicitly provided. Moving to a pandora::InputUInt would make the algs bail when information that wasn't explicitly provided was accessed. All the other input parameters (barring the provision of track-like and shower-like probabilities) _must_ be explicitly provided by the app/pndr/xml, I think.

It's just this new one that the app/pndr/xml can skip providing, but we'll (in LArContent) choose what it's default value shall be. Hence it's worth having this discussion. Shall we say the unsigned int (default value zero) is a preferred option right now (as opposed to pandora::InputUInt)? Thanks!

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
MariaBrigida: Sounds good to me, thanks!

larpandoracontent/LArObjects/LArCaloHit.h
MariaBrigida: Thanks for the clarifications @PandoraPFA  - I actually hadn't seen your previous comment when I submitted mine... so, sorry for the repetition!

At the moment we use this daughter volume ID in matching algorithms in [this PR down the line](https://github.com/PandoraPFA/LArContent/pull/142) in a check on whether cluster hits share a common daughter volume, so there should be no issues if the daughter volume ID is a default value, same for all.

So yes, happy to leave it as it is for now! Thanks!

PandoraPFA (commented):
Many thanks - just some super-minor cosmetics I picked up on a second look through. Once done, this can be folded in with PR#142 as works best. i.e. Keep these commits as approved, and replay PR#142 commits on top of them to reach development target for that feature. Thanks again.

PandoraPFA (approved):
Thanks, @MariaBrigida

AndyChappell (approved):

PR #156, opened 2021-02-10
MC and PFO visualisation algorithm

AndyChappell:
This PR introduces new visualisation tools for MC particles and PFOs to provide more coherent and comprehensive information to aid assessment of reconstructed events. MC particles are presented with PDG information and matched between views, PFOs are presented with track/shower characterisation, are matched between views and can be linked to their best matched MC particle.

### DIFF
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -829,6 +829,27 @@ bool LArMCParticleHelper::IsDescendentOf(const MCParticle *const pMCParticle, co
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArMCParticleHelper::GetBreadthFirstHierarchyRepresentation(const MCParticle *const pMCParticle, MCParticleList &mcParticleList)
+{
+    const MCParticle *const root{LArMCParticleHelper::GetParentMCParticle(pMCParticle)};
+    MCParticleList queue;
+    mcParticleList.emplace_back(root);
+    queue.emplace_back(root);
+
+    while (!queue.empty())
+    {
+        const MCParticleList &daughters{queue.front()->GetDaughterList()};
+        queue.pop_front();
+        for (const MCParticle *pDaughter : daughters)
+        {
+            mcParticleList.emplace_back(pDaughter);
+            queue.emplace_back(pDaughter);
+        }
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
 void LArMCParticleHelper::SelectGoodCaloHits(const CaloHitList *const pSelectedCaloHitList, const LArMCParticleHelper::MCRelationMap &mcToTargetMCMap,
     CaloHitList &selectedGoodCaloHitList, const bool selectInputHits, const float minHitSharingFraction)
 {
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.cc
@@ -831,10 +831,10 @@ bool LArMCParticleHelper::IsDescendentOf(const MCParticle *const pMCParticle, co
 
 void LArMCParticleHelper::GetBreadthFirstHierarchyRepresentation(const MCParticle *const pMCParticle, MCParticleList &mcParticleList)
 {
-    const MCParticle *const root{LArMCParticleHelper::GetParentMCParticle(pMCParticle)};
+    const MCParticle *const pRoot{LArMCParticleHelper::GetParentMCParticle(pMCParticle)};
     MCParticleList queue;
-    mcParticleList.emplace_back(root);
-    queue.emplace_back(root);
+    mcParticleList.emplace_back(pRoot);
+    queue.emplace_back(pRoot);
 
     while (!queue.empty())
     {
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -393,7 +393,9 @@ public:
     static bool IsDescendentOf(const pandora::MCParticle *const pMCParticle, const int pdg, const bool isChargeSensitive = false);
 
     /**
-     *  @brief  Retrieve a linearised representation of the MC particle hierarchy in breadth first order
+     *  @brief  Retrieve a linearised representation of the MC particle hierarchy in breadth first order. This iterates over the MC
+     *          hierarchy in a manor that sees primaries at the front of the list, with progressively deeper tiers later in the list.
+     *          This is useful for some visualisation cases.
      *
      *  @param  pMCParticle an MC particle in the hierarchy - can be any particle
      *  @param  mcParticleList the output MC particle list
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -392,6 +392,14 @@ public:
      */
     static bool IsDescendentOf(const pandora::MCParticle *const pMCParticle, const int pdg, const bool isChargeSensitive = false);
 
+    /**
+     *  @brief  Retrieve a linearised representation of the MC particle hierarchy in breadth first order
+     *
+     *  @param  pMCParticle an MC particle in the hierarchy - can be any particle
+     *  @param  mcParticleList the output MC particle list
+     */
+    static void GetBreadthFirstHierarchyRepresentation(const pandora::MCParticle *const pMCParticle, pandora::MCParticleList &mcParticleList);
+
 private:
     /**
      *  @brief  For a given Pfo, collect the hits which are reconstructable (=good hits belonging to a selected reconstructable MCParticle)
--- a/larpandoracontent/LArHelpers/LArPfoHelper.cc
+++ b/larpandoracontent/LArHelpers/LArPfoHelper.cc
@@ -648,18 +648,18 @@ bool LArPfoHelper::SortByNHits(const ParticleFlowObject *const pLhs, const Parti
 
 void LArPfoHelper::GetBreadthFirstHierarchyRepresentation(const pandora::ParticleFlowObject *const pPfo, pandora::PfoList &pfoList)
 {
-    const ParticleFlowObject *root{pPfo};
-    PfoList parents{root->GetParentPfoList()};
+    const ParticleFlowObject *pRoot{pPfo};
+    PfoList parents{pRoot->GetParentPfoList()};
     while (!parents.empty())
     {
         if (parents.size() > 1)
             throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
-        root = parents.front();
-        parents = root->GetParentPfoList();
+        pRoot = parents.front();
+        parents = pRoot->GetParentPfoList();
     }
     PfoList queue;
-    pfoList.emplace_back(root);
-    queue.emplace_back(root);
+    pfoList.emplace_back(pRoot);
+    queue.emplace_back(pRoot);
 
     while (!queue.empty())
     {
--- a/larpandoracontent/LArHelpers/LArPfoHelper.cc
+++ b/larpandoracontent/LArHelpers/LArPfoHelper.cc
@@ -646,6 +646,35 @@ bool LArPfoHelper::SortByNHits(const ParticleFlowObject *const pLhs, const Parti
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArPfoHelper::GetBreadthFirstHierarchyRepresentation(const pandora::ParticleFlowObject *const pPfo, pandora::PfoList &pfoList)
+{
+    const ParticleFlowObject *root{pPfo};
+    PfoList parents{root->GetParentPfoList()};
+    while (!parents.empty())
+    {
+        if (parents.size() > 1)
+            throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
+        root = parents.front();
+        parents = root->GetParentPfoList();
+    }
+    PfoList queue;
+    pfoList.emplace_back(root);
+    queue.emplace_back(root);
+
+    while (!queue.empty())
+    {
+        const PfoList &daughters{queue.front()->GetDaughterPfoList()};
+        queue.pop_front();
+        for (const ParticleFlowObject *pDaughter : daughters)
+        {
+            pfoList.emplace_back(pDaughter);
+            queue.emplace_back(pDaughter);
+        }
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
 template <typename T>
 void LArPfoHelper::SlidingFitTrajectoryImpl(const T *const pT, const CartesianVector &vertexPosition, const unsigned int layerWindow,
     const float layerPitch, LArTrackStateVector &trackStateVector, IntVector *const pIndexVector)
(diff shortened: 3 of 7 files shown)


CrossR:
As a general comment (about this, and the existing vis), should the lists that are being visualised be sorted?

I can't remember what the existing vis uses for sorting (I believe the `VisualisePFOs`/`Tracks`/`MC` etc methods have built in sorting based on energy?), but the way I use the event display, I almost always want the particle with the most hits at the very top of the list. I'd give up the tiny bit of CPU time to do the sorting vs me clicking randomly to find the biggest track.

That said, I'm not sure how other people use the event display, and if they are toggling elements in the sidebar to look at stuff more cleanly. Either way, maybe an extra config option, if a default doesn't seem smart? (Apologies if this already does something like this, I couldn't see it in the code, or from a quick test in a local build).

AndyChappell:
Hi Ryan, thanks for the feedback. The sorting here is essentially done via a breadth first search through the relevant hierarchy, so that the order should be some variation on primaries -> secondaries  -> tertiaries etc.

I could also provide an XML option to sort by the number of hits where this is desirable.

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -829,6 +829,27 @@ bool LArMCParticleHelper::IsDescendentOf(const MCParticle *const pMCParticle, co
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArMCParticleHelper::GetBreadthFirstHierarchyRepresentation(const MCParticle *const pMCParticle, MCParticleList &mcParticleList)
+{
+    const MCParticle *const root{LArMCParticleHelper::GetParentMCParticle(pMCParticle)};
```
PandoraPFA, line 834: Style: usually use `pRoot`, or `pRootMCParticle` here.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -392,6 +392,14 @@ class LArMCParticleHelper
      */
     static bool IsDescendentOf(const pandora::MCParticle *const pMCParticle, const int pdg, const bool isChargeSensitive = false);
 
+    /**
+     *  @brief  Retrieve a linearised representation of the MC particle hierarchy in breadth first order
+     *
```
PandoraPFA, line 397: Would be great if can set a precedent for adding a bit more info here, and confirm there's nothing already present that is very similar.

larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -646,6 +646,35 @@ bool LArPfoHelper::SortByNHits(const ParticleFlowObject *const pLhs, const Parti
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArPfoHelper::GetBreadthFirstHierarchyRepresentation(const pandora::ParticleFlowObject *const pPfo, pandora::PfoList &pfoList)
+{
+    const ParticleFlowObject *root{pPfo};
```
PandoraPFA, line 651: Suggest `pRoot` or `pRootPfo` or similar.

larpandoracontent/LArHelpers/LArPfoHelper.h
```diff
@@ -387,6 +387,14 @@ class LArPfoHelper
      */
     static bool SortByNHits(const pandora::ParticleFlowObject *const pLhs, const pandora::ParticleFlowObject *const pRhs);
 
+    /**
+     *  @brief  Retrieve a linearised representation of the PFO hierarchy in breadth first order
+     *
```
PandoraPFA, line 392: As per MCParticleHelper (although less urgency, as role of existing individual functions is clearer in PfoHelper)

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.h
```diff
@@ -0,0 +1,105 @@
...
+
+#ifdef MONITORING
+    /**
+     *  @brief  Visualize all MC particles independently.
+     *
+     *  This function ensures each MC particle has its own entry in the TEve tree, though common PDG colour scheme is used.
+     *
+     *  @param  mcMap The map from MC particles to calo hits
+     **/
+    void VisualizeIndependentMC(const LArMCParticleHelper::MCContributionMap &mcMap);
```
PandoraPFA, line 43: Can any of these member functions be const?

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+
+VisualParticleMonitoringAlgorithm::~VisualParticleMonitoringAlgorithm()
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode VisualParticleMonitoringAlgorithm::Run()
+{
+#ifdef MONITORING
```
PandoraPFA, line 42: Looks like this should all be good, but a test with MONITORING=OFF always helpful (and we could add this to our travis CI build matrix).

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+
+void VisualParticleMonitoringAlgorithm::VisualizeIndependentMC(const LArMCParticleHelper::MCContributionMap &mcMap)
+{
+    std::map<int, const std::string> keys = {{13, "mu"}, {11, "e"}, {22, "gamma"}, {321, "kaon"}, {211, "pi"}, {2212, "p"}};
+    std::map<int, Color> colors = {{0, RED}, {1, BLACK}, {2, BLUE}, {3, CYAN}, {4, MAGENTA}, {5, GREEN}, {6, ORANGE}, {7, GRAY}};
+    MCParticleList linearisedMC;
+    if (mcMap.empty())
+        return;
+
+    PANDORA_MONITORING_API(SetEveDisplayParameters(this->GetPandora(), true, DETECTOR_VIEW_XZ, -1.f, 1.f, 1.f));
```
PandoraPFA, line 92: OK to hardcode these last parameters? They are unlikely to change any time soon, but could check on what's done in usual LAr visual monitoring alg, and then make whatever decision you think best.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+#ifdef MONITORING
+
+void VisualParticleMonitoringAlgorithm::VisualizeIndependentMC(const LArMCParticleHelper::MCContributionMap &mcMap)
+{
+    std::map<int, const std::string> keys = {{13, "mu"}, {11, "e"}, {22, "gamma"}, {321, "kaon"}, {211, "pi"}, {2212, "p"}};
+    std::map<int, Color> colors = {{0, RED}, {1, BLACK}, {2, BLUE}, {3, CYAN}, {4, MAGENTA}, {5, GREEN}, {6, ORANGE}, {7, GRAY}};
```
PandoraPFA, line 87: Maps could be const, although means using `.at()` below, rather than `operator[]`.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+
+    PANDORA_MONITORING_API(ViewEvent(this->GetPandora()));
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void VisualParticleMonitoringAlgorithm::VisualizeMCByPdgCode(const LArMCParticleHelper::MCContributionMap &mcMap)
+{
+    std::map<int, const std::string> keys = {{13, "mu"}, {11, "e"}, {22, "gamma"}, {321, "kaon"}, {211, "pi"}, {2212, "p"}};
+    std::map<std::string, Color> colors = {{"mu", MAGENTA}, {"e", RED}, {"gamma", ORANGE}, {"kaon", BLACK}, {"pi", GREEN}, {"p", BLUE},
```
PandoraPFA, line 143: const, if poss.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+        {"other", GRAY}};
+
+    std::map<std::string, CaloHitList> uHits, vHits, wHits;
+    for (const auto [ key, value ] : keys)
+    {
+        (void)key;  // GCC 7 support, 8+ doesn't need this
```
PandoraPFA, line 149: Good to note, and to look out for in our different build use cases.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+                    wHits[key].emplace_back(pCaloHit);
+            }
+            catch (const StatusCodeException&)
+            {
+                continue;
+            }
+        }
+    }
+
+    PANDORA_MONITORING_API(SetEveDisplayParameters(this->GetPandora(), true, DETECTOR_VIEW_XZ, -1.f, 1.f, 1.f));
```
PandoraPFA, line 185: As per above. (Just to think about.)

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+    // ATTN - If we aren't showing matched MC, just pass in an empty MC to hits map
+    LArMCParticleHelper::MCContributionMap mcMap;
+    this->VisualizeIndependentPfo(pfoList, mcMap);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void VisualParticleMonitoringAlgorithm::VisualizeIndependentPfo(const PfoList &pfoList, const LArMCParticleHelper::MCContributionMap &mcMap)
+{
+    std::map<int, Color> colors = {{0, RED}, {1, BLACK}, {2, BLUE}, {3, CYAN}, {4, MAGENTA}, {5, GREEN}, {6, ORANGE}, {7, GRAY}};
```
PandoraPFA, line 230: const if poss.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+    PfoList linearisedPfo;
+    if (pfoList.empty())
+        return;
+
+    PANDORA_MONITORING_API(SetEveDisplayParameters(this->GetPandora(), true, DETECTOR_VIEW_XZ, -1.f, 1.f, 1.f));
```
PandoraPFA, line 235: As per above.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    PANDORA_MONITORING_API(ViewEvent(this->GetPandora()));
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void VisualParticleMonitoringAlgorithm::VisualizePfoByParticleId(const PfoList &pfoList)
+{
+    std::map<int, Color> colors = {{0, RED}, {1, BLUE}};
```
PandoraPFA, line 313: Const if poss.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+    PfoList linearisedPfo;
+    if (pfoList.empty())
+        return;
+
+    PANDORA_MONITORING_API(SetEveDisplayParameters(this->GetPandora(), true, DETECTOR_VIEW_XZ, -1.f, 1.f, 1.f));
```
PandoraPFA, line 318: As per above.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+void VisualParticleMonitoringAlgorithm::MakeSelection(const MCParticleList *pMCList, const CaloHitList *pCaloHitList,
+    LArMCParticleHelper::MCContributionMap &mcMap)
+{
+    // Default reconstructability criteria are very liberal to allow for unfolded hierarchy
+    LArMCParticleHelper::PrimaryParameters parameters;
+    parameters.m_minPrimaryGoodHits = 2;
+    parameters.m_minHitsForGoodView = 1;
+    parameters.m_maxPhotonPropagation = std::numeric_limits<float>::max();
+    parameters.m_minHitSharingFraction = 0;
+    parameters.m_foldBackHierarchy = false;
```
PandoraPFA, line 394: Are these OK to hardcode? Or worth making everything configurable, just as a matter of course? (Just to think about.)

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
```diff
@@ -0,0 +1,428 @@
...
+}
+#endif // MONITORING
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode VisualParticleMonitoringAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
+{
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "CaloHitListName", m_caloHitListName));
+    if (m_caloHitListName.empty())
+        m_caloHitListName = "CaloHitList2D";
```
PandoraPFA, line 415: This works OK. In other, similar cases, we've assigned these string values to the member variables in the constructor, and then allowed a STATUS_CODE_NOT_FOUND to slide by.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
AndyChappell: Yes, this has been tested with MONITORING=OFF and it works as expected

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
AndyChappell: Yes, this scenario has been seen before and our CI is GCC 7, so this does get picked up.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
AndyChappell: At some point I want to add configurability here, but I need to give some thought as to whether any of these parameters have implications for the way I prepare the visualisations, so I was thinking that would be for the next round of updates.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
AndyChappell: I've preferred this method here, because it covers the case where the tag is supplied but empty. Happy to adopt the constructor-based approach if that's preferred.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.h
AndyChappell: Yep, all of them can, will do.

larpandoracontent/LArMonitoring/VisualParticleMonitoringAlgorithm.cc
PandoraPFA: Your call 😄

Issue #157, opened 2021-02-24
Bug in TrainedVertexSelectionAlgorithm


henrylay97:
Found (what I believe to be) a bug in the TrainedVertexSelectionAlgorithm. In the GetEventShapeFeatures method the y and z spans are defined using the wrong max and min values:

`const float ySpan(this->GetCoordinateSpan(yMax, zMin));`
`const float zSpan(this->GetCoordinateSpan(yMax, zMin));`

PR #158, opened 2021-02-24
Bug fix: definintion of y and z spans

henrylay97:
Fixing the error in issue #157

### DIFF
--- a/larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
+++ b/larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
@@ -296,8 +296,8 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
     }
 
     const float xSpan(this->GetCoordinateSpan(xMax, xMin));
-    const float ySpan(this->GetCoordinateSpan(yMax, zMin));
-    const float zSpan(this->GetCoordinateSpan(yMax, zMin));
+    const float ySpan(this->GetCoordinateSpan(yMax, yMin));
+    const float zSpan(this->GetCoordinateSpan(zMax, zMin));
 
     // Calculate the volume and longitudinality of the event (ySpan often 0 - to be investigated).
     if ((xSpan > std::numeric_limits<float>::epsilon()) && (ySpan > std::numeric_limits<float>::epsilon()))


larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -296,8 +296,8 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
     }
 
     const float xSpan(this->GetCoordinateSpan(xMax, xMin));
-    const float ySpan(this->GetCoordinateSpan(yMax, zMin));
-    const float zSpan(this->GetCoordinateSpan(yMax, zMin));
+    const float ySpan(this->GetCoordinateSpan(yMax, yMin));
+    const float zSpan(this->GetCoordinateSpan(zMax, zMin));
```
PandoraPFA, line 300: This looks sound, thank you, but opens up some further discussion about other (existing) implementation, below.

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -296,8 +296,8 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
...
 
     // Calculate the volume and longitudinality of the event (ySpan often 0 - to be investigated).
```
PandoraPFA, line 302: There seems to be some 3D approach here applied to (currently always) a 2D input. In our usage, we would expect a `ySpan` (if implemented correctly, as modified above) to always be zero. That's because the 2D hits are always placed in a `y = constant = 0` plane. With `ySpan` always explicitly zero, this impacts the checks below.

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -296,8 +296,8 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
...
     if ((xSpan > std::numeric_limits<float>::epsilon()) && (ySpan > std::numeric_limits<float>::epsilon()))
```
PandoraPFA, line 303: Will now never evaluate as true (or not unless algorithm is retasked to work with 3D hit inputs; unlikely). 

Similarly, line 303 will never evaluate as true.

This then leaves line 315, which should nearly always evaluate as true. In this case, we probably want to modify `eventVolume` to `eventArea = xSpan * zSpan` and `longitudinality = zSpan / (xSpan + zSpan)`?

If `xSpan` and `zSpan` are zero, should leave the event feature uninitialised, so it should skip over in MVA.

PandoraPFA (commented):
Many thanks for spotting this! Please see attached comments, and comment below, which try to draw this together:

Idea here is (of course) to provide some features that help to distinguish between different event topologies, so that the selection can vary topology-by-topology. The suggested changes look good, but we think they should be accompanied by additional, knock-on changes.

We suggest that you implement all the new changes, but leave an option to still work with existing training outputs. So e.g. a Boolean XML parameter `Use2DEventShapes`, which, if true, uses new proposals; if false, reverts to existing implementation. We could even make it print a "deprecated!" message if attempting to train with `Use2DEventShapes` as false.

Thanks again.

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
henrylay97: Thanks for the comments @PandoraPFA, I'm currently updating the branch to reflect this. I agree that `eventVolume` should become `eventArea`. However, for `longitudinality` I would imagine we want it to be `zSpan / (xSpan + xSpan)` ? Or even potentially `zSpan / xSpan` for the 2D approach?

PandoraPFA: Hi @henrylay97 - I think we now have the opportunity to redevelop the event shapes as works best under new training and testing. In terms of my suggestion above, I was (without any deep thought) going with the idea of something that would range from ~0 (when the zSpan is significantly smaller than the xSpan), to 0.5 (for square aspect events) to ~1 (when the zSpan is significantly greater than the xSpan). That might be an OK starting point, but please do feel free to really dig into the role these event shapes are providing and let the final decisions be performance driven. Many thanks!

AndyChappell: I like John's suggestion here; having a well-defined range is usually a useful thing for machine learning applications, so I imagine it will end up yielding the best performance of those options, it also has the nice property that the only undefined case is for xSpan **and** zSpan equal to zero.

henrylay97: I see what you mean. It sounds good to me but I can certainly do some tests to check the performance implications. I've updated the branch with the things we've discussed here but this is mostly for book-keeping. Going through this has raised some points on how we will deal with deprecation for any other changes we potentially make during this study. I intend to train various permutations of the different changes so will include this particular change in those permutations so we can see the effect of that change. Thanks to both of you for your comments!

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -295,28 +296,49 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
...
     {
-        eventVolume     = ySpan * ySpan * zSpan;
-        longitudinality = zSpan / (ySpan + ySpan);
+      const float xSpan(this->GetCoordinateSpan(xMax, xMin));
+      const float zSpan(this->GetCoordinateSpan(zMax, zMin));
+      
+      if ((xSpan > std::numeric_limits<float>::epsilon()))
+      {
+	eventArea       = xSpan * zSpan;
+	longitudinality = zSpan / (xSpan + xSpan);
```
PandoraPFA: Should this be `zSpan / (xSpan + zSpan)`? Is it now actually OK to calculate this, even if `xSpan` is exactly zero (it shouldn't be negative, though!)? Check indentation throughout; seems that block of code is at same level of indentation as surrounding braces. Also look out for stray whitespace added. Thanks!

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
PandoraPFA: Hi @henrylay97 - a very good comment about how we will deal with deprecation. I would say that you ultimately don't want to be constrained by what's there now, so you should feel free to develop so as to obtain the best performance.

When we review what you find, and if you find any outright problems, then we'll have to discuss whether there's a need to allow new code to run with old training outputs, or whether we force a new training pass for everything. We could even split the algorithm implementation if needed. But keep us posted and we can discuss as things evolve. Thanks!

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -295,28 +296,49 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
...
-    else if (xSpan > std::numeric_limits<float>::epsilon())
+    
+    else
     {
-        eventVolume     = xSpan * xSpan * zSpan;
-        longitudinality = zSpan / (xSpan + xSpan);
-    }
+      if (m_trainingSetMode)
+      {
+	std::cout << "TrainedVertexSelectionAlgorithm: WARNING -- Using deprecated event shape features for training, "
```
johnmarshall80, line 316: This warning will be printed a lot (maybe a good thing!), but could consider moving it to the ReadSettings method, so that it's printed clearly once, at the start of a training job, after the XML parsing has been done and configuration is fixed. We can also add a TODO comment here, next to the error, to help note that the legacy mode is to disappear once new trainings are available for the different use cases.

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -295,28 +296,49 @@ void TrainedVertexSelectionAlgorithm::GetEventShapeFeatures(const ClusterList &c
...
+    if (m_use2DEventShapes)
     {
-        eventVolume     = ySpan * ySpan * zSpan;
-        longitudinality = zSpan / (ySpan + ySpan);
+      const float xSpan(this->GetCoordinateSpan(xMax, xMin));
+      const float zSpan(this->GetCoordinateSpan(zMax, zMin));
+      
+      if ((xSpan > std::numeric_limits<float>::epsilon()))
+      {
+	eventArea       = xSpan * zSpan;
```
johnmarshall80, line 307: Some whitespace and indentation issues, but suggest we merge after the clang format changes have been applied, and then we can simply apply clang format to these updated files too, so these are non-issues now.

johnmarshall80 (commented):
A couple of small suggestions made, but this otherwise ready to merge, I think. So approved, pending a tiny change to warning print out and adding a TODO comment. We'll then handle application of the .clang-format file, although brief instructions available [here](https://github.com/PandoraPFA/LArContent/pull/162) if keen. Many thanks!

henrylay97:
As discussed offline, there are some further question marks over the interchangeable use of the z-coordinate across the three planes and the vulnerability of the span calculations to single outlying hits. I'm looking at possible solutions to these issues and will try and prepare an intermediate version with these improvements before the rest of the development work.

henrylay97:
A further developed version of this bug fix (along with other vertex selection improvements) can be found in PR #169.

PR #159, opened 2021-03-02
Fix CosmicRayTrackRecovery algorithm failure when matching small clusters

AndyChappell:
This PR addresses an issue in which a cluster with fewer than 3 hits could be considered for matching via a sliding linear fit, which would in turn throw an exception due to insufficient hits for the procedure.

This PR also includes a minor update to conditional expressions in the deep learning library that would fail to build if LArContent was built without Pandora Monitoring enabled, and a build flag update to avoid suppressing sign comparison warnings in the Pandora standalone build (and therefore matching the warnings generated in a LArSoft build).


absolution1:
I had a quick look.  Nothing jumped out to me 👌 🎉

PandoraPFA (approved):
Looks good, thanks.

Good spot/testing r.e. the `PANDORA_MONITORING_API` usage. Some of the other preprocessor macros actually include the opening and closing curly braces to help with this kind of issue. But here, the macro just explicitly evaluates to no code if monitoring is disabled.

PR #160, opened 2021-03-19
Calomatchval

absolution1:
---Please ignore the cheating vertex finder files, those were included by accident---

Validation tool for the two view transverse tracks 2D->3D matching alg.
The validation tool's only role is to write data to a TTree so that its easy to make charge profile plots.

The main bit I'd like to draw attention to is the underlying data structure, rather than which data is dumped to the TTree.  The data structure provides a way of storing data in a TTree using a single line of code and removes the need to explicitly define individual data members for each piece of data that you'd want to store in a TTree.  All of the data to be saved (regardless of type) is stored in a single map.

The easy-to-use interface comes at a cost; the underlying implementation is a bit whacky.  I've deliberately tried to avoid use of `dynamic_cast` and solely rely on polymorphism.
I am very much interested in feedback/comments/whatever on the data structure and how it could be improved.

The data structure still needs certain additions (such as registering branch data with default values) but its already usable enough, so heyho...


absolution1:
Oopsie, looks like I've included @MariaBrigida's cheating vertex finder in this PR.  I will disentangle next week.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+    }
+    else
+        throw pandora::StatusCodeException(pandora::STATUS_CODE_NOT_ALLOWED);
+    return;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template <typename T>
+void TwoViewTransverseTracksValidationTool::StoreAndRegisterDatum(const T &t, const std::string datumName, TreeDataBox &treeDataBox)
```
AndyChappell, line 211: Called store and register, but doesn't appear to register

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+
+TwoViewTransverseTracksValidationTool::TwoViewTransverseTracksValidationTool() :
+    m_treeName("mytree"), 
+    m_outputFileName("output.root")
+{
+}
+
+TwoViewTransverseTracksValidationTool::~TwoViewTransverseTracksValidationTool()
+{
+    PANDORA_MONITORING_API(SaveTree(this->GetPandora(), m_treeName.c_str(), m_outputFileName.c_str(), "UPDATE"));
```
AndyChappell, line 27: Do we want tree writing to be optional here?

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewTransverseTracksValidationTool::Run(const Cluster *const pCluster1, const Cluster *const pCluster2,
+    const DiscreteProbabilityVector &discreteProbabilityVector1, const DiscreteProbabilityVector &discreteProbabilityVector2,
+    const TwoViewTransverseOverlapResult &overlapResult)
+{
+
```
AndyChappell, line 36: Excess whitespace

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+    TreeDataBox treeDataBox;
+
+    this->CollectTruthInformation(pCluster1, treeDataBox, "1");
+    this->CollectTruthInformation(pCluster2, treeDataBox, "2");
+
+    StoreAndRegisterDatum(static_cast<int>(this->IsSameMCParticle(pCluster1, pCluster2)), "isSameMCParticle", treeDataBox);
```
AndyChappell, line 42: Do we always want to retain all validation information?

AndyChappell (commented):
Hi Dom. My main comments on the implementation itself relates to whether one aspect of the public interface can be made private. I also wonder if the datum classes might be moved to their own file, as the structures appear to be somewhat independent of the actual transverse tracks validation tool.

The bigger question for me is how much do we gain with this implementation over the existing monitoring function calls? I may be missing something in just reviewing the code that would become more obvious if actually using it (is this more general/reusable than I'm thinking?), but do we gain anything beyond the reduction in variable declarations, given that all of the variables still require registration? Thanks!

absolution1:
Hi @AndyChappell 

> Hi Dom. My main comments on the implementation itself relates to whether one aspect of the public interface can be made private.

Ah yeah, an oversight on my part.  Thank you!

>I also wonder if the datum classes might be moved to their own file, as the structures appear to be somewhat independent of the actual transverse tracks validation tool.

This was actually a deliberate choice as one of the functions needed access to one of the tool's private functions and it was also a lazy attempt to hide away some of the internals.  I was thinking about the class structure a bit of the weekend and I'm not sure I actually do want to hide the behind-the-scenes classes this much.  The class does still need access to a private function of the tool but I guess I could use befriending for that...

> The bigger question for me is how much do we gain with this implementation over the existing monitoring function calls? I may be missing something in just reviewing the code that would become more obvious if actually using it (is this more general/reusable than I'm thinking?), but do we gain anything beyond the reduction in variable declarations, given that all of the variables still require registration? Thanks!

The long term intent is that the code is completely reusable elsewhere in pandora.  Right now its contained within this particular tool while its features/quirks are being tested and ironed out.  
There's no benefit for memory reduction or CPU cycles by there is, in my opinion, a large benefit when it comes to writing/maintaining/extending user code that writes variables to a TTree.  For example here: https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArMonitoring/TestBeamEventValidationAlgorithm.cc#L118
it's really annoying having to explicitly define every single variable that you'd like to dump to a TTree.  The code in this PR would eliminate having to do this.
There are also the potential for other ease-of-use benefits such as automatic variable reseting which I'd like to implement somewhere down the line.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+ *  @file   larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
+ *
+ *  @brief  Header file for the two view transverse tracks validation tool
+ *
+ *  $Log: $
+ */
+#ifndef TWO_VIEW_TRANSVERSE_TRACKS_VALIDATION_TOOL_H
+#define TWO_VIEW_TRANSVERSE_TRACKS_VALIDATION_TOOL_H 1
+
+#include "PandoraMonitoringApi.h"
```
CrossR, line 11: Will this (and maybe some other places) need wrapping in an `#ifdef MONITORING`?

Since its possible to compile Pandora without monitoring support, I would assume so, though I believe all the macro usage is fine. Could be worth compiling in standalone without it to check maybe?

AndyChappell: Good spot, yes they will.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+
+bool TwoViewTransverseTracksValidationTool::IsSameMCParticle(const Cluster *const pCluster1, const Cluster *const pCluster2)
+{
+    const MCParticle *const pMCParticle1(MCParticleHelper::GetMainMCParticle(pCluster1));
+    const MCParticle *const pMCParticle2(MCParticleHelper::GetMainMCParticle(pCluster2));
+
+    if (pMCParticle1 && pMCParticle2)
+    {
+        return pMCParticle1->GetUid() == pMCParticle2->GetUid();
+    }
```
lhwhitehead, line 75: I don't really mind this, but style is to remove braces here?

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewTransverseTracksValidationTool::CollectChargeProfileInformation(
+    const DiscreteProbabilityVector &discreteProbabilityVector, TreeDataBox &treeDataBox, std::string clusterName)
+{
+    std::string xName("xValues" + clusterName);
+    std::string probabilityDensityName("probDensityValues" + clusterName);
+    std::string probabilityName("probValues" + clusterName);
+    std::string cumulativeProbabilityName("cumulProbValues" + clusterName);
+    std::string widthName("widthValues" + clusterName);
```
lhwhitehead, line 131: Do const strings work here, or does it cause issues with the `TreeDataBox`?

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+{
+
+    TreeDataBox treeDataBox;
+
+    this->CollectTruthInformation(pCluster1, treeDataBox, "1");
+    this->CollectTruthInformation(pCluster2, treeDataBox, "2");
+
+    StoreAndRegisterDatum(static_cast<int>(this->IsSameMCParticle(pCluster1, pCluster2)), "isSameMCParticle", treeDataBox);
+
+    StoreAndRegisterDatum(static_cast<int>(discreteProbabilityVector1.GetSize()), "discProbVecSize1", treeDataBox);
```
johnmarshall80, line 44: Usually like to see the explicit `this->` to signify a member function, but is called frequently, so might look OTT. (Subsequent discussion) If we move this to PandoraMonitoring will make context clear anyway.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+    emplace_back_impl(std::any(t));
+    return;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template <typename T>
+TwoViewTransverseTracksValidationTool::TreeDatum<T>::TreeDatum(
+    const T &t, const TwoViewTransverseTracksValidationTool *const pTool, const std::string datumName) :
+    m_Datum(t)
```
johnmarshall80, line 189: So this is the big line, I think. Previously the variable needed to be in scope at the point when FillTree was called. Now that's no longer the case, I think, and indeed can have multiple different algorithms (under same Pandora instance) writing to the same tree until one finally calls FillTree. Allows calls to run off rvalues, tidying things up!

johnmarshall80 (commented):
Hi @absolution1 - thanks very much. Just a few additional thoughts, alongside the good coverage from others. I have tried to avoid saying the same thing (there might be a repeated comment or two) and agree with the other comments sent so far.

I agree with the key point you've flagged ("it's really annoying having to explicitly define every single variable that you'd like to dump to a TTree"):

The question then is whether to merge this here, and risk it staying here for ages, to move it to a handy part of LArContent and upgrade other usage, or to make it available in PandoraMonitoring, and tie release of this tool to an upgrade of the Pandora external product (fine, if patient!).

Sticking with the current approach to quickly release this tool, then adding the new PandoraMonitoring functionality (based off this implementation) might work well? Thanks again.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h
```diff
@@ -90,6 +92,8 @@ class TwoViewTransverseTracksAlgorithm : public NViewTrackMatchingAlgorithm<TwoV
     float                       m_minOverallMatchingScore;               ///< The minimum required global matching score to fill the overlap result
     float                       m_minOverallLocallyMatchedFraction;      ///< The minimum required lcoally matched fraction to fill the overlap result
     std::mt19937                m_randomNumberGenerator;                 ///< The random number generator
+
+    TwoViewTransverseTracksValidationTool* p_ValidationTool;             ///< The validation tool
```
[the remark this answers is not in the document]
absolution1: @AndyChappell TY! I have updated the name.

@johnmarshall80 ahh there is a non-const member function but that might be changeable.  The bigger issue is that I have to assign the data member when reading from the xml:
```
m_pValidationTool = dynamic_cast<TwoViewTransverseTracksValidationTool*>(*iter);
```

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+         *  @param  datumName the label for the data
+         */
+        TreeDatum(const T &t, const TwoViewTransverseTracksValidationTool *const pTool, const std::string datumName);
+
+        /**
+         *  @brief  Overriding implementation of emplace_back.   
+         *
+         *  @param  value the data to be added to the data structure
+         */
+        void emplace_back_impl(const std::any &value) override;
```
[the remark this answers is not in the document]
absolution1: It can and should!
Complete oversight on my part.  TY!

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+    std::string m_outputFileName; ///< name of output file
+};
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template <typename T>
+void TwoViewTransverseTracksValidationTool::AbstractDatum::emplace_back(const T &t)
+{
+    emplace_back_impl(std::any(t));
+    return;
```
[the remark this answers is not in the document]
absolution1: How much of a deal breaker is this?
This is a totally personal thing but I really like functions to have a reference at the end...  If it goes against house style then I will remove.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+    pTool->RegisterTreeDatum(m_Datum, datumName);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template <typename T>
+void TwoViewTransverseTracksValidationTool::TreeDatum<T>::emplace_back_impl(const std::any &value)
+{
+    if constexpr (is_std_vector<T>::value)
+    {
```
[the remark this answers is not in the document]
absolution1: Got it.  TY!

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
absolution1: The registering happens in the `TreeDatum` that's constructed as it is `emplace_back`'d onto the map.  From an interface point of view, storing and registering does happen within that function call but the registering is outsourced to a constructor call, where this happens:
`pTool->RegisterTreeDatum(m_Datum, datumName);`

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
absolution1: The validation tool, as it is, is supposed to make a TTree by design.  As we look to move the base functionality out (such as moving the TTree data registration to the SDK, for example) then we may want to revisit this question.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
absolution1: Got it.  TY!

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
absolution1: For this particular validation tool, I think so.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
absolution1: Got it.  TY Ryan.
As far as I can tell, I only need to the protection for including the `PandoraMonitoringApi.h` header as, elsewhere, I only talk to the API via the provided C macro: (`PANDORA_MONITORING_API`).  Am I correct here?

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
absolution1: Got the braces.  TY @lhwhitehead

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
absolution1: No issues caused, just a slip up on my end.
TY @lhwhitehead

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
```diff
@@ -0,0 +1,164 @@
...
+    StoreAndRegisterDatum(isPrimaryMCParticle, ("isPrimaryMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(energyMCParticle, ("energyMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(momXMCParticle, ("momXMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(momYMCParticle, ("momYMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(momZMCParticle, ("momZMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(posXMCParticle, ("posXMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(posYMCParticle, ("posYMCParticle" + clusterName), treeDataBox);
+    StoreAndRegisterDatum(posZMCParticle, ("posZMCParticle" + clusterName), treeDataBox);
+
+    return;
```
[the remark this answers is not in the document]
absolution1: I mentioned this in a similar comment by Andy.
I really like functions to have a well defined return at the end (completely personal preference obviously!).
If this clashes with house style then I am happy to change but otherwise I'd rather leave in place...

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     *  @brief  Store some data in a box and register it with the pandora monitoring api for writing to a TTree 
+     *
+     *  @param  t the data to be saved
+     *  @param  datumName the data label
+     *  @param  treeDataBox the box where the data is stored before TTree writing
+     */
+    template <typename T>
+    void StoreAndRegisterDatum(const T &t, const std::string datumName, TreeDataBox &treeDataBox);
```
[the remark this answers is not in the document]
absolution1: Got em.  TY

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+        throw pandora::StatusCodeException(pandora::STATUS_CODE_NOT_ALLOWED);
+    return;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template <typename T>
+void TwoViewTransverseTracksValidationTool::StoreAndRegisterDatum(const T &t, const std::string datumName, TreeDataBox &treeDataBox)
+{
+    auto [iterator, inserted] = treeDataBox.try_emplace(datumName, std::make_unique<TreeDatum<T>>(t, this, datumName));
```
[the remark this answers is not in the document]
absolution1: Got it.  The solution was icky and is a bit harder to read unfortunately
```
    if (!(treeDataBox.try_emplace(datumName, std::make_unique<TreeDatum<T>>(t, this, datumName))).second)
        throw pandora::StatusCodeException(pandora::STATUS_CODE_NOT_FOUND);
```

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h
```diff
@@ -11,6 +11,8 @@
 #include "Pandora/Algorithm.h"
 #include "Pandora/AlgorithmTool.h"
 
+#include "larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h"
```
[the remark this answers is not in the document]
absolution1: I'm calling public functionality in the implementation file:
`m_pValidationTool->Run(pCluster1, pCluster2, resampledDiscreteProbabilityVector1, resampledDiscreteProbabilityVector2, overlapResult);`
Does a forward declaration still make sense in that case?

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
```diff
@@ -0,0 +1,237 @@
...
+
+        /**
+         *  @brief  Overriding implementation of emplace_back.   
+         *
+         *  @param  value the data to be added to the data structure
+         */
+        void emplace_back_impl(const std::any &value) override;
+
+    private:
+        T m_Datum;      ///< the datum that is destined for a TTree
```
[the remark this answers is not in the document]
absolution1: got it

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
absolution1: Yeah, I generally agree.  Given that it will cause me to go into an excess number of characters per line (and what you said above also applies) I might just leave as is in this instance.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
absolution1: Exactly!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
```diff
@@ -137,6 +138,9 @@ pandora::StatusCode TwoViewTransverseTracksAlgorithm::CalculateOverlapResult(con
     overlapResult = TwoViewTransverseOverlapResult(matchingScore, m_downsampleFactor, nComparisons, nLocallyMatchedSamplingPoints,
         correlation, twoViewXOverlap);
 
+    if (p_ValidationTool) 
+        p_ValidationTool->Run(pCluster1, pCluster2, resampledDiscreteProbabilityVector1, resampledDiscreteProbabilityVector2, overlapResult);
```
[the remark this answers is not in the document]
absolution1: tyty

absolution1:
OK, I think I have addressed the comments.  I've marked the trivial/style changes as resolved but left all others open so that the reviewers can decide on whether the fixes resolve the matter.
I think the next (and I guess final) step is to untangle the mistaken merge with Maria's branch.
I had a chat with @AndyChappell about this and an interactive rebase + forced push should work.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h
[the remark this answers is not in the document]
johnmarshall80: Actually, there doesn't seem to be too much precedent for working hard to make the tools const, so just handle as best fits needs here.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
johnmarshall80: Anything enclosed by `PANDORA_MONITORING_API` will reduce to no code if Monitoring has not been enabled in the build. Still typically want to test a no-monitoring build, seeing as we have -wUnused and -wError as build flags, and can often have end up with unused variables in the monitoring-off situation.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
johnmarshall80: (This'll go in PandoraMonitoring, typically released alongside the SDK as a single Pandora external product, with two libraries.)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.h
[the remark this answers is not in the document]
johnmarshall80: Yes, as the header file can still be included in the implementation file.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.cc
johnmarshall80: Characters per line now under control of .clang-format, so suggest do what's intrinsically best and leave the line-length management to that.

larpandoracontent/LArMonitoring/TwoViewTransverseTracksValidationTool.h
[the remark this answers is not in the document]
johnmarshall80: I'll leave the casting vote to @AndyChappell - I don't mind this, but wouldn't do it myself. Wonder if there's even another magic line we can add to .clang-format to make this a rule? Not a deal breaker, though!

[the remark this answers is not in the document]
AndyChappell: It's not a deal breaker for me either, so I'm ok with this remaining as is, but wouldn't advocate mandating this in void functions generally.

PR #161, opened 2021-03-22
Fix for V-shaped events in ClusterAssociationAlgorithm

astblake:
I've simplified the ClusterAssociationAlgorithm::UpdateForAmbiguousMerge(...) method so that ambiguously-merged clusters are entirely removed from the cluster association map.

Previously, the UpdateForAmbiguousMerge(...) method was more elaborate and also called the UpdateForUnambiguousMerge(...) method. I think this last part was the cause of our V-shaped events.

Although the previous implementation was more elaborate, I think the net effect was to simply remove the clusters-to-be-merged from the cluster association map. So, I've coded this in a simpler way - but I worry that I've missed something!

### DIFF
--- a/larpandoracontent/LArTwoDReco/LArClusterAssociation/ClusterAssociationAlgorithm.cc
+++ b/larpandoracontent/LArTwoDReco/LArClusterAssociation/ClusterAssociationAlgorithm.cc
@@ -106,6 +106,7 @@ void ClusterAssociationAlgorithm::UnambiguousPropagation(const Cluster *const pC
         return;
 
     this->UpdateForUnambiguousMerge(pClusterToEnlarge, pClusterToDelete, isForward, clusterAssociationMap);
+
     PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::MergeAndDeleteClusters(*this, pClusterToEnlarge, pClusterToDelete));
     m_mergeMade = true;
 
@@ -147,11 +148,14 @@ void ClusterAssociationAlgorithm::AmbiguousPropagation(const Cluster *const pClu
 
     for (ClusterVector::iterator dIter = daughterClusterVector.begin(), dIterEnd = daughterClusterVector.end(); dIter != dIterEnd; ++dIter)
     {
-        this->UpdateForAmbiguousMerge(pCluster, *dIter, isForward, clusterAssociationMap);
+        this->UpdateForAmbiguousMerge(*dIter, clusterAssociationMap);
+
         PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::MergeAndDeleteClusters(*this, pCluster, *dIter));
         m_mergeMade = true;
         *dIter = NULL;
     }
+
+    this->UpdateForAmbiguousMerge(pCluster, clusterAssociationMap);
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


astblake:
More (slightly cryptic) information on why this is a bug...

The "UpdateForAmbiguousMerge" method was designed to receive clusters in their precise order along the chain:

   c0 --> c1 --> c2 --> c3  etc...

The ordering is important because forward/backward links are pruned and propagated after each merge (so c1 has to be added to c0, before c2 can be added to c0). 

However, at some point we sorted the input vector according to the number of hits, so that the order changed, and we started merging clusters more like this:

  c0 --> c3 --> c1 --> c2 etc...

The careful manipulation of forward/backward associations no longer works. In the example above, the cluster ends up growing out along the forward links of "c2" and creating a V-like event, rather than stopping at "c3".

I've changed the behaviour of "UpdateForAmbiguousMerge" so that the clusters can be merged in any order. Before each merge, the "clusterToDelete" is simply removed from all points of the clusterAssociationMap.

This fixes the V-shaped events, and I think it preserves the intended behaviour of this method - but I'm not completely sure!

johnmarshall80:
Thanks @astblake - I've been looking at this again to remind myself of what's going on. I think I now understand everything, but there's still a small amount of hesitancy about all of the below:

This is indeed all about the sorting call that has been added, and the problem is that it's meant to sort any ambiguities at a given _link_ in the chain, and **not** sort all collected clusters along the _entire_ chain.

The update mechanics are intricate and look good, but they are indeed assuming that we are moving along the chain, link by link. Updating using the clusters in the wrong order will leave dodgy links in the map, but all still valid from the point of view of the memory management (should never crash, but could head off along the wrong direction).

Your fix removes the clusters entirely from the map, whereas the previous code selectively removed just the associations at the relevant link, link by link, keeping the cluster association map dynamically updated.

Now, removing everything from the map is fine, so long as we don't want to do any more work here. Unfortunately, I can't guarantee at the moment that it's sensible to break the map. Nothing bad will happen in terms of crashes, etc., but we could miss out on subsequent merges.

At least, that's how it works in principle. The map should always remain up to date. But... it might actually be the case that processing a chain for any given cluster only needs to be done once, and it's OK to proceed with the map losing information.

Really, it's about removing the ClusterSets, and sorting (if we actually decide to sort) only if there are multiple possible merges at a given link. That's what wants to be sorted, I think, and not the whole set of linked clusters, which is where the mistake you spotted entered (at a later date to original alg implementation, of course!!!).

I hope this all makes sense. Might need to upload some hand-drawn diagrams soon. Thanks very much, and let's see how easy it is to address this sorting in the correct manner, and then can see if that fixes the V events still.

astblake:
Hi John, Yes, you're quite right: we could fix this issue by ordering the "daughterClusterVector" so that the clusters appear in the vector in the order of their forward (or backward) associations. Initially I thought this would be really hard to do, but it occurs to me that we could use some adaptation or variation of the "NavigateAlongAssociations" method. 

The current cluster-merging mechanics would then stay in place - however, it may be okay (and possibly simpler?) to just remove the ambiguously-merged clusters from the cluster association map, as we identify them. This wouldn't work for the unambiguously-merged clusters (these clusters have to stay in the map until the ambiguities are resolved), but I think it may be okay for ambiguously-merged clusters. I'm not sure though - I vaguely recall that we had a reason for keeping them in the map!

In any case, I spotted a few events where the cluster associations look like this:

  c0 --> c1 --> c2 --> c4 --> c5
                 \ --> c3 -->/
 
i.e. c1 connects to both c2 and c3 before merging back in to c4
I wonder whether we handle these cases correctly...?

astblake:
We could use a method like this to fill the cluster vector in the correct order:

```
void Sort(const ClusterAssociationMap clusterAssociationMap, const Cluster *const pCluster, const bool isForward, const ClusterSet &secondClusterSet, ClusterVector &daughterClusterVector)
{
    ClusterAssociationMap::const_iterator iterAssociation = clusterAssociationMap.find(pCluster);

    if (clusterAssociationMap.end() == iterAssociation)
        throw StatusCodeException(STATUS_CODE_NOT_INITIALIZED);

    const ClusterSet &associatedClusterSet(isForward ? iterAssociation->second.m_forwardAssociations : iterAssociation->second.m_backwardAssociations);

    for (ClusterSet::const_iterator iter = associatedClusterSet.begin(), iterEnd = associatedClusterSet.end(); iter != iterEnd; ++iter)
    {
        if ((secondClusterSet.count(*iter)) &&
            (daughterClusterVector.end() == std::find(daughterClusterVector.begin(), daughterClusterVector.end(), *iter)))  
            daughterClusterVector.push_back(*iter);
            
        this->Sort(clusterAssociationMap, *iter, isForward, secondClusterSet, daughterClusterVector);
    }
}
```

johnmarshall80:
Hi @astblake - Can you try looking at your events again with the following branch of this repository: `feature/ClusterAssociationOrdering`, and see how things look. If things have worked as intended (_if!_), then the different clusters should now be ordered in the container in the same order as they are encountered along the chain of possible merges. This may well be be at the cost of some efficiency, but these algorithms are far from rate-limiting. Please let us know how this goes, and we can hopefully move this forwards. Thanks!

johnmarshall80:
A summary of latest thoughts/findings associated with this PR.

Issues:

- In the ambiguous merging, links in the chosen chain can have multiple clusters. All that's required is that they feature on the forwards navigation to the extremal cluster and again in the backwards navigation from the extremal cluster. Presumably "links" can start to overlap/shortcut, and so our picture of this as a chain can get messy too.
- If there are multiple clusters at a given link, the order in which the clusters are presented for merging does not match the logical ordering along the  trajectory. This is because of the way clusters are picked up by navigation. The key point is that the order in which clusters are merged determines the leftover start/end associations of the final, enlarged cluster, which can go on to see additional unambiguous merges.

Approaches:

- We can completely remove the enlarged cluster from the cluster association map, as in this PR. This is completely safe, but it's difficult to gauge its effects entirely. We could just do this if we can't think of (or find, under testing) a way in which the alg is actually badly affected. Will need to add an ATTN comment to the code about removing information from the map. This approach doesn't stop multiple clusters being merged at a given link, but just stops any later unambiguous merges.
- Also not preventing multiple clusters being merged at a given link, but keeping the ClusterAssociation map up to date, we could perform a PCA to all of the hits involved and then order the clusters to be merged by their projected endpoints along that trajectory. Whichever endpoint of the cluster projects furthest away from the original parent cluster.
- Somewhat separately, and preventing multiple clusters being merged at a given link, we could choose a unique route along, where there's only one cluster per link. The way the navigation works means that ordering would still be required to prevent inappropriate associations being attached to the enlarged cluster and actioned in later unambiguous merges.

larpandoracontent/LArTwoDReco/LArClusterAssociation/ClusterAssociationAlgorithm.cc
```diff
@@ -194,69 +198,29 @@ void ClusterAssociationAlgorithm::UpdateForUnambiguousMerge(const Cluster *const
...
-            clusterSetEnlarge.erase(iter++);
-        }
-        else
-        {
-            ++iter;
-        }
-    }
-
-    for (ClusterSet::iterator iter = clusterSetDelete.begin(); iter != clusterSetDelete.end();)
+    for (ClusterAssociationMap::iterator mIter = clusterAssociationMap.begin(), mIterEnd = clusterAssociationMap.end(); mIter != mIterEnd; ++mIter)
```
johnmarshall80, line 208: We could sort the clusters before looping over them here. It shouldn't make any difference, but means this should always happen in the same order (a nice thing, especially if anything unexpected happens when e.g. if this code was modified in future).

johnmarshall80 (commented):
Performance testing indicates that this simple fix does not harm performance (it does change some event outcomes, but not for systematically better or worse). It makes sense to take this simple fix, then, rather than add some significant complexities in order to keep the map perfectly up-to-date (and avoid the cluster sorting issue), given that the map doesn't seem to have significant further usage.

astblake:
Hi there - John has written an excellent of summary of the issues/approaches in the above comments!  (I'm pleased that we still understand this algorithm after >5 years - it must have been written by a pro...!).

Yes, I'd suggest moving forward with my PR, which I think will do the job. I'll take another look over the weekend, and it's good that Andy C will review this too. Hopefully it will get a double-Andy seal of approval!

A few notes:
 1) By construction, the "ambiguous navigation" of associations in the map always begins and ends on clusters that have no further forward or backward associations. This means that, once ambiguously-merged clusters have been made, they will never contribute to any further ambiguous or unambiguous associations and/or merges, and can be safely removed from the association map.
 2) Hopefully the following makes sense ... There's a mechanism whereby the resolution of ambiguous associations can lead to new unambiguous associations. For example, if we have a criss-cross association (c1, c2) <--> (c3, c4) and we decide to merge c1 <--> c3, then this will lead to the subsequent merger of c2 <--> c4. I believe that this would happen with the existing algorithm, but it will definitely happen after my PR! 
3) I think this is a very rare bug! It takes extreme events like Dom's small-angle di-muons to trigger a V-shaped event, so I wouldn't expect to see much change in performance. It's good that the performance doesn't change much, as it means that the algorithm retains a similar behaviour after the bug fix.
4) It wouldn't surprise me if the shower reconstruction was slightly worse after the bug fix, as the creation of V-shaped clusters probably helps the shower reconstruction...

AndyChappell (approved):
Nothing to add beyond John's comments, all looks good to me.

AndyChappell:
Replicated with associated clang format changes in [PR 163](https://github.com/PandoraPFA/LArContent/pull/163)

PR #165, opened 2021-03-29
Feature/delta ray pre clang

imawby:
Hello all, this is the pull request for the delta rays. I think there might be a bit of work to do, but I wanted to get the ball rolling on this. In this PR I replace the DeltaRayMatching algorithm with three view matching, two view matching and a one view recovery algorithm. 

The three view algorithm uses some tools:
- UnambiguousDeltaRay - this is equivalent to the ClearTracks tool i.e. it makes pfos out of the really clear matches
- DeltaRayRemoval - this removes delta ray hits out of cosmic ray tracks
- CosmicRayRemoval - this removes cosmic ray hits out of delta ray tracks
- DeltaMergeTool - used matching info to merge delta ray clusters together
- AmbiguousDeltaRay - creates a pfo out of the best match in an ambiguous set of cluster matches (e.g. 3:2:1)

Two view tools:
- TwoViewCosmicRayRemoval - as above
- TwoViewDeltaRayMerge - this is the two view equivalent of ambiguous delta ray tool (a legacy name thing that i'll change)

I've cleaned up the three and two view algs and their tools and just need to sort out the one view alg. Although it shouldn't be too shabby. I also need to add doxygen comments, they'll be coming shortly. 

Any questions, feel free to reach out on slack.


larpandoracontent/LArThreeDReco/LArCosmicRay/ThreeViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,164 @@
...
+    for (const ParticleFlowObject *const pNearbyMuonU : nearbyMuonPfosU)
+    {
+        for (const ParticleFlowObject *const pNearbyMuonV : nearbyMuonPfosV)
+        {
+            for (const ParticleFlowObject *const pNearbyMuonW : nearbyMuonPfosW)
+            {
+                if ((pNearbyMuonU == pNearbyMuonV) && (pNearbyMuonV == pNearbyMuonW))
+                    commonMuonPfoList.push_back(pNearbyMuonU);
+            }
+        }
```
AndyChappell, line 105: There are a couple of options here for improving the efficiency. Under the assumption that a PFO can only appear in each list once you can split the conditional expressions to check U == V inside the V loop and simply continue if there is no match, and then perform a similar V == W check for the actual push_back (you could opt for emplace_back here, which is more efficient given we don't usually need a copy for these operations).

The other option would be to use `std::find(pNearbyMuonV->begin(), pNearbyMuonV->end(), pNearbyMuonU)` to see if it's the same PFO and then only do the second find if it is, avoiding the need for the two inner loops altogether (obviously find is still performing iteration, but I suspect it will be more efficient).

larpandoracontent/LArThreeDReco/LArCosmicRay/ThreeViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ThreeViewDeltaRayMatchingAlgorithm::ExamineOverlapContainer()
+{
+    unsigned int repeatCounter(0);
+
+    for (TensorToolVector::const_iterator toolIter = m_algorithmToolVector.begin(); toolIter != m_algorithmToolVector.end(); )
```
AndyChappell, line 115: Can we replace this with:

```
for (auto tool : m_algorithmToolVector)
    for (unsigned int rep = 0; rep <= m_nMaxTensorToolRepeats; ++rep)
        if (!tool->Run(this, this->GetMatchingControl().GetOverlapTensor()))
            break;
```

? (Check that I got the loop counter right - and open to thoughts on if and where we want braces here - they're not necessary, but perhaps desirable for clarity?)

larpandoracontent/LArThreeDReco/LArCosmicRay/ThreeViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,164 @@
...
+
+StatusCode ThreeViewDeltaRayMatchingAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
+{   
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ReadValue(xmlHandle, "MuonPfoListName", m_muonPfoListName));
+    
+    AlgorithmToolVector algorithmToolVector;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ProcessAlgorithmToolList(*this, xmlHandle,
+        "DeltaRayTools", algorithmToolVector));
+
+    for (AlgorithmToolVector::const_iterator iter = algorithmToolVector.begin(), iterEnd = algorithmToolVector.end(); iter != iterEnd; ++iter)
```
AndyChappell, line 141: `for (auto tool : algorithmToolVector)`

This is generally preferable if you don't explicitly need the iterator (which usually you don't if you aren't try to manipulate the underlying data structure), as it places the onus to manage comparisons and increments on the compiler rather than the developer

AndyChappell (commented):
Mini-review focusing on a couple of files identified as being of interest by Isobel (note I haven't looked at the base class yet).

larpandoracontent/LArThreeDReco/LArCosmicRay/ThreeViewDeltaRayMatchingAlgorithm.cc
imawby: i think i'm inclined to go with 

for (auto tool : m_algorithmToolVector)
{
    for (unsigned int rep = 0; rep <= m_nMaxTensorToolRepeats; ++rep)
    {
        if (!tool->Run(this, this->GetMatchingControl().GetOverlapTensor()))
            break;
    }
}

?

imawby: obvioulsy with the right indentations...

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -28,6 +28,8 @@ class LArMCParticleParameters : public object_creation::MCParticle::Parameters
 {
 public:
     pandora::InputInt   m_nuanceCode;               ///< The nuance code
+    pandora::InputBool  m_isDR;
+    pandora::InputBool  m_isDecay;
```
AndyChappell, line 32: (This comment applies across the entire file)

I think it's worthwhile to make this more generic, and I think this can be done with relatively little work. If you recall our previous discussions about the process information from some time ago, I started putting together such a feature branch (comparison to the 3.22.03 base here):

https://github.com/PandoraPFA/LArContent/compare/feature/larpandoracontent_v03_22_03...AndyChappell:feature/mc_updates

I've not had the opportunity to finish this off, but it should be pretty close to being release-worthy and I'd be happy to see it go out as part of this PR - essentially it just needs some versioning of the LArMCParticle so we can still load pndr files that don't have this information (my take would be this becomes the default version and old versions have to specify version 1 if they want it, but maybe @johnmarshall80 has a view on this?). Happy to chat about the versioning if needed.

I think you can largely replace what's here with the cherry picked commits from my feature branch, as the logic is essentially the same (The visualisation tools are unnecessary of course, but if it turns out to be easier to include them as part of the cherry pick, that's fine). If you want to retain the GetIsDR and GetIsDecay helper functions, I think that's fine, they can just perform a check against the value returned by GetProcess -  though I'd drop the "Get" from the function names.

johnmarshall80: Hi @AndyChappell, Yes, I'd agree with the new version becoming the default because pndr files are for development use by the team. We should ensure that the XML snippets to switch the version of the LArMCParticle are documented somewhere, and we should take a look at the error handling when different file/deserialising combinations are used to check that the error messages aren't too cryptic! Nice suggestion to connect-up these features. Thanks!

AndyChappell (commented):
Next mini review focusing on TwoViewDeltaRayMatching

larpandoracontent/LArThreeDReco/LArThreeDBase/TwoViewMatchingControl.h
```diff
@@ -73,6 +73,7 @@ class TwoViewMatchingControl : public NViewMatchingControl
 
     template <typename U>
     friend class NViewMatchingAlgorithm;
+    friend class TwoViewDeltaRayMatchingAlgorithm;
```
AndyChappell, line 76: As far as I can tell, the only reason for this class to be a friend is to have access to the private member variable m_hitTypeToIndexMap. If that's correct, I think it's preferable to define a public function in this class that can take a request for an index based on a view (returning something like -1 if the view isn't found in the map), rather than giving another class access to all of the implementation details of this class for the sake of a single variable.

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.h
```diff
@@ -0,0 +1,220 @@
...
+    HitTypeVector GetHitTypeVector();
+
+    /**
+     *  @brief  Get the address of the given hit type cluster
+     *
+     *  @param  hitType hit type of the required cluster
+     *
+     *  @return address of the required cluster
+     */
+    const pandora::Cluster *GetCluster(const MatrixType::Element &element, const pandora::HitType &hitType);
```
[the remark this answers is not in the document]
imawby: this is going to be a common comment, i'll try to remove as many as I can find..

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.h
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     *  @brief  Starting with an input seed cluster, sequentially merge in matched clusters that retain a good reduced chi-squared
+     *
+     *  @param  element the matrix element
+     *  @param  pSeedCluster the address of the input seed cluster
+     */        
+    void MergeThirdView(const MatrixType::Element &element, const pandora::Cluster *const pSeedCluster);
+
+    void ExamineOverlapContainer();
```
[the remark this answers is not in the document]
imawby: This function is documented in a base class (https://github.com/PandoraPFA/LArContent/blob/d6f31bd77da416976fd6f1c2a5dbfe12ccb6883b/larpandoracontent/LArThreeDReco/LArThreeDBase/MatchingBaseAlgorithm.h#L169) so do I still need to document it again?

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,466 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewDeltaRayMatchingAlgorithm::DoesClusterPassTensorThreshold(const Cluster *const pCluster) const
+{
+    return (pCluster->GetNCaloHits() >= m_minClusterCaloHits);
+}    
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+    
+void TwoViewDeltaRayMatchingAlgorithm::CalculateOverlapResult(const Cluster *const pCluster1, const Cluster *const pCluster2, const Cluster *const)
```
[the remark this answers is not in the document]
imawby: this should be the base function https://github.com/PandoraPFA/LArContent/blob/d6f31bd77da416976fd6f1c2a5dbfe12ccb6883b/larpandoracontent/LArThreeDReco/LArThreeDBase/MatchingBaseAlgorithm.h#L99

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,466 @@
...
+
+    if (matchedClusterList.empty())
+        return STATUS_CODE_NOT_FOUND;
+    
+    const Cluster *pBestMatchedCluster(nullptr); 
+    float reducedChiSquared(std::numeric_limits<float>::max());
+
+    this->GetBestMatchedCluster(pCluster1, pCluster2, commonMuonPfoList, matchedClusterList, pBestMatchedCluster, reducedChiSquared);
+
+    //ATTN: Ignore if other clusters matches have more hits
```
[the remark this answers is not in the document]
imawby: great question! so, the TwoViewDeltaRayMatching alg is essentially trying to pick up fragments in the third view (or allow clusters in two views to be matched if one understands that in the third view the cosmic ray is hiding the delta ray or one sees that the third view of the match has already been taken) but it's the first thing that is important here. So the alg runs three times, each time looking in a different view. Essentially when I create the match I want to do so with the best pair of clusters since this will give me the biggest projection. So if i find that my best matched projected cluster has more hits then one of the clusters i use to do the projection that is the one i want to form my pair. So if i find this, I leave the pair and don't put it in the matrix with the idea that when the next version of the alg runs it will pick up the pair and do the projecting with the best pair. It's important to make sure the 'pBestMatchedCluster' is available, you see it could be a muon so then this would always win the 'highest hit' contest which I don't really want to happen

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,466 @@
...
+        if (reducedChiSquared > m_maxGoodMatchReducedChiSquared)
+            continue;
+            
+        matchedClusters.push_back(pCluster);
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewDeltaRayMatchingAlgorithm::GetBestMatchedCluster(const Cluster *const pCluster1, const Cluster *const pCluster2, const PfoList &commonMuonPfoList,
```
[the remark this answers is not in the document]
imawby: agreed

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,466 @@
...
+{
+    auto &theMatrix(this->GetMatchingControl().GetOverlapMatrix());
+
+    for (auto iter1 = theMatrix.begin(); iter1 != theMatrix.end(); ++iter1)
+    {
+        auto iter2(iter1->second);
+        
+        for (auto &entry : iter2)
+        {
+            TwoViewDeltaRayOverlapResult &overlapResult(entry.second);
```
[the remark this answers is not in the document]
imawby: i didn't know this could be done, so snazzy! thanks!

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayMergeTool.cc
```diff
@@ -0,0 +1,362 @@
...
+    for (const ParticleFlowObject *const pCommonMuonPfo : commonMuonPfoList)
+    {
+        if (this->IsConnected(pCommonMuonPfo, pDeltaRayCluster))
+            connectedMuonPfoList.push_back(pCommonMuonPfo);
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+    
+bool DeltaRayMergeTool::IsConnected(const Pfo *const pCommonMuonPfo, const Cluster *const pCluster) const
```
AndyChappell, line 202: As above on argument order

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayMergeTool.cc
```diff
@@ -0,0 +1,362 @@
...
+{
+    for (const TensorType::Element &element1 : elementList)
+    {
+        for (const TensorType::Element &element2 : elementList)
+        {
+            if ((element1.GetCluster(TPC_VIEW_U) == element2.GetCluster(TPC_VIEW_U)) && (element1.GetCluster(TPC_VIEW_V) == element2.GetCluster(TPC_VIEW_V)) &&
+                (element1.GetCluster(TPC_VIEW_W) == element2.GetCluster(TPC_VIEW_W)))
+            {
+                continue;
+            }
```
AndyChappell, line 279: Same comment as with the two view version of this function

AndyChappell (commented):
Next mini-review focusing on the DR merge tool

larpandoracontent/LArThreeDReco/LArCosmicRay/UnambiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,114 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool UnambiguousDeltaRayTool::Run(ThreeViewDeltaRayMatchingAlgorithm *const pAlgorithm, TensorType &overlapTensor)
+{
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+       std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    bool changesMade(false);
```
AndyChappell, line 34: I don't think you need this variable (see below)

larpandoracontent/LArThreeDReco/LArCosmicRay/UnambiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,114 @@
...
+
+    TensorType::ElementList elementList;
+    overlapTensor.GetUnambiguousElements(true, elementList);
+
+    this->ExamineUnambiguousElements(pAlgorithm, elementList, changesMade);
```
AndyChappell, line 39: Just return the result of this function (see below)

larpandoracontent/LArThreeDReco/LArCosmicRay/UnambiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,114 @@
...
+
+    return changesMade;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void UnambiguousDeltaRayTool::ExamineUnambiguousElements(ThreeViewDeltaRayMatchingAlgorithm *const pAlgorithm, TensorType::ElementList &elementList, bool &changesMade)
```
AndyChappell, line 46: I think you can just return a boolean here, rather than passing in the reference variable - the logic of the function doesn't appear to be dependent on the value of the boolean that gets passed in

larpandoracontent/LArThreeDReco/LArCosmicRay/UnambiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,114 @@
...
+            continue;
+
+        ProtoParticle protoParticle;
+        protoParticle.m_clusterList.push_back(element.GetClusterU());
+        protoParticle.m_clusterList.push_back(element.GetClusterV());
+        protoParticle.m_clusterList.push_back(element.GetClusterW());
+        protoParticleVector.push_back(protoParticle);
+    }
+
+    changesMade |= pAlgorithm->CreatePfos(protoParticleVector);
```
AndyChappell, line 62: Can just make this a return statement

larpandoracontent/LArThreeDReco/LArCosmicRay/AmbiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,130 @@
...
+        pAlgorithm->CreatePfos(protoParticleVector);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void AmbiguousDeltaRayTool::PickOutGoodMatches(const TensorType::ElementList &elementList, ClusterSet &usedClusters, ProtoParticleVector &protoParticleVector) const
+{
+    bool found(true);
+    
+    while (found)
```
AndyChappell, line 74: Prefer do {...} while () when you know the code in the loop should be executed at least once

larpandoracontent/LArThreeDReco/LArCosmicRay/AmbiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,130 @@
...
+        const Cluster *pBestClusterU(nullptr), *pBestClusterV(nullptr), *pBestClusterW(nullptr);
+
+        for (const TensorType::Element &element : elementList)
+        {            
+            const Cluster *const pClusterU(element.GetClusterU()), *const pClusterV(element.GetClusterV()), *const pClusterW(element.GetClusterW());
+            
+            if (usedClusters.count(pClusterU) || usedClusters.count(pClusterV) || usedClusters.count(pClusterW))
+                continue;
+
+            const float chiSquared = element.GetOverlapResult().GetReducedChi2();
```
AndyChappell, line 89: Initialise () rather than assign

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewAmbiguousDeltaRayTool.h
```diff
@@ -0,0 +1,40 @@
...
+     */
+    TwoViewAmbiguousDeltaRayTool();
+
+private:
+    bool Run(TwoViewDeltaRayMatchingAlgorithm *const pAlgorithm, MatrixType &overlapMatrix);    
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+
+    void ExamineConnectedElements(TwoViewDeltaRayMatchingAlgorithm *const pAlgorithm, MatrixType &overlapMatrix) const;
+    
+    bool PickOutGoodMatches(TwoViewDeltaRayMatchingAlgorithm *const pAlgorithm, const MatrixType::ElementList &elementList) const;
```
AndyChappell, line 34: Following on from our discussion offline, could consider making the algorithm a member variable to avoid the need to specify it in multiple functions (no strong view on this),

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewAmbiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,108 @@
...
+    return false;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewAmbiguousDeltaRayTool::ExamineConnectedElements(TwoViewDeltaRayMatchingAlgorithm *const pAlgorithm, MatrixType &overlapMatrix) const
+{
+    bool particleCreated(true);
+
+    while (particleCreated)
```
AndyChappell, line 41: Prefer do {} while () structure when a loop must run at least once

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewAmbiguousDeltaRayTool.cc
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewAmbiguousDeltaRayTool::PickOutGoodMatches(TwoViewDeltaRayMatchingAlgorithm *const pAlgorithm, const MatrixType::ElementList &elementList) const
+{
+    unsigned int highestHitCount(0);
+    float bestChiSquared(std::numeric_limits<float>::max());
+    MatrixType::Element bestElement(nullptr, nullptr, TwoViewDeltaRayOverlapResult(TwoViewXOverlap(0.f, 0.f, 0.f, 0.f), PfoList(), nullptr, ClusterList(), 0.f));
```
AndyChappell, line 76: Rather than creating a dummy object to hold the best element, you could opt for a best index variable, initialised to -1, (you'd need to either maintain a loop counter or adopt the classic for loop style over the list) and then assuming the a best index is found, just use that to access the appropriate element of the list when creating the pfo

AndyChappell (commented):
Mini review of ambiguous and unambiguous tools

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,466 @@
...
+    element.GetCluster1()->GetOrderedCaloHitList().FillCaloHitList(caloHitList1);
+    element.GetCluster2()->GetOrderedCaloHitList().FillCaloHitList(caloHitList2);
+
+    // ATTN: Need copy as original will change throughout function 
+    ClusterList matchedClusters(element.GetOverlapResult().GetMatchedClusterList());
+        
+    ClusterSet checkedClusters;
+    checkedClusters.insert(pSeedCluster);
+        
+    while (checkedClusters.size() != matchedClusters.size())
```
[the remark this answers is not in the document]
imawby: Unless something has gone wrong, the answer is no. For this function to be called the pBestCluster has to not be a nullptr which means that the matchedClusters list has to have at least one element. The function UpdateForThirdViewClusterModification does all the cleanup of the matchedClusters list and the pBestMatchedCluster whenever a pfo is created. So, you have to let the underlying matrix update between pfo creations to prevent reading 'outdated' matched clusters lists? Which is why (when you come to it) in the TwoViewUnambiguousDR alg i have to create pfos one at a time.

However, you made me realise something.. I really hate this loop, essentially i am trying to get every element in matchedClusters checked but in a specific order dictated by their hit number. I try to stop the pSeedCluster from being evaluated since i dont want to merge something with itself so I add it into the checkedCluster set beforehand. However, i realised that the pSeedCluster isn't always in that list since, in the case of muons, I can actually create a new cluster which is then the pSeedCluster => I changed it to:

`    if (checkedClusters.count(pSeedCluster))
        checkedClusters.insert(pSeedCluster);`

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayRemovalTool.cc
```diff
@@ -0,0 +1,232 @@
...
+
+    changesMade = modifiedClusters.size();
+}
+    
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool DeltaRayRemovalTool::PassElementChecks(const TensorType::Element &element, const HitType hitType) const
+{
+    // ATTN: Avoid endpoints, topological michel reconstruction is very ambiguous
+    if (this->IsMuonEndpoint(element, false))
```
AndyChappell, line 115: This appears to be the only difference relative to the CosmicRayRemovalTool version of this function. This function may be a candidate to move up to the RemovalBaseTool, either providing the boolean conditional as a parameter and working out which version of IsMuonEndpoint to call, or by moving the code below this point to a common (protected) RemovalBaseTool function that could be called from this function in each of the inheriting classes.

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+    m_minContaminationLength(3.f),
+    m_maxDistanceToHit(1.f),
+    m_minRemnantClusterSize(3),
+    m_maxDistanceToTrack(2.f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewCosmicRayRemovalTool::Run(TwoViewDeltaRayMatchingAlgorithm *const pAlgorithm, MatrixType &overlapMatrix)
```
AndyChappell, line 36: This function appears to be identical for for each of the removal tools, aside from the type of algorithm that gets passed in. So, I think for the classes that inherit from the RemovalTool, you may want to change the algorithm type to the corresponding NView class they inherit from, then this function can be implemented just once in the RemovalBaseTool, rather than in each of the inheriting classes. For this class, I wonder if this indicates that the inheritance hierarchy isn't quite right? Perhaps the DeltaRayMatrixTool is not needed/might inherit from the RemovalBaseTool/might merge with the RemovalBaseTool? I'm not immediately sure which, if any of these options would be best, but I think there is a lot of overlap between these classes, so that may want further consideration.

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+
+        this->ExamineConnectedElements(elementList, changesMade);
+    }
+    
+    return changesMade;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewCosmicRayRemovalTool::ExamineConnectedElements(const MatrixType::ElementList &elementList, bool &changesMade) const
```
AndyChappell, line 68: This looks to be the same function as in the CosmicRayRemoval class, so is subject to the same comment as above.

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+            this->SplitDeltaRayCluster(element, hitType, deltaRayHits, deltaRayRemnantHits);
+        }
+    }
+
+    changesMade = modifiedClusters.size();
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewCosmicRayRemovalTool::PassElementChecks(const MatrixType::Element &element, const HitType hitType) const
```
AndyChappell, line 121: As above

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+    
+    if (separation > m_minSeparation)
+        return false;
+
+    return true;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewCosmicRayRemovalTool::IsMuonEndpoint(const MatrixType::Element &element, const HitType hitType) const
```
AndyChappell, line 142: Almost identical to the RemovalBaseTool version of this function

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+        if ((zMinDR < zMinCR) || (zMaxDR > zMaxCR))
+            return false;
+    }
+
+    return true;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewCosmicRayRemovalTool::IsContaminated(const MatrixType::Element &element, const HitType hitType) const
```
AndyChappell, line 178: Also essentially the same as the CosmicRayRemovalTool version. It appears to be just the element type that may be different, I'd need to check the relationship between the different types here, but I suspect this could similarly be consolidated into a base class.

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+        if (this->IsInLineSegment(deltaRayVertex, extendedPoint, pCaloHit->GetPositionVector()))
+            return false;
+    }
+
+    return true;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewCosmicRayRemovalTool::IsCloseToLine(const CartesianVector &hitPosition, const CartesianVector &lineStart, const CartesianVector &lineEnd, const float distanceToLine) const
```
AndyChappell, line 235: Same as RemovalBaseTool

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+    
+    if (transverseDistanceFromLine > distanceToLine)
+       return false;
+
+    return true;
+}    
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewCosmicRayRemovalTool::IsInLineSegment(const CartesianVector &lowerBoundary, const CartesianVector &upperBoundary, const CartesianVector &point) const
```
AndyChappell, line 250: Same as RemovalBaseTool

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+        if (this->PassElementChecks(testElement, hitType))
+            return false;
+    }
+
+    return true;
+}    
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewCosmicRayRemovalTool::CreateSeed(const MatrixType::Element &element,
```
AndyChappell, line 302: Very similar to CosmicRayRemovalTool. Again appears to be the difference between TensorType and MatrixType that is causing this duplication. Once you've had a chance to confirm if this is driving these differences, we can potentially try to take a look at getting these types to offer a (relatively) common public interface as a way to eliminate this.

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+                
+                break;
+            }
+        }
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+    
+StatusCode TwoViewCosmicRayRemovalTool::GrowSeed(const MatrixType::Element &element,
```
AndyChappell, line 352: As above

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+
+    // Identify remnant hits
+    this->CollectHitsFromDeltaRay(positionOnMuon, muonDirection, pDeltaRayCluster, m_maxDistanceToHit, false, deltaRayHits, remnantHits);
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewCosmicRayRemovalTool::CollectHitsFromDeltaRay(const CartesianVector &positionOnMuon, const CartesianVector &muonDirection, const Cluster *const pDeltaRayCluster,
```
AndyChappell, line 379: As above

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+                
+            collectedHits.push_back(pCaloHit);
+            hitsAdded = true;
+        }
+    }
+} 
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewCosmicRayRemovalTool::SplitDeltaRayCluster(const MatrixType::Element &element, const HitType hitType,
```
AndyChappell, line 412: As above

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+
+    clusterVector.push_back(pMuonCluster); pfoVector.push_back(element.GetOverlapResult().GetCommonMuonPfoList().front());
+    clusterVector.push_back(pDeltaRay); pfoVector.push_back(nullptr);
+
+    m_pParentAlgorithm->UpdateForNewClusters(clusterVector, pfoVector);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewCosmicRayRemovalTool::FragmentRemnant(const HitType hitType, const Cluster *const pMuonCluster,
```
AndyChappell, line 465: Identical to CosmicRayRemovalTool

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewCosmicRayRemovalTool.cc
```diff
@@ -0,0 +1,560 @@
...
+            }
+        }
+
+        clusterVector.push_back(pRemnant); pfoVector.push_back(nullptr);
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode TwoViewCosmicRayRemovalTool::ProjectDeltaRayPositions(const MatrixType::Element &element,
```
AndyChappell, line 501: Very similar to RemovalBaseTool

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayMergeTool.h
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     *  @brief  Determine whether two delta ray clusters are actually a single cluster that is hidden behind a cosmic ray track
+     *
+     *  @param  pMuonPfo the address of the cosmic ray pfo
+     *  @param  pCluster1 the address of one delta ray cluster
+     *  @param  pCluster2 the address of the other delta ray cluster
+     *
+     *  @return  whether the delta ray clusters are one delta ray cluster, hidden behind a cosmic ray track
+     */
+    bool IsHiddenTrack(const pandora::ParticleFlowObject *const pMuonPfo, const pandora::Cluster *const pCluster1, const pandora::Cluster *const pCluster2) const;
```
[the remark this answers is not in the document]
imawby: went with IsHiddenByTrack

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayMergeTool.cc
```diff
@@ -0,0 +1,362 @@
...
+bool DeltaRayMergeTool::Run(ThreeViewDeltaRayMatchingAlgorithm *const pAlgorithm, TensorType &overlapTensor)
+{
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+       std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    bool mergesMade(false);
+
+    this->ExamineConnectedElements(pAlgorithm, overlapTensor, mergesMade);
+
+    return mergesMade;
```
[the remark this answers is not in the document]
imawby: went with single line, just to match the style of the other tools - but happy to change if needed

larpandoracontent/LArThreeDReco/LArCosmicRay/TwoViewAmbiguousDeltaRayTool.h
imawby: implemented

larpandoracontent/LArThreeDReco/LArCosmicRay/RemovalBaseTool.cc
```diff
@@ -0,0 +1,185 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode RemovalBaseTool::ProjectDeltaRayPositions(ThreeViewDeltaRayMatchingAlgorithm *const pAlgorithm, const TensorType::Element &element,
+    const HitType &hitType, CartesianPointVector &projectedPositions) const
+{
+    const Cluster *pCluster1(nullptr), *pCluster2(nullptr);
+    
+    for (const HitType &hitType1 : {TPC_VIEW_U, TPC_VIEW_V, TPC_VIEW_W})
```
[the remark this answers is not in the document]
imawby: so snazzy, will implement

larpandoracontent/LArThreeDReco/LArCosmicRay/CosmicRayRemovalTool.h
```diff
@@ -0,0 +1,129 @@
...
+    bool Run(ThreeViewDeltaRayMatchingAlgorithm *const pAlgorithm, TensorType &overlapTensor);    
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+
+    /**
+     *  @brief  Remove hits from delta ray clusters that belong to the parent muon
+     *
+     *  @param  elementList the tensor element list
+     *  @param  changesMade receive boolean indicating whether clusters in element have been modified
+     */
+    void ExamineConnectedElements(const TensorType::ElementList &elementList, bool &changesMade) const;
```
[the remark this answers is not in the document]
imawby: changed to RemoveCosmicRayHits

larpandoracontent/LArThreeDReco/LArCosmicRay/NViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,1105 @@
...
+            LArPfoHelper::GetClusters(pParentMuon, hitType2, muonClusterList2);
+            
+            if ((muonClusterList1.size() != 1) || (muonClusterList2.size() != 1))
+                return STATUS_CODE_NOT_FOUND;
+            
+            break;
+        }
+
+        break;
+    }
```
AndyChappell, line 578: I think this can use a similar approach to that suggested here: https://github.com/PandoraPFA/LArContent/pull/165#discussion_r616715607

larpandoracontent/LArThreeDReco/LArCosmicRay/NViewDeltaRayMatchingAlgorithm.cc
```diff
@@ -0,0 +1,1105 @@
...
+    float rL(0.f), rT(0.f);
+    slidingFitResult.GetLocalPosition(positionOnMuon, rL, rT);
+    slidingFitResult.GetGlobalFitDirection(rL, muonDirection);
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template<typename T>
```
AndyChappell, line 761: Probably worth moving this one to immediately below the alternative CollectHitsFromMuon function

AndyChappell (commented):
Mini review of remaining matching algorithms.

larpandoracontent/LArObjects/LArTrackOverlapResult.h
```diff
@@ -174,7 +174,9 @@ class TransverseOverlapResult : public TrackOverlapResult
     TransverseOverlapResult &operator=(const TransverseOverlapResult &rhs);
 
 private:
-    XOverlap        m_xOverlap;                     ///< The x overlap object
+    XOverlap          m_xOverlap;                     ///< The x overlap object
+
+    friend class DeltaRayOverlapResult;
```
AndyChappell, line 179: I don't think you need this

larpandoracontent/LArObjects/LArTrackOverlapResult.h
```diff
@@ -343,6 +345,64 @@ class FragmentOverlapResult : public TrackOverlapResult
...
+     *  @brief  Copy constructor
+     *
+     *  @param  rhs
+     */
+    DeltaRayOverlapResult(const DeltaRayOverlapResult &rhs);
+
+    /**
+     *  @brief  Destructor
+     */
+    ~DeltaRayOverlapResult();
```
AndyChappell, line 383: In general, declare destructors as virtual.

larpandoracontent/LArObjects/LArTrackTwoViewOverlapResult.h
```diff
@@ -10,13 +10,122 @@
...
+     *  @brief  Copy constructor
+     *
+     *  @param  rhs
+     */
+    TwoViewDeltaRayOverlapResult(const TwoViewDeltaRayOverlapResult &rhs);
+
+    /**
+     *  @brief  Destructor
+     */
+    ~TwoViewDeltaRayOverlapResult();
```
AndyChappell, line 53: virtual

AndyChappell (commented):
Mini review of various overlap classes, vertex building and the DR parent algorithm

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+            this->PrintHits(cosmicRayHitList, "MC_CR", BLUE);
+            
+            const CartesianVector vertex(pCosmicRay->GetVertex()), endpoint(pCosmicRay->GetEndpoint());
+            const float x0(cosmicRayHitList.front()->GetX0());
+            const CartesianVector shiftedVertex(vertex.GetX() - x0, vertex.GetY(), vertex.GetZ());
+            const CartesianVector shiftedEndpoint(endpoint.GetX() - x0, vertex.GetY(), vertex.GetZ());
+
+            PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &shiftedVertex, "T0 shifted vertex ", BLACK, 2);
+            PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &shiftedEndpoint, "T0 shifted endpoint", BLACK, 2);
+            PandoraMonitoringApi::ViewEvent(this->GetPandora());
```
AndyChappell, line 340: Need the macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+                this->PrintHits(leadingParticleHitList, "MC_DR", RED);
+
+                const CartesianVector vertex(pLeadingParticle->GetVertex()), endpoint(pLeadingParticle->GetEndpoint());
+                const float x0(leadingParticleHitList.front()->GetX0());
+                const CartesianVector shiftedVertex(vertex.GetX() - x0, vertex.GetY(), vertex.GetZ());
+                const CartesianVector shiftedEndpoint(endpoint.GetX() - x0, vertex.GetY(), vertex.GetZ());
+
+                PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &shiftedVertex, "T0 shifted vertex", BLACK, 2);
+                PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &shiftedEndpoint, "T0 shifted endpoint", BLACK, 2);
+                PandoraMonitoringApi::ViewEvent(this->GetPandora());
```
AndyChappell, line 397: Macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+                    {
+                        PfoList deltaRayList({pMatchedPfo});
+                        PandoraMonitoringApi::VisualizeParticleFlowObjects(this->GetPandora(), &deltaRayList, "Delta Ray Pfo", BLACK, false, false);
+
+                        PfoList cosmicList({pParentPfo});
+                        PandoraMonitoringApi::VisualizeParticleFlowObjects(this->GetPandora(), &cosmicList, "Delta Ray Pfo", RED, false, false);
+                        
+                        const CartesianVector &vertex(deltaRayVertexList.front()->GetPosition());
+                        PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &vertex, "vertex", RED, 2);
+                        PandoraMonitoringApi::ViewEvent(this->GetPandora());
```
AndyChappell, line 589: Macro wrapper around the PandoraMonitoringApi calls

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void MuonLeadingEventValidationAlgorithm::PrintHits(const CaloHitList caloHitList, const std::string &stringTag, const Color &colour) const
+{
+    for (const CaloHit *const pCaloHit : caloHitList)
+    {
+        const CartesianVector hitPosition(pCaloHit->GetPositionVector().GetX() - pCaloHit->GetX0(), pCaloHit->GetPositionVector().GetY(), pCaloHit->GetPositionVector().GetZ());
+
+        PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &hitPosition, stringTag, colour, 2);
```
AndyChappell, line 735: Macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+    }
+
+    PandoraMonitoringApi::ViewEvent(this->GetPandora());
```
AndyChappell, line 738: Macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+            color = RED;
+        }
+
+        if (std::find(parentTrackCaloHitList.begin(), parentTrackCaloHitList.end(), pCaloHit) != parentTrackCaloHitList.end())
+	    {
+            newStringTag += "_PARENT_TRACK"; 
+            color = BLUE;
+        }
+
+        PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &hitPosition, newStringTag, color, 2);
```
AndyChappell, line 770: Macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+        if (std::find(parentTrackCaloHitList.begin(), parentTrackCaloHitList.end(), pCaloHit) != parentTrackCaloHitList.end())
+	    {
+            newStringTag += "_PARENT_TRACK"; 
+            color = BLUE;
+        }
+
+        PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &hitPosition, newStringTag, color, 2);    
+    }    
+    
+    PandoraMonitoringApi::ViewEvent(this->GetPandora());
```
AndyChappell, line 773: Macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+        std::string newStringTag(stringTag);
+        const CartesianVector hitPosition(pCaloHit->GetPositionVector().GetX() - pCaloHit->GetX0(), pCaloHit->GetPositionVector().GetY(), pCaloHit->GetPositionVector().GetZ());
+        
+        if (std::find(leadingCaloHitList.begin(), leadingCaloHitList.end(), pCaloHit) != leadingCaloHitList.end())
+	    {
+            newStringTag += "_LEADING"; 
+            color = RED;
+        }
+
+        PandoraMonitoringApi::AddMarkerToVisualization(this->GetPandora(), &hitPosition, newStringTag, color, 2);
```
AndyChappell, line 793: Macro wrapper

larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -0,0 +1,889 @@
...
+    }
+
+    PandoraMonitoringApi::ViewEvent(this->GetPandora());
```
AndyChappell, line 796: Macro wrapper

AndyChappell (commented):
Monitoring related review

larpandoracontent/LArHelpers/LArMuonLeadingHelper.cc
```diff
@@ -0,0 +1,524 @@
...
+}     
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+const MCParticle *LArMuonLeadingHelper::GetLeadingParticle(const MCParticle *const pMCParticle)
+{
+    MCParticleVector mcParticleVector;
+
+    const MCParticle *pParentMCParticle = pMCParticle;
+    mcParticleVector.push_back(pParentMCParticle);
```
AndyChappell, line 88: Do we need this vector? I think if you get beyond the if statement in the first loop iteration, you're guaranteed to find a leading particle to assign to pParentMCParticle, so maybe you can just set a found flag and simply return the pointer if found, throw an exception if not?

larpandoracontent/LArHelpers/LArMuonLeadingHelper.cc
```diff
@@ -0,0 +1,524 @@
...
+                {
+                    bestTargetWeight = targetWeight;
+                }
+            }
+
+            if ((targetWeightSum < std::numeric_limits<float>::epsilon()) || ((bestTargetWeight / targetWeightSum) < minHitSharingFraction))
+                continue;
+
+            // Remove and record post bremsstrahlung hits
+            if (RejectBremsstrahlungHits(pCaloHit, leadingMCParticleToPostBremsstrahlungHitList))
```
AndyChappell, line 225: LArMuonLeadingHelper::Reject...

larpandoracontent/LArHelpers/LArMuonLeadingHelper.cc
```diff
@@ -0,0 +1,524 @@
...
+    {
+        if (pCaloHit->GetHitType() == tpcView)
+            postBremsstrahlungViewHitList.push_back(pCaloHit);
+    }
+
+    if (postBremsstrahlungViewHitList.empty())
+        return;
+
+    bool hitsAdded(true);
+    while (hitsAdded)
```
AndyChappell, line 316: do {} while ()

AndyChappell (commented):
Helper classes review

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -172,11 +246,13 @@ inline pandora::StatusCode LArMCParticleFactory::Write(const Object *const pObje
     {
         pandora::BinaryFileWriter &binaryFileWriter(dynamic_cast<pandora::BinaryFileWriter&>(fileWriter));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(pLArMCParticle->GetNuanceCode()));
+        PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileWriter.WriteVariable(static_cast<int>(pLArMCParticle->GetProcess())));
```
AndyChappell, line 249: Should probably add version checking to the output too, if we're rewriting from an old pndr file for some reason, the process still won't exist

AndyChappell (commented):
We'll also want a PR for the corresponding larpandora update once we're ready to go with this, and we'll need to coordinate the releases to ensure they are updated together.

larpandoracontent/LArObjects/LArMCParticle.h
AndyChappell: And a brief request to @johnmarshall80 to take a look at the LArMCParticle.h changes in particular, given I had input into the changes to this file, having an additional set of eyes for the review would be useful. Thanks!

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -21,13 +21,46 @@
 namespace lar_content
 {
 
+// ATTN - Enumeration replicates the numbering in the CellTree module and thus has a dummy UNUSED enumeration at 15 to account for an used
+// duplication in the CellTree process map
+// We also add an INCIDENT_NU to assign a process for neutrino MC given that we don't have a distinct class for neutrino MC
```
johnmarshall80, line 26: Wonder if this comment can be made a bit more self-explanatory, and give slightly more general context? Aim to keep very brief and easy to maintain in future, of course (comments otherwise can "go off"!).

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -94,14 +142,19 @@ class LArMCParticleFactory : public pandora::ObjectFactory<object_creation::MCPa
      *  @param  pObject to receive the address of the object created
      */
     pandora::StatusCode Create(const Parameters &parameters, const Object *&pObject) const;
+
+private:
+
```
johnmarshall80, line 147: Suggest remove empty line.

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayRemovalTool.cc
```diff
@@ -0,0 +1,232 @@
...
+
+        this->ExamineConnectedElements(elementList, changesMade);
+    }
+    
+    return changesMade;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void DeltaRayRemovalTool::ExamineConnectedElements(const TensorType::ElementList &elementList,
```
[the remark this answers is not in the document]
imawby: changed, also realised that the variable 'changes made' was being set wrt the last iteration of the for loop in Run (i.e. with respect to the final key cluster). This is not intended - should be whether any changes at all were made. Have changed this

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayRemovalTool.cc
```diff
@@ -0,0 +1,232 @@
...
+    {
+        for (const HitType hitType : {TPC_VIEW_U, TPC_VIEW_V, TPC_VIEW_W})
+	    {
+            const Cluster *pDeltaRayCluster(element.GetCluster(hitType));
+            const ParticleFlowObject *const pMuonPfo(element.GetOverlapResult().GetCommonMuonPfoList().front());
+
+            if (checkedClusters.count(pDeltaRayCluster))
+                continue;
+            
+            if ((modifiedClusters.count(element.GetClusterU())) || (modifiedClusters.count(element.GetClusterV())) || (modifiedClusters.count(element.GetClusterW())))
```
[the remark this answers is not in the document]
imawby: resolved on slack - clusters from all views will be in the modifiedClusters set

larpandoracontent/LArThreeDReco/LArCosmicRay/DeltaRayRemovalTool.cc
imawby: i think i'll do the latter since the boolean seems a bit far removed to be a parameter to PassElementChecks

larpandoracontent/LArThreeDReco/LArCosmicRay/NViewDeltaRayMatchingAlgorithm.h
```diff
@@ -0,0 +1,366 @@
...
+     *  @param  clusterW the W CaloHitList
+     *  @param  chiSquaredSum the sum of the chi-squared values of the sampled points
+     *  @param  nSamplingPoints the number of sampled points
+     *  @param  nMatchedSamplingPoints the number of matched sampled points
+     *  @param  xOverlap the xOverlap object  
+     *
+     *  @return  a status code reflecting whether the matching procedure ran smoothly and if the outcome is good 
+     */
+    pandora::StatusCode PerformThreeViewMatching(const pandora::CaloHitList &clusterU, const pandora::CaloHitList &clusterV, const pandora::CaloHitList &clusterW,
+        float &chiSquaredSum, unsigned int &nSamplingPoints, unsigned int &nMatchedSamplingPoints, XOverlap &XOverlap) const;
```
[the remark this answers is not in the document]
imawby: these algorithms are also used in the TwoViewMatchingAlgorithm to perform matching between the two views of input clusters and 'matched clusters' in the third view

larpandoracontent/LArObjects/LArTrackOverlapResult.h
imawby: you are correct

larpandoracontent/LArHelpers/LArPfoHelper.h
```diff
@@ -184,7 +184,7 @@ class LArPfoHelper
      *  @param  pPfo the address of the first Pfo
      *  @param  pPfo the address of the second Pfo
      */
```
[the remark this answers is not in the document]
imawby: function has been moved to the DeltaRayParent algorithm which is the only use case. Functionality has been kept and function description updated

larpandoracontent/LArHelpers/LArMuonLeadingHelper.cc
```diff
@@ -0,0 +1,524 @@
...
+    return ((LArMCParticleHelper::GetHierarchyTier(pMCParticle) == 1) && (LArMCParticleHelper::IsCosmicRay(pParentMCParticle)));
+}     
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+const MCParticle *LArMuonLeadingHelper::GetLeadingParticle(const MCParticle *const pMCParticle)
+{
+    MCParticleVector mcParticleVector;
+
+    const MCParticle *pParentMCParticle = pMCParticle;
```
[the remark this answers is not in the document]
imawby: it seems like it to me? but i'll add in a throw just to be safe

larpandoracontent/LArHelpers/LArMuonLeadingHelper.cc
imawby: have changed this to:

`const MCParticle *LArMuonLeadingHelper::GetLeadingParticle(const MCParticle *const pMCParticle)
{
    if (!pMCParticle)

    if (LArMCParticleHelper::GetHierarchyTier(pMCParticle) == 0)
        throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);
        
    const MCParticle *pParentMCParticle = pMCParticle;

    while (LArMCParticleHelper::GetHierarchyTier(pParentMCParticle) != 1))
    {
        const MCParticleList &parentList(pParentMCParticle->GetParentList());
        
        if (parentList.size() != 1)
            throw StatusCodeException(STATUS_CODE_INVALID_PARAMETER);

        pParentMCParticle = parentList.front();
    }

    return pParentMCParticle;
} `

imawby: the idea being to find the address of the MCParticle below the cosmic ray muon in the hierarchy

AndyChappell (approved):

PR #167, opened 2021-04-21
Feature/two view merging tool updated clean

MariaBrigida:
This PR introduces three new tools for two-view 2D->3D calorimetric matching, as detailed below:

- TwoViewUndershootTracksTool, to merge clusters that calorimetrically match same cluster in other view, if there is no kink
- TwoViewOvershootTracksTool, to split clusters that calorimetrically match separate clusters in other view, if there is a kink
- TwoViewThreeDKinkBaseTool, base class for the above two tools

These are adaptations for the two-view calorimetric matching of the TransverseTrackMatching tools with the same names.

LArContent is edited accordingly, and typos in LArOverlapMatrix documentation are corrected.


johnmarshall80 (commented):
Hi @MariaBrigida - thanks very much for this PR. I have some tentative thoughts/suggestions we can dig into:

It's obviously very similar to the three view case, to the extent that the code actually looks ready to go in and it should operate effectively right away. It may not, however, be minimal and perfect by doing that, though! All suggestions below are really about making it minimal.

It occurs to me that, with two views, we actually no longer have undershoots and overshoots, as those names refer to the cases where we have:

- Undershoot: two long clusters (one in each of two views) and two shorter clusters (both in a third view)
- Overshoot: two pairs of shorter clusters (two in each of two views) and one long cluster (in a third view)

Now we only ever have two shorter clusters (in one view) and one longer cluster (in the other view), I think? In that case, I suggest removing the Overshoot tool, and merging the Undershoot tool methods back into the kink base tool, which we then just call TwoViewThreeDKink tool?

The number of common clusters would have to be one for two views, and the comments about majority wins also disappear (there's no majority with only two views). We should probably keep the split mode option, but leave it on by default, so that the one tool can make merges or splits to deal with both over and undershoots (it splits if it thinks there's a 3D kink, merges if thinks there is no 3D kink).

After that, I think there's just some stylistic bits about comments wrapping over multiple lines and doxygen formatting, and we're done.

One thing to add to the above, is that there is _potential_ for the current overshoot and undershoot tools to do something slightly different, but (conceptually) they ought to be doing the same thing. I've arbitrarily suggested basing the single tool off the undershoot version at the moment, because it's a bit simpler. But, we should be able to add a few config options to our single tool to allow us to cover all the parameter space currently covered.

So, suggestions: rename TwoViewThreeDKinkBaseTool to TwoViewThreeDKinkTool. Merge in the functions it used to call in the derived TwoViewUndershootTracksTool into this single class, remove the unused member variables noted above, remove wrapping comments.

Thanks again!

MariaBrigida:
Thanks for the suggestions @johnmarshall80! Here's a new version, where the TwoViewThreeDKinkTool replaces the previous triplet of tools, the majorityRulesMode is removed, and only one common cluster is allowed. The displayed pfos looks identical between the two versions when testing on a small sample.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
+/**
+ *  @file   larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
+ *
+ *  @brief  Implementation of the three d kink base tool class.
```
johnmarshall80, line 4: Can update this.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewThreeDKinkTool::PassesElementCuts(MatrixType::ElementList::const_iterator eIter, const ClusterSet &usedClusters) const
+{
+    if (usedClusters.count(eIter->GetCluster1()) || usedClusters.count(eIter->GetCluster2()))
+        return false;
+
+    if (eIter->GetOverlapResult().GetTwoViewXOverlap().GetXOverlapFraction0() - m_minXOverlapFraction < -1.f * std::numeric_limits<float>::epsilon())
+        return false;
```
johnmarshall80, line 65: Suggest empty line here.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
...
+                    modification.m_clusterMergeMap[pLowXCluster].push_back(pHighXCluster);
+                }
+
+                modification.m_affectedClusters.push_back(particle.m_pClusterA);
+                modification.m_affectedClusters.push_back(particle.m_pClusterB);
+                modification.m_affectedClusters.push_back(particle.m_pCommonCluster);
+
+                modificationList.push_back(modification);
+            }
+            catch (StatusCodeException &statusCodeException)
```
johnmarshall80, line 359: This can be const, although suspect it isn't (yet) in three-view equivalent!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
...
+        if (dotProduct < m_cosThetaCutForKinkSearch)
+            return true;
+    }
+    catch (StatusCodeException &s)
+    {
+    }
+
+    return false;
+}
+
```
johnmarshall80, line 425: Suggest remove two lines here.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
...
+        // Apply final cuts
+        const CartesianVector minusToSplit((split - minus).GetUnitVector());
+        const CartesianVector splitToPlus((plus - split).GetUnitVector());
+        const float dotProduct(minusToSplit.GetDotProduct(splitToPlus));
+
+        if (dotProduct < m_cosThetaCutForKinkSearch)
+            return true;
+    }
+    catch (StatusCodeException &s)
```
johnmarshall80, line 419: Can also catch by const reference here. Suggest remove variable `s` if unused.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
```diff
@@ -0,0 +1,172 @@
+/**
+ *  @file   larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
+ *
+ *  @brief  Header file for the two view three d kink base tool
```
johnmarshall80, line 4: Can update name in brief.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
```diff
@@ -0,0 +1,172 @@
...
+         *  @param  elementB the matrix element B
+         */
+        Particle(const MatrixType::Element &elementA, const MatrixType::Element &elementB);
+
+        const pandora::Cluster *m_pClusterA;      ///< Address of non-shared cluster in element A
+        const pandora::Cluster *m_pClusterB;      ///< Address of non-shared cluster in element B
+        const pandora::Cluster *m_pCommonCluster; ///< Address of the common cluster
+    };
+
+    void GetIteratorListModifications(
```
johnmarshall80, line 136: Does this now need the doxygen description that used to live in the base tool?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
```diff
@@ -0,0 +1,172 @@
...
+    float m_minMatchingScore;               ///< The min global matching score for particle creation
+    float m_minLocallyMatchedFraction;      ///< The min locally matched fraction for particle creation
+    float m_minLongitudinalImpactParameter; ///< The min longitudinal impact parameter for connecting accompanying clusters
+    int m_nLayersForKinkSearch;             ///< The number of sliding fit layers to step in the kink search
+    float m_additionalXStepForKinkSearch;   ///< An additional (safety) step to tack-on when choosing x sampling
+    bool m_splitMode;                       ///< Whether to run in cluster splitting mode, as opposed to cluster merging mode
+    float m_maxTransverseImpactParameter;   ///< The maximum transverse impact parameter for connecting broken clusters
+    float m_minImpactParameterCosTheta;     ///< The minimum cos theta (angle between vertex directions) for connecting broken clusters
+    float m_cosThetaCutForKinkSearch;       ///< The cos theta cut used for the kink search in three dimensions
+
```
johnmarshall80, line 166: Can remove empty lines.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
...
+                const CartesianVector splitPosition((vertexA.GetPosition() + vertexB.GetPosition()) * 0.5f);
+                const bool isThreeDKink(this->IsThreeDKink(pAlgorithm, particle, splitPosition, isALowestInX));
+
+                if (isThreeDKink != m_splitMode)
+                    continue;
+
+                // Construct the modification object
+                Modification modification;
+
+                if (m_splitMode)
```
johnmarshall80, line 333: Biggest change: would you agree with removing the split mode entirely, and replacing the `if (m_splitMode)` with `if (isThreeDKink) {...} else {...}`. In that way, the 3D kink finding always drives whether we split (if it's kinked) or merge (if it's not). Can you test that? Thanks!

johnmarshall80 (commented):
Hi @MariaBrigida - many thanks for the changes. I think this looks good, and have just one suggestion for structural change, and a few cosmetic changes. The structural change would be to try testing without the split mode logic, and just make this run dynamically, with splits for identified 3D kinks, merges for cases where no 3D kink identified. Thanks!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
MariaBrigida: That's the case. Should I change it there as well, or leave it for a later revision?

johnmarshall80: Please feel free to change this wherever it's applicable. Thanks!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
```diff
@@ -0,0 +1,172 @@
...
+     *  @brief  Constructor
+     *
+     *  @param  nCommonClusters the number of common clusters to select
+     */
+    TwoViewThreeDKinkTool();
+
+    /**
+     *  @brief  Destructor
+     */
+    ~TwoViewThreeDKinkTool();
```
AndyChappell, line 34: Declare as virtual.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
```diff
@@ -0,0 +1,172 @@
...
+     *
+     *  @param  eIter iterator to a candidate element
+     *  @param  elementList the provided element list
+     *  @param  usedClusters the list of used clusters
+     *  @param  iteratorList to receive a list of iterators to relevant elements
+     */
+    void SelectMatrixElements(MatrixType::ElementList::const_iterator eIter, const MatrixType::ElementList &elementList,
+        const pandora::ClusterSet &usedClusters, IteratorList &iteratorList) const;
+
+    /**
```
AndyChappell, line 117: Minor point, but I would keep the inner class definitions together.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,480 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TwoViewThreeDKinkTool::PassesElementCuts(MatrixType::ElementList::const_iterator eIter, const ClusterSet &usedClusters) const
+{
+    if (usedClusters.count(eIter->GetCluster1()) || usedClusters.count(eIter->GetCluster2()))
+        return false;
+
+    if (eIter->GetOverlapResult().GetTwoViewXOverlap().GetXOverlapFraction0() - m_minXOverlapFraction < -1.f * std::numeric_limits<float>::epsilon())
```
AndyChappell, line 64: Hopefully the compiler handles this anyway, but `-std::numeric_limits<float>::epsilon()` will ensure the multiplication isn't actually performed at runtime (here and elsewhere)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
MariaBrigida: Does this need to be declared virtual, now that TwoViewThreeDKinkTool has no derived classes in the current implementation?

AndyChappell: Yes, you still want it, as it's always possible for future updates to inherit from this class, and this class already inherits from another. In general, all destructors should be tagged as virtual, this ensures that the correct destructor is called on an object however it is being accessed at any given time. If you don't declare the destructor virtual you can end up in situations where a base class destructor gets called instead and so the derived object may not be completely destroyed.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
MariaBrigida: Happy to test the version you suggest! Is it okay to just throw an exception in case the following check fails, in that case?
https://github.com/PandoraPFA/LArContent/pull/167/files#diff-fd01ba767797964c499ddec52b36d00b1251aee746f209ab4e217ced8d215a0eR392

johnmarshall80: Hi, I think the options are pick a default operation (merge or split) in the case where the kink finding algorithm doesn't feel it can provide a decision, or make it skip over this modification entirely. This could be by raising an exception at that point, but I think you'll want to catch and continue back in the original calling function, rather than risk bailing the whole processing of this algorithm for this event. Picking a default split/merge doesn't sound so nice, but at least it means the tool will have responded to a problem identified in the matrix. It will have resolved an ambiguity, but perhaps not in the optimal way. Suggest you have a play about and see what you think is best. Thanks!

MariaBrigida: Okay, great, thanks!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,479 @@
...
+
+    CartesianVector minus(0.f, 0.f, 0.f), plus(0.f, 0.f, 0.f);
+    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, fitResult1.GetGlobalFitPosition(fitResult1.GetL(lowLayer), minus));
+    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, fitResult1.GetGlobalFitPosition(fitResult1.GetL(highLayer), plus));
+
+    if (minus.GetX() > plus.GetX())
+    {
+        CartesianVector temporary(minus);
+        minus = plus;
+        plus = temporary;
```
AndyChappell, line 114: There's std::swap(minus, plus) function that will handle this functionality for you

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,479 @@
...
+        (pointingClusterA.GetInnerVertex().GetPosition().GetX() < pointingClusterB.GetOuterVertex().GetPosition().GetX()))
+    {
+        return true;
+    }
+
+    if ((pointingClusterA.GetOuterVertex().GetPosition().GetX() < pointingClusterB.GetInnerVertex().GetPosition().GetX()) &&
+        (pointingClusterA.GetOuterVertex().GetPosition().GetX() < pointingClusterB.GetOuterVertex().GetPosition().GetX()))
+    {
+        return true;
+    }
```
AndyChappell, line 139: No strong view on this, but I wonder if this can be made a little more concise via:

```
const float xMinA{std::min(pointingClusterA.GetInnerVertex().GetPosition().GetX(), pointingClusterA.GetOuterVertex().GetPosition().GetX())};
const float xMinB{std::min(pointingClusterB.GetInnerVertex().GetPosition().GetX(), pointingClusterB.GetOuterVertex().GetPosition().GetX())};
return xMinA < xMinB;
```

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
johnmarshall80: It's still [here](https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArThreeDReco/LArTransverseTrackMatching/ThreeDKinkBaseTool.cc#L90-L95) too, but we do probably want to leave the three-view cases alone in this PR.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
```diff
@@ -0,0 +1,177 @@
...
+     */
+    virtual ~TwoViewThreeDKinkTool();
+
+    bool Run(TwoViewTransverseTracksAlgorithm *const pAlgorithm, MatrixType &overlapMatrix);
+
+private:
+    /**
+     *  @brief  Modification class
+     */
+    class Modification
```
AndyChappell, line 42: Minor point, but I would make this a struct, which has its members public by default and probably fits the intent a little better.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,479 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TwoViewThreeDKinkTool::GetIteratorListModifications(
+    TwoViewTransverseTracksAlgorithm *const pAlgorithm, const IteratorList &iteratorList, ModificationList &modificationList) const
+{
+    for (IteratorList::const_iterator iIter1 = iteratorList.begin(), iIter1End = iteratorList.end(); iIter1 != iIter1End; ++iIter1)
+    {
+        for (IteratorList::const_iterator iIter2 = iIter1; iIter2 != iIter1End; ++iIter2)
```
AndyChappell, line 288: `iIter2 = std::next(iIter1)` avoids the need for the equality check below

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,479 @@
...
+        {
+            return m_splitMode; // split mode rules, by default
+        }
+
+        // Extract results
+        const HitType hitType1(LArClusterHelper::GetClusterHitType(particle.m_pCommonCluster));
+        const HitType hitType2(LArClusterHelper::GetClusterHitType(particle.m_pClusterA));
+
+        CartesianVector minus(0.f, 0.f, 0.f), split(0.f, 0.f, 0.f), plus(0.f, 0.f, 0.f);
+        float chi2Minus(std::numeric_limits<float>::max()), chi2Split(std::numeric_limits<float>::max()),
```
AndyChappell, line 406: Given you don't use the chi2 values, could just define one dummyChi2 variable and reuse that in all three cases.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
MariaBrigida: I assume it would be the same for the Particle class. I am fine with it, although I think sub-classes are used in this way elsewhere in Pandora. Is this something we plan to implement in all similar cases in the future?

AndyChappell: I don't have a strong view on this (I just think it conveys the intent more clearly), where there are new classes that are just acting as containers for variables that will be publicly accessible, I would make those structs and continue to do so in the future, but I wouldn't go back and replace historical classes of this form with structs.

johnmarshall80: Hi @MariaBrigida, we’ve previously stuck to classes across the board, rather than using structs in our codebase. Doesn’t hurt to have to make it explicit that the members are public and occasionally helps avoid confusion. Take the point that @AndyChappell is making (about intent and suitability), but counter that it’s better to just remain consistent here. Thanks.

MariaBrigida:
I hopefully addressed all comments. Will do some more testing with/without split mode.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,451 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+float TwoViewThreeDKinkTool::GetXSamplingPoint(const CartesianVector &splitPosition1, const bool isForwardInX,
+    const TwoDSlidingFitResult &fitResult1, const TwoDSlidingFitResult &fitResult2) const
+{
+    // Nearest common x position
+    float xMin1(std::numeric_limits<float>::max()), xMax1(-std::numeric_limits<float>::max());
+    float xMin2(std::numeric_limits<float>::max()), xMax2(-std::numeric_limits<float>::max());
```
absolution1, line 85: would `std::numeric_limits<float>::lowest()`, rather than `-std::numeric_limits<float>::max()` be appropriate here?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,451 @@
...
+
+bool TwoViewThreeDKinkTool::PassesElementCuts(MatrixType::ElementList::const_iterator eIter, const ClusterSet &usedClusters) const
+{
+    if (usedClusters.count(eIter->GetCluster1()) || usedClusters.count(eIter->GetCluster2()))
+        return false;
+
+    if (eIter->GetOverlapResult().GetTwoViewXOverlap().GetXOverlapFraction0() - m_minXOverlapFraction < -std::numeric_limits<float>::epsilon())
+        return false;
+
+    if (eIter->GetOverlapResult().GetTwoViewXOverlap().GetXOverlapFraction1() - m_minXOverlapFraction < -std::numeric_limits<float>::epsilon())
```
absolution1, line 66: How come these two use `-std::numeric_limits<float>::epsilon()` and the other checks drop the minus sign?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
```diff
@@ -0,0 +1,451 @@
...
+                float transverseAB(std::numeric_limits<float>::max()), transverseBA(std::numeric_limits<float>::max());
+                float longitudinalAB(-std::numeric_limits<float>::max()), longitudinalBA(-std::numeric_limits<float>::max());
+
+                LArPointingClusterHelper::GetImpactParameters(vertexA, vertexB, longitudinalAB, transverseAB);
+                LArPointingClusterHelper::GetImpactParameters(vertexB, vertexA, longitudinalBA, transverseBA);
+
+                if (std::min(longitudinalAB, longitudinalBA) < m_minLongitudinalImpactParameter)
+                    continue;
+
+                if (std::min(transverseAB, transverseBA) > m_maxTransverseImpactParameter)
```
absolution1, line 298: Should this one be `std::max(args)` ?

absolution1 (commented):
Looks nice @MariaBrigida 
Literally 2 clarification points and one std::numeric_limits suggestion.

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.h
absolution1: Purely because this thread is about struct vs classes, the c++ guidelines suggest ways to use struct vs class for implying intent: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rc-org

Not that this should trump consistency across the codebase...

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
MariaBrigida: I think you're right :)

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
MariaBrigida: Ah, these cuts are the same as those implemented in the TwoViewClearTracksTool. Not sure why. I would lean to adding a minus sign to the following two comparisons too, what do you reckon?

johnmarshall80: No strong preference from me, although would make sense for the two tools mentioned above to be consistent.

(Only a small snippet, so can live with separate implementations for now, unless there's a _very_ simple way to share implementation.)

absolution1: probably means I included the minus sign then.  Honestly no idea why...

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
MariaBrigida: So, here as well I followed the same philosophy of an existing tool (the UndershootTracksTool, [here](https://github.com/PandoraPFA/LArContent/blob/master/larpandoracontent/LArThreeDReco/LArTransverseTrackMatching/UndershootTracksTool.cc#L68)), but I do see your point that max is a stronger condition and more appropriate. I don't know if the original one was a choice or a bug!

MariaBrigida: Yes you're right - the logic here also comes from an existing tool (the UndershootTracksTool [here](So, here as well I followed the same logic of an existing tool (the UndershootTracksTool, here), but I do see your point that max is probably more appropriate. I don't know if the original one was a choice or a bug!)) but as you point out I think it was probably a bug there?

MariaBrigida: (I managed to write 3 copies of this comment before it came out as I intended it to... :D)

johnmarshall80: Hi, I think this is correct as is, and correct in the original tools. It's taking the smallest available impact parameter, and then checking to see if that value is still bigger than some threshold check. So pretty sure we want to keep this intact (unless want to start up some explicit study to check).

MariaBrigida: Does this not allow for the case when the minimum is smaller than m_maxTransverseImpactParameter, but the maximum isn't? (It may be that things are getting confused in my head)

johnmarshall80: It's been a while now(!), but I think it's looking at the consistency of vertex A and B in two ways. One using a position extrapolated to B, based on the position and direction estimated at A; the other using a position extrapolated to A, based on the position and direction estimated at B. Good agreement (small values) either way around would indicate some kind of association. So, we take the smallest values (representing the most plausible agreement, we think) and then we apply quality thresholds to those values. Would need to start drawing some diagrams out, and thinking through the surrounding logic too, to completely refamiliarise with the issues at stake. Hope that helps!

MariaBrigida: I see, thanks John - I thought this condition was meant to ensure that both transverseAB and transverseBA were below the m_maxTransverseImpactParameter threshold, just like the condition above ensures that longitudinalAB and longitudinalBA are larger than m_minLongitudinalImpactParameter. So I was under the wrong assumption on the intention behind the checks.

johnmarshall80: Yes, sometimes the only thing you can do here is draw out an example event, think through the logic you want and check that the code is in agreement. Here, can imagine a 2D kink topology, with two tracks in something like a v-shape, but a bit of a gap between the track endpoints (i.e. a chunk missing from one of the tracks).

Extrapolating one of the tracks will successfully allow you to reach the vertex of the second track, but extrapolating the other way around will not do the same thing. We use the successful extrapolation here to say that the tracks might be associated, but then apply some quality cuts to that possible association.

MariaBrigida: Yes, I see, that makes sense! Thanks for the clarifications!

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
MariaBrigida: It was most likely me, @absolution1!
As for @johnmarshall80's suggestion to share implementation, could maybe think of adding a PassesElementCuts method to TransverseMatrixTool, that some (but not all) of the derived tools would make use of?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
absolution1: I am little confused here.
With the way this is configured now, only one of the A,B clusters has to look like it passes in close proximity to the other cluster?  Do you not need the double coincidence to know that the two clusters actually do align?

larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
absolution1: I'd advocate for not tying the two tools together in this way.  I can imagine that we may want to update the implementations in a non-uniform way in the future and having them both be tied to each other just makes that procedure more difficult.
If there were huge chunks of the tools that were boilerplated then, sure, but we're talking about single functions that contain ~5 if statements, right?

PR #169, opened 2021-05-10
Updates to MVA vertex selection algorithms.

henrylay97:
This PR includes a series of updates/improvements to the MVA vertex selection procedure. This includes the fix to issue #157, the addition of a new feature tool for calculating the `EnergyDepositionAsymmetry` and provides functionality to use features "shared" between the two vertex candidates. All operational changes are controlled by xml-parameters so the PR shouldn't have any affect without opting to move out of "legacy" mode.

There is a large amount of code here but the separate commits should represent distinct ideas, I hope this makes it easier for reviewing! 
- [b6d1cd](https://github.com/henrylay97/LArContent/commit/b6d1cd27ba0e46d83caf0969b6349980b1888760) adds the new `EnergyDepositionAsymmetryTool` 
- [9e3c99](https://github.com/henrylay97/LArContent/commit/9e3c99a4a087cd6d82206ae60021e989fe7b837c) adds the functionality that calculates the `VertexEnergy` variables
- [66f9c2](https://github.com/henrylay97/LArContent/commit/66f9c2cb4fc7c21e39238556fdff7276e205d9c9) adds the functionality for the shared variables and [93d5ec](https://github.com/henrylay97/LArContent/commit/93d5ec75e64939dae88e646c7e1bc20baa19f46a) adds the new object based on `VertexFeatureInfo` and `EventFeatureInfo` to store the shared variables.
- [d94d24](https://github.com/henrylay97/LArContent/commit/d94d2422a3774f7edd420ad2060114721706f4d9) fixes the event shape features, default is to use the old calculation still, [5b028c](https://github.com/henrylay97/LArContent/commit/5b028c0c561e63aac1fa73d70b605c85b7f8fb02) adds a warning message if using the old calculation whilst in training mode.
- [50008d](https://github.com/henrylay97/LArContent/commit/50008debb6856f01df45a7b1993ebbd88be712ea) actually uses all this in the configuration I presented last week. Again the default mode is to use the old configuration.
- [ddf431](https://github.com/henrylay97/LArContent/commit/ddf43161ad38b66fb5aca0da94ed9c98490a2ed9) and [9258e3](https://github.com/henrylay97/LArContent/commit/9258e35286b93246cc16dc63f084de8fd7333d53) tidy up the commenting and apply the clang format.

### DIFF
--- a/larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc
+++ b/larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc
@@ -63,9 +63,8 @@ void AsymmetryFeatureBaseTool::IncrementAsymmetryParameters(
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-template <typename C, typename T = typename C::value_type>
 float AsymmetryFeatureBaseTool::CalculateAsymmetry(const bool useEnergyMetrics, const CartesianVector &vertexPosition2D,
-    const C &asymmetryClusters, const CartesianVector &localWeightedDirectionSum) const
+    const ClusterVector &asymmetryClusters, const CartesianVector &localWeightedDirectionSum) const
 {
     // Project every hit onto local event axis direction and record side of the projected vtx position on which it falls
     float beforeVtxHitEnergy(0.f), afterVtxHitEnergy(0.f);
@@ -120,13 +119,5 @@ StatusCode AsymmetryFeatureBaseTool::ReadSettings(const TiXmlHandle xmlHandle)
 
     return STATUS_CODE_SUCCESS;
 }
-
-//------------------------------------------------------------------------------------------------------------------------------------------
-
-template float AsymmetryFeatureBaseTool::CalculateAsymmetry(const bool useEnergyMetrics, const CartesianVector &vertexPosition2D,
-    const ClusterVector &asymmetryClusters, const CartesianVector &localWeightedDirectionSum) const;
-
-template float AsymmetryFeatureBaseTool::CalculateAsymmetry(const bool useEnergyMetrics, const CartesianVector &vertexPosition2D,
-    const ClusterList &asymmetryClusters, const CartesianVector &localWeightedDirectionSum) const;
 
 } // namespace lar_content
--- a/larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc
+++ b/larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc
@@ -1,0 +1,132 @@
+/**
+ *  @file   larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc
+ *
+ *  @brief  Implementation of the  asymmetry feature tool class.
+ *
+ *  $Log: $
+ */
+
+#include "larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.h"
+#include "Pandora/AlgorithmHeaders.h"
+#include "larpandoracontent/LArHelpers/LArClusterHelper.h"
+#include "larpandoracontent/LArHelpers/LArGeometryHelper.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+AsymmetryFeatureBaseTool::AsymmetryFeatureBaseTool() : m_maxAsymmetryDistance(5.f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void AsymmetryFeatureBaseTool::Run(LArMvaHelper::MvaFeatureVector &featureVector, const VertexSelectionBaseAlgorithm *const pAlgorithm,
+    const Vertex *const pVertex, const VertexSelectionBaseAlgorithm::SlidingFitDataListMap &slidingFitDataListMap,
+    const VertexSelectionBaseAlgorithm::ClusterListMap &, const VertexSelectionBaseAlgorithm::KDTreeMap &,
+    const VertexSelectionBaseAlgorithm::ShowerClusterListMap &showerClusterListMap, const float, float &)
+{
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+        std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    float asymmetry(0.f);
+
+    asymmetry += this->GetAsymmetryForView(
+        LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex->GetPosition(), TPC_VIEW_U), slidingFitDataListMap.at(TPC_VIEW_U), showerClusterListMap.at(TPC_VIEW_U));
+
+    asymmetry += this->GetAsymmetryForView(
+        LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex->GetPosition(), TPC_VIEW_V), slidingFitDataListMap.at(TPC_VIEW_V), showerClusterListMap.at(TPC_VIEW_V));
+
+    asymmetry += this->GetAsymmetryForView(
+        LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex->GetPosition(), TPC_VIEW_W), slidingFitDataListMap.at(TPC_VIEW_W), showerClusterListMap.at(TPC_VIEW_W));
+
+    featureVector.push_back(asymmetry);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void AsymmetryFeatureBaseTool::IncrementAsymmetryParameters(
+    const float weight, const CartesianVector &clusterDirection, CartesianVector &localWeightedDirectionSum) const
+{
+    // If the new axis direction is at an angle of greater than 90 deg to the current axis direction, flip it 180 degs.
+    CartesianVector newDirection(clusterDirection);
+
+    if (localWeightedDirectionSum.GetMagnitudeSquared() > std::numeric_limits<float>::epsilon())
+    {
+        if (localWeightedDirectionSum.GetCosOpeningAngle(clusterDirection) < 0.f)
+            newDirection *= -1.f;
+    }
+
+    localWeightedDirectionSum += newDirection * weight;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template <typename C, typename T = typename C::value_type>
+float AsymmetryFeatureBaseTool::CalculateAsymmetry(const bool useEnergyMetrics, const CartesianVector &vertexPosition2D,
+    const C &asymmetryClusters, const CartesianVector &localWeightedDirectionSum) const
+{
+    // Project every hit onto local event axis direction and record side of the projected vtx position on which it falls
+    float beforeVtxHitEnergy(0.f), afterVtxHitEnergy(0.f);
+    unsigned int beforeVtxHitCount(0), afterVtxHitCount(0);
+
+    const CartesianVector localWeightedDirection(localWeightedDirectionSum.GetUnitVector());
+    const float evtProjectedVtxPos(vertexPosition2D.GetDotProduct(localWeightedDirection));
+
+    for (const Cluster *const pCluster : asymmetryClusters)
+    {
+        CaloHitList caloHitList;
+        pCluster->GetOrderedCaloHitList().FillCaloHitList(caloHitList);
+
+        CaloHitVector caloHitVector(caloHitList.begin(), caloHitList.end());
+        std::sort(caloHitVector.begin(), caloHitVector.end(), LArClusterHelper::SortHitsByPosition);
+
+        for (const CaloHit *const pCaloHit : caloHitVector)
+        {
+            if (pCaloHit->GetPositionVector().GetDotProduct(localWeightedDirection) < evtProjectedVtxPos)
+            {
+                beforeVtxHitEnergy += pCaloHit->GetElectromagneticEnergy();
+                ++beforeVtxHitCount;
+            }
+
+            else
+            {
+                afterVtxHitEnergy += pCaloHit->GetElectromagneticEnergy();
+                ++afterVtxHitCount;
+            }
+        }
+    }
+
+    // Use energy metrics if possible, otherwise fall back on hit counting.
+    const float totHitEnergy(beforeVtxHitEnergy + afterVtxHitEnergy);
+    const unsigned int totHitCount(beforeVtxHitCount + afterVtxHitCount);
+
+    if (useEnergyMetrics && (totHitEnergy > std::numeric_limits<float>::epsilon()))
+        return std::fabs((afterVtxHitEnergy - beforeVtxHitEnergy)) / totHitEnergy;
+
+    if (0 == totHitCount)
+        throw StatusCodeException(STATUS_CODE_FAILURE);
+
+    return std::fabs((static_cast<float>(afterVtxHitCount) - static_cast<float>(beforeVtxHitCount))) / static_cast<float>(totHitCount);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode AsymmetryFeatureBaseTool::ReadSettings(const TiXmlHandle xmlHandle)
+{
+    PANDORA_RETURN_RESULT_IF_AND_IF(
+        STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "MaxAsymmetryDistance", m_maxAsymmetryDistance));
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+template float AsymmetryFeatureBaseTool::CalculateAsymmetry(const bool useEnergyMetrics, const CartesianVector &vertexPosition2D,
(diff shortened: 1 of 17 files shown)


larpandoracontent/LArVertex/EnergyDepositionAsymmetryFeatureTool.cc
```diff
@@ -0,0 +1,193 @@
...
+
+    const float totalEnergyDeposition(beforeVtxEnergyDeposition + afterVtxEnergyDeposition);
+
+    if (useEnergyMetrics && totalEnergyDeposition)
+        return std::fabs((afterVtxEnergyDeposition - beforeVtxEnergyDeposition)) / totalEnergyDeposition;
+
+    if (0 == totHitCount)
+        throw StatusCodeException(STATUS_CODE_FAILURE);
+
+    const float beforeVtxHitDeposition(
```
MariaBrigida, line 169: Hey Henry, this is certainly just me being extra picky but I see a minor inconsistency in the naming convention here... "After vertex hit energy" becomes "After vertex energy deposition" when dividing by the axis length, but "After vertex hit count" becomes "after vertex hit deposition". I also wonder if "deposition" is the most descriptive name that could be used here?

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -561,6 +640,30 @@ const pandora::Vertex *TrainedVertexSelectionAlgorithm::ProduceTrainingExamples(
...
+
+                else
+                {
+                    LArMvaHelper::ProduceTrainingExample(trainingOutputFile + "_" + interactionType + ".txt", false, eventFeatureList,
+                        featureList, bestVertexFeatureList, sharedFeatureList);
+                }
+            }
+        }
+
         if (pBestVertex && (bestVertexDr < maxRadius))
```
AndyChappell, line 667: I'm a little confused by the second block of training example production. This seems to imply that two lots of training examples will be produced for the legacy code. Is this intended?

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -582,6 +685,96 @@ const pandora::Vertex *TrainedVertexSelectionAlgorithm::ProduceTrainingExamples(
...
+    this->IncrementSharedAxisValues(LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex1->GetPosition(), TPC_VIEW_V),
+        LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex2->GetPosition(), TPC_VIEW_V), kdTreeMap.at(TPC_VIEW_V), axisHits);
+
+    this->IncrementSharedAxisValues(LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex1->GetPosition(), TPC_VIEW_W),
+        LArGeometryHelper::ProjectPosition(this->GetPandora(), pVertex2->GetPosition(), TPC_VIEW_W), kdTreeMap.at(TPC_VIEW_W), axisHits);
+
+    if (!separation)
+        axisHits = 0;
+
+    axisHits /= separation;
```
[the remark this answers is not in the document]
AndyChappell: As Maria notes, I'd be inclined to replace this conditional and division with

```
axisHits = separation > std::numeric_limits<float>::epsilon() ? axisHits / separation : 0.f;
```

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -582,6 +685,96 @@ const pandora::Vertex *TrainedVertexSelectionAlgorithm::ProduceTrainingExamples(
...
+
+    // Use a kd search to find the hits in the 'wide' area
+    KDTreeBox searchBox(xMin, zMin, xMax, zMax);
+    HitKDNode2DList found;
+    kdTree.search(searchBox, found);
+
+    // Use IsHitInBox method to check the 'wide' area hits for those in the search box
+    for (auto f : found)
+    {
+        const CartesianVector hitPos = f.data->GetPositionVector();
```
AndyChappell, line 745: Make this a reference variable to avoid a copy

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -582,6 +685,96 @@ const pandora::Vertex *TrainedVertexSelectionAlgorithm::ProduceTrainingExamples(
...
+    // Find the total coordinate span these points cover
+    float xMin(std::numeric_limits<float>::max());
+    float xMax(-std::numeric_limits<float>::max());
+    float zMin(std::numeric_limits<float>::max());
+    float zMax(-std::numeric_limits<float>::max());
+
+    this->UpdateBoxEdges(point1, xMin, xMax, zMin, zMax);
+    this->UpdateBoxEdges(point2, xMin, xMax, zMin, zMax);
+    this->UpdateBoxEdges(point3, xMin, xMax, zMin, zMax);
+    this->UpdateBoxEdges(point4, xMin, xMax, zMin, zMax);
```
AndyChappell, line 735: You'd want to double-check this, but I think you can get the same result by simply defining

```
const float xMin{std::min({point1.GetX(), point2.GetX(), point3.GetX(), point4.GetX()})}
```
And equivalently for the other minima and maxima
(note the curly braces that allow std::min to compare more than two values).

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -582,6 +685,96 @@ const pandora::Vertex *TrainedVertexSelectionAlgorithm::ProduceTrainingExamples(
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool TrainedVertexSelectionAlgorithm::IsHitInBox(const CartesianVector &hitPos, const CartesianVector &point1,
+    const CartesianVector &point2, const CartesianVector &point3, const CartesianVector &point4) const
+{
+    bool b1 = std::signbit(((point2 - point1).GetCrossProduct(point2 - hitPos)).GetY());
+    bool b2 = std::signbit(((point4 - point3).GetCrossProduct(point4 - hitPos)).GetY());
+    bool b3 = std::signbit(((point3 - point1).GetCrossProduct(point3 - hitPos)).GetY());
```
AndyChappell, line 770: You could check the b1 xor b2 case here to allow for an early return false if it's zero

larpandoracontent/LArVertex/EnergyDepositionAsymmetryFeatureTool.cc
```diff
@@ -0,0 +1,193 @@
+/**
```
AndyChappell, line 1: A general comment about this class: As @MariaBrigida  pointed out to me, this class is extremely similar to the existing GlobalAsymmetryFeatureTool; the only difference appears to be in the CalculateEnergyDepositionAsymmetry function, corresponding to the CalculateGlobalAsymmetry function. I wonder if this might simply inherit from that class and simply override the CalculateGlobalAsymmetry function, or perhaps even modify the existing class to optionally compute the energy asymmetry instead?

AndyChappell (commented):
Thanks for the updates, Henry. Most of my comments are pretty minor, just a couple of more notable queries regarding production of the training examples and the similarity between the global and energy asymmetry tools. Otherwise this looks good to me.

larpandoracontent/LArVertex/EnergyDepositionAsymmetryFeatureTool.cc
henrylay97: Thanks for taking a look Maria! Yeah you're right it probably could be a little more consistent here. I'll try and wrap this up with the comments on linking the tool to the `GlobalAsymmetryFeatureTool`.

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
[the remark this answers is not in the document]
henrylay97: Yep, amazingly I ran full training samples without this screaming at me . I'm either very lucky or there is a check built into the candidate creation that avoids `separation==0`. I'll definitely add the check though.

larpandoracontent/LArVertex/VertexSelectionBaseAlgorithm.h
```diff
@@ -306,6 +306,39 @@ class VertexSelectionBaseAlgorithm : public pandora::Algorithm
...
 
+    /**
+     *  @brief  Whether the vertex lies in a registered gap
+     *
+     *  @param  pVertex the address of the vertex
+     *  @param  hitType the relevant hit type
+     *
+     *  @return boolean
+     */
+    bool IsVertexInGap(const pandora::Vertex *const pVertex, const pandora::HitType hitType) const;
```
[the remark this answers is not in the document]
henrylay97: No there isn't sorry. I had originally put my new methods in private and must have been over-zealous when moving them to protected!

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
```diff
@@ -387,7 +453,19 @@ void TrainedVertexSelectionAlgorithm::PopulateVertexFeatureInfoMap(const BeamCon
...
 
-    VertexFeatureInfo vertexFeatureInfo(beamDeweighting, 0.f, energyKick, localAsymmetry, globalAsymmetry, showerAsymmetry);
+    double dEdxAsymmetry(0.f), vertexEnergy(0.f);
+
+    if (!m_legacyVariables)
+    {
+        dEdxAsymmetry = LArMvaHelper::CalculateFeaturesOfType<EnergyDepositionAsymmetryFeatureTool>(m_featureToolVector, this, pVertex,
+            slidingFitDataListMap, clusterListMap, kdTreeMap, showerClusterListMap, beamDeweighting, bestFastScore)
+                            .at(0)
+                            .Get();
```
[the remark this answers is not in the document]
henrylay97: Yep it did, likewise to the other tools above. I hate it too ;)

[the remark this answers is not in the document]
AndyChappell: Hopefully, we'll get a newer release of clang available via CVMFS and we can update the format to take advantage of some of the newer features!

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
henrylay97: Yep this is a mistake. The second set should be in an else block. It should also be `!m_legacyVariables`. This didn't scream at me locally because unlike with the operational legacy mode it will still run fine. Thanks so much for picking this one up!

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
henrylay97: I'll take a look, if so this is much neater thank you!

larpandoracontent/LArVertex/EnergyDepositionAsymmetryFeatureTool.cc
henrylay97: This is very true, there are quite strong overlaps with the other two asymmetry variables as well (shower and local). I could take a look at whether a base version of the tool with 4 inheriting versions might be neater?

AndyChappell: Sounds good thanks, Henry. Anything that can reasonably reduce code duplication without over-burdening a base class is useful for future maintenance.

larpandoracontent/LArVertex/TrainedVertexSelectionAlgorithm.cc
henrylay97: I have done some checks locally to ensure the files are producing the number of features I'm expecting. Everything looks okay now :)

henrylay97:
Thanks for all the useful comments. These recent commits should respond to them all. They also create the `AsymmetryFeatureBaseTool` from which the `Global`, `Local` and `Shower` asymmetry tools all inherit. The `EnergyDeposition` tool then inherits further from the `Global` tool. Apologies for the more messy commit history, I've tested the functionality of this new setup but I'm sure there are improvements to the way I've implemented the inheritance.

AndyChappell:
Thanks, Henry. Will take a look at the updates, and no problem on the commit history, we can tidy that up afterwards if necessary.

larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc
```diff
@@ -0,0 +1,123 @@
...
+        std::sort(caloHitVector.begin(), caloHitVector.end(), LArClusterHelper::SortHitsByPosition);
+
+        for (const CaloHit *const pCaloHit : caloHitVector)
+        {
+            if (pCaloHit->GetPositionVector().GetDotProduct(localWeightedDirection) < evtProjectedVtxPos)
+            {
+                beforeVtxHitEnergy += pCaloHit->GetElectromagneticEnergy();
+                ++beforeVtxHitCount;
+            }
+
```
AndyChappell, line 91: Remove excess newline

larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.h
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     *  @brief  Run the tool
+     *
+     *  @param  pAlgorithm address of the calling algorithm
+     *  @param  pVertex address of the vertex
+     *  @param  slidingFitDataListMap map of the sliding fit data lists
+     *
+     *  @return the asymmetry feature
+     */
```
AndyChappell, line 35: Parameter documentation needs updating

AndyChappell (approved):
Hi Henry, so subject to a few pretty minor comments, I think the update is looking good, so once those are addressed and pending anything from @johnmarshall80 and @MariaBrigida, I think this one is ready to merge into a future release.

larpandoracontent/LArVertex/LocalAsymmetryFeatureTool.h
```diff
@@ -8,37 +8,24 @@
...
-    void Run(LArMvaHelper::MvaFeatureVector &featureVector, const VertexSelectionBaseAlgorithm *const pAlgorithm, const pandora::Vertex *const pVertex,
-        const VertexSelectionBaseAlgorithm::SlidingFitDataListMap &slidingFitDataListMap, const VertexSelectionBaseAlgorithm::ClusterListMap &,
-        const VertexSelectionBaseAlgorithm::KDTreeMap &, const VertexSelectionBaseAlgorithm::ShowerClusterListMap &, const float, float &);
-
 private:
-    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle) override;
 
     /**
      *  @brief  Get the local asymmetry feature in a given view
```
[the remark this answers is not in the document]
henrylay97: Does it? I'm probably being thick but I can't see which bit needs updating. The method is called `GetAsymmetryForView` because its overriding from the base tool but in this scenario the calculation is specifically for the local asymmetry. The shower cluster list is not used and so is not passed in this overriden version.

larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.h
```diff
@@ -0,0 +1,84 @@
...
+    void IncrementAsymmetryParameters(
+        const float weight, const pandora::CartesianVector &clusterDirection, pandora::CartesianVector &localWeightedDirectionSum) const;
+
+    /**
+     *  @brief  Calculate the asymmetry feature
+     *
+     *  @param  useEnergyMetrics whether to use energy-based metrics instead of hit-counting-based metrics
+     *  @param  vertexPosition2D the vertex position in this view
+     *  @param  slidingFitDataList the list of sliding fit data objects
```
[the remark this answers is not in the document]
henrylay97: Sorry for all of these, clearly an oversight on my part in not updating the documentation for the base tool after all the updates!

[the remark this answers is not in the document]
henrylay97: Sorry for all of these, clearly an oversight on my part not to update the documentation on the base tool after all the changes were done!

larpandoracontent/LArVertex/LocalAsymmetryFeatureTool.h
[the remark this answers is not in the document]
AndyChappell: No, you're right, the third argument is a placeholder for this override use case, I was reading ShowerClusterList as the parameter name, not part of the type. Sorry.

larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.h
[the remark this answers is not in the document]
AndyChappell: No problem, easily done.

larpandoracontent/LArVertex/LocalAsymmetryFeatureTool.h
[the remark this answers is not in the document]
henrylay97: Fab! No worries :D

henrylay97:
Thanks again for all the useful comments! Have responded in the new commit.

PR #170, opened 2021-05-14
feature/hierarchy_tools

AndyChappell:
This PR adds new reco/truth matching tools. These tools are designed to make it easier to determine the correspondence between reconstructed particles and MC particles, and assess the quality of the matches, with a view to increasing the sophistication of the analysis of reconstructed interaction hierarchies moving forward. Given lists of MC particles, CalHits and PFOs, matching can now be performed with the following short segment of code:

```
LArHierarchyHelper::MCHierarchy mcHierarchy;
LArHierarchyHelper::FillMCHierarchy(*pMCParticleList, *pCaloHitList, true, false, mcHierarchy);
LArHierarchyHelper::RecoHierarchy recoHierarchy;
LArHierarchyHelper::FillRecoHierarchy(*pPfoList, true, false, recoHierarchy);
LArHierarchyHelper::MatchInfo matchInfo;
LArHierarchyHelper::MatchHierarchies(mcHierarchy, recoHierarchy, matchInfo);
```

Different hierarchy folding options are available according to the booleans passed to the hierarchy filling functions. Matching information is made available to quickly establish links between reco and truth and the summary picture of the reconstruction. The MC and Reco hierarchies can also be queried in greater detail as needed.

### DIFF
--- a/larpandoracontent/LArHelpers/LArHierarchyHelper.cc
+++ b/larpandoracontent/LArHelpers/LArHierarchyHelper.cc
@@ -47,7 +47,6 @@ void LArHierarchyHelper::MCHierarchy::FillHierarchy(const MCParticleList &mcPart
         }
         catch(const StatusCodeException&)
         {
-            std::cout << "Found calo hit with no MC" << std::endl;
         }
     }
 
@@ -241,6 +240,7 @@ const std::string LArHierarchyHelper::MCHierarchy::ToString() const
 
 LArHierarchyHelper::MCHierarchy::Node::Node(const MCHierarchy &hierarchy, const MCParticle *pMC) :
     m_hierarchy(hierarchy),
+    m_mainParticle(pMC),
     m_pdg{0}
 {
     if (pMC)
@@ -254,10 +254,14 @@ LArHierarchyHelper::MCHierarchy::Node::Node(const MCHierarchy &hierarchy, const 
 
 LArHierarchyHelper::MCHierarchy::Node::Node(const MCHierarchy &hierarchy, const MCParticleList &mcParticleList, const CaloHitList &caloHitList) :
     m_hierarchy(hierarchy),
+    m_mainParticle(nullptr),
     m_pdg{0}
 {
     if (!mcParticleList.empty())
-        m_pdg = mcParticleList.front()->GetParticleId();
+    {
+        m_mainParticle = mcParticleList.front();
+        m_pdg = m_mainParticle->GetParticleId();
+    }
     m_mcParticles = mcParticleList; m_mcParticles.sort();
     m_caloHits = caloHitList; m_caloHits.sort();
 }
@@ -402,9 +406,29 @@ int LArHierarchyHelper::MCHierarchy::Node::GetParticleId() const
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+bool LArHierarchyHelper::MCHierarchy::Node::IsTestBeamParticle() const
+{
+    if (m_mainParticle)
+        return LArMCParticleHelper::IsBeamParticle(m_mainParticle);
+    else
+        return false;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool LArHierarchyHelper::MCHierarchy::Node::IsCosmicRay() const
+{
+    if (m_mainParticle)
+        return LArMCParticleHelper::IsCosmicRay(m_mainParticle);
+    else
+        return false;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
 const std::string LArHierarchyHelper::MCHierarchy::Node::ToString(const std::string &prefix) const
 {
-    std::string str(prefix + "PDG: " + std::to_string(m_pdg) + " Energy: " + std::to_string(m_mcParticles.front()->GetEnergy()) +
+    std::string str(prefix + "PDG: " + std::to_string(m_pdg) + " Energy: " + std::to_string(m_mainParticle ? m_mainParticle->GetEnergy() : 0) +
         " Hits: " + std::to_string(m_caloHits.size()) + "\n");
     for (const Node *pChild : m_children)
         str += pChild->ToString(prefix + "   ");
@@ -873,10 +897,13 @@ void LArHierarchyHelper::MatchHierarchies(const MCHierarchy &mcHierarchy, const 
 
     for (const MCHierarchy::Node *pMCNode : mcNodes)
     {
-        if (mcToMatchMap.find(pMCNode) == mcToMatchMap.end())
-        {   // Unmatched MC
-            MCMatches match(pMCNode);
-            matchVector.emplace_back(match);
+        if (pMCNode->IsReconstructable())
+        {
+            if (mcToMatchMap.find(pMCNode) == mcToMatchMap.end())
+            {   // Unmatched MC
+                MCMatches match(pMCNode);
+                matchVector.emplace_back(match);
+            }
         }
     }
 
@@ -886,13 +913,25 @@ void LArHierarchyHelper::MatchHierarchies(const MCHierarchy &mcHierarchy, const 
         };
     std::sort(matchVector.begin(), matchVector.end(), predicate);
 
+    int nNeutrinoMCParticles{0}, nNeutrinoRecoParticles{0}, nNeutrinoRecoBTParticles{0};
+    int nCosmicMCParticles{0}, nCosmicRecoParticles{0}, nCosmicRecoBTParticles{0};
+    int nTestBeamMCParticles{0}, nTestBeamRecoParticles{0}, nTestBeamRecoBTParticles{0};
+    //int nTriggeredTestBeamMCParticles{0}, nTriggeredTestBeamRecoParticles{0};
     for (const MCMatches &match : matchVector)
     {
         const MCHierarchy::Node *pMCNode{match.GetMC()};
+        if (pMCNode->IsTestBeamParticle())
+            ++nTestBeamMCParticles;
+        else if (pMCNode->IsCosmicRay())
+            ++nCosmicMCParticles;
+        else
+            ++nNeutrinoMCParticles;
         const int pdg{pMCNode->GetParticleId()};
         const size_t mcHits{pMCNode->GetCaloHits().size()};
-        std::cout << "MC " << pdg << " hits " << mcHits << std::endl;
+        const std::string tag{pMCNode->IsTestBeamParticle() ? "(Beam) " : pMCNode->IsCosmicRay() ? "(Cosmic) " : ""};
+        std::cout << "MC " << tag << pdg << " hits " << mcHits << std::endl;
         const RecoHierarchy::NodeVector &nodeVector{match.GetRecoMatches()};
+        int nGoodReco{0};
         for (const RecoHierarchy::Node *pRecoNode : nodeVector)
         {
             const int recoHits{static_cast<int>(pRecoNode->GetCaloHits().size())};
@@ -901,9 +940,66 @@ void LArHierarchyHelper::MatchHierarchies(const MCHierarchy &mcHierarchy, const 
             const float completeness{match.GetCompleteness(pRecoNode)};
             std::cout << "   Matched " << sharedHits << " out of " << recoHits << " with purity " << purity << " and completeness " <<
                 completeness << std::endl;
+            if (purity > 0.5f && completeness > 0.1f)
+                ++nGoodReco;
         }
-        if (nodeVector.empty())
+        if (nodeVector.size() == 1 && nGoodReco == 1)
+        {   // We have a good, single match
+            if (pMCNode->IsTestBeamParticle())
+                ++nTestBeamRecoParticles;
+            else if (pMCNode->IsCosmicRay())
+                ++nCosmicRecoParticles;
+            else
+                ++nNeutrinoRecoParticles;
+        }
+        else if (!nodeVector.empty())
+        {   // We have either below threshold or multiple matches
+            if (pMCNode->IsTestBeamParticle())
+                ++nTestBeamRecoBTParticles;
+            else if (pMCNode->IsCosmicRay())
+                ++nCosmicRecoBTParticles;
+            else
+                ++nNeutrinoRecoBTParticles;
+        }
+        else if (nodeVector.empty())
             std::cout << "   Unmatched" << std::endl;
+    }
+
+    if (mcHierarchy.IsNeutrinoHierarchy())
+    {
+        std::cout << "Neutrino Interaction Summary:" << std::endl;
+        std::cout << std::fixed << std::setprecision(1);
+        if (nNeutrinoMCParticles)
+        {
+            std::cout << "Matched final state particles: " << nNeutrinoRecoParticles << " of " << nNeutrinoMCParticles << " : " <<
+                (100 * nNeutrinoRecoParticles / static_cast<float>(nNeutrinoMCParticles)) << "%" << std::endl;
+        }
+        if (nCosmicMCParticles)
+        {
+            std::cout << "Matched cosmics: " << nCosmicRecoParticles << " of " << nCosmicMCParticles << " : " <<
+                (100 * nCosmicRecoParticles / static_cast<float>(nCosmicMCParticles)) << "%" << std::endl;
+        }
+    }
+    else if (mcHierarchy.IsTestBeamHierarchy())
+    {
+        std::cout << "Test Beam Interaction Summary:" << std::endl;
+        std::cout << std::fixed << std::setprecision(1);
+        if (nTestBeamMCParticles)
+        {
+            std::cout << "Matched test beam particles: " << nTestBeamRecoParticles << " of " << nTestBeamMCParticles << " : " <<
+                (100 * nTestBeamRecoParticles / static_cast<float>(nTestBeamMCParticles)) << "%" << std::endl;
+            std::cout << "Loosely matched test beam particles: " << (nTestBeamRecoParticles + nTestBeamRecoBTParticles) << " of " <<
+                nTestBeamMCParticles << " : " << (100 * (nTestBeamRecoParticles + nTestBeamRecoBTParticles) /
+                    static_cast<float>(nTestBeamMCParticles)) << "%" << std::endl;
+        }
+        if (nCosmicMCParticles)
+        {
+            std::cout << "Matched cosmics: " << nCosmicRecoParticles << " of " << nCosmicMCParticles << " : " <<
+                (100 * nCosmicRecoParticles / static_cast<float>(nCosmicMCParticles)) << "%" << std::endl;
+            std::cout << "Loosely matched cosmics: " << (nCosmicRecoParticles + nCosmicRecoBTParticles) << " of " <<
+                nCosmicMCParticles << " : " << (100 * (nCosmicRecoParticles + nCosmicRecoBTParticles) /
+                    static_cast<float>(nCosmicMCParticles)) << "%" << std::endl;
+        }
     }
 }
 
@@ -918,13 +1014,16 @@ const MCParticle *LArHierarchyHelper::GetMCPrimaries(const MCParticleList &mcPar
         try
         {
             const MCParticle *const pPrimary{LArMCParticleHelper::GetPrimaryMCParticle(pMC)};
-            primaries.insert(pPrimary);
+            if (!LArMCParticleHelper::IsTriggeredBeamParticle(pPrimary))
+                primaries.insert(pPrimary);
+            else
+                primaries.insert(pPrimary);
         }
         catch (const StatusCodeException&)
         {
             if (LArMCParticleHelper::IsNeutrino(pMC))
                 pRoot = pMC;
-            else if (pMC->GetParticleId() != 111)
+            else if (pMC->GetParticleId() != 111 && pMC->GetParticleId() < 1e9)
                 std::cout << "LArHierarchyHelper::MCHierarchy::FillHierarchy: MC particle with PDG code " << pMC->GetParticleId() <<
                     " at address " << pMC << " has no associated primary particle" << std::endl;
         }
@@ -936,6 +1035,7 @@ const MCParticle *LArHierarchyHelper::GetMCPrimaries(const MCParticleList &mcPar
 const ParticleFlowObject *LArHierarchyHelper::GetRecoPrimaries(const PfoList &pfoList, PfoSet &primaries)
 {
     const ParticleFlowObject *pRoot{nullptr};
+    PfoSet cosmicPfos;
     for (const ParticleFlowObject *pPfo : pfoList)
     {
         if (LArPfoHelper::IsNeutrino(pPfo))
@@ -946,20 +1046,39 @@ const ParticleFlowObject *LArHierarchyHelper::GetRecoPrimaries(const PfoList &pf
         else
         {
             const ParticleFlowObject *const pParent{LArPfoHelper::GetParentPfo(pPfo)};
-            if (LArPfoHelper::IsNeutrino(pParent))
+            if (pParent && LArPfoHelper::IsNeutrino(pParent))
             {
                 pRoot = pParent;
                 break;
             }
             else
-            {
-                std::cout << "Still need to handle test beam, cosmics, ..." << std::endl;
+            {   // Should be in a test beam scenario
+                const int tier{LArPfoHelper::GetHierarchyTier(pPfo)};
+                if (tier == 0 && LArPfoHelper::IsTestBeam(pPfo))
+                {   // Triggered beam particle
+                    primaries.insert(pPfo);
+                    continue;
+                }
+                if (tier > 1)
+                    continue;
+                if (!LArPfoHelper::IsTestBeam(pPfo))
+                {   // Cosmic induced
+                    cosmicPfos.insert(pPfo);
+                }
             }
         }
     }
-    const PfoList &children{pRoot->GetDaughterPfoList()};
-    for (const ParticleFlowObject *pPrimary : children)
-        primaries.insert(pPrimary); 
+    if (pRoot && LArPfoHelper::IsNeutrino(pRoot))
+    {
+        const PfoList &children{pRoot->GetDaughterPfoList()};
+        for (const ParticleFlowObject *pPrimary : children)
+            primaries.insert(pPrimary);
+    }
+    else
+    {
+        for(const ParticleFlowObject *pPfo : cosmicPfos)
+            primaries.insert(pPfo);
+    }
 
     return pRoot;
 }
--- a/larpandoracontent/LArHelpers/LArHierarchyHelper.cc
+++ b/larpandoracontent/LArHelpers/LArHierarchyHelper.cc
@@ -11,6 +11,8 @@
 
 #include "larpandoracontent/LArHelpers/LArHierarchyHelper.h"
 
+#include <numeric>
+
 namespace lar_content
 {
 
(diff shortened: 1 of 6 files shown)


larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+            const bool isNeutron{pdg == NEUTRON};
+            if (isShower || isNeutron)
+            {
+                if (!m_recoCriteria.m_removeNeutrons)
+                {
+                    LArMCParticleHelper::GetAllDescendentMCParticles(pPrimary, allParticles);
+                }
+                else
+                {   // Throw away neutrons
+                    MCParticleList dummy;
```
lhwhitehead, line 106: Not really an issue, but you used "neutrons" as the name of the dummy list on line 69?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+        Node *pNode{new Node(m_hierarchy, allParticles, allHits)};
+        m_children.emplace_back(pNode);
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool LArHierarchyHelper::MCHierarchy::Node::IsReconstructable() const
+{
+    const bool enoughHits{m_caloHits.size() >= m_hierarchy.m_recoCriteria.m_minHits};
```
lhwhitehead, line 350: Couldn't you return false here if enoughHits is false and save doing the rest of the function?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+            }
+        }
+        else
+        {
+            m_unmatchedReco.emplace_back(pRecoNode);
+        }
+    }
+
+    for (auto [ pMCNode, matches ] : mcToMatchMap)
+    {   (void)pMCNode;  // gcc 7 requirement
```
lhwhitehead, line 911: Is there a reason why this is on the same line as the {?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+        else
+        {
+            m_subThresholdMatches.emplace_back(matches);
+        }
+    }
+
+    auto predicate = [](const MCMatches &lhs, const MCMatches &rhs)
+        {
+            return lhs.GetMC()->GetCaloHits().size() > rhs.GetMC()->GetCaloHits().size();
+        };
```
lhwhitehead, line 932: Over-indented?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
AndyChappell: Just a personal preference for these sort of technical requirements. It has no relevance to the logic of the code and it stands out here as something that can be removed as and when we drop support for legacy versions of gcc.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
AndyChappell: We've started using a .clang-format file to standardise formatting for all future PRs. So this is a clang choice. Hopefully once a newer version of clang is available on CVMFS we'll have more flexibility to fix some of the things we don't like.

larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -173,6 +173,38 @@ void LArPfoHelper::GetAllDownstreamPfos(const ParticleFlowObject *const pPfo, Pf
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArPfoHelper::GetAllDownstreamPfos(
+    const pandora::ParticleFlowObject *const pPfo, pandora::PfoList &outputTrackPfoList, pandora::PfoList &outputLeadingShowerPfoList)
+{
+    if (LArPfoHelper::IsTrack(pPfo))
+    {
+        outputTrackPfoList.emplace_back(pPfo);
+        const PfoList &childPfoList(pPfo->GetDaughterPfoList());
```
lhwhitehead, line 182: Do we need this definition as opposed to using `pPfo->GetDaughterPfoList()` in the for loop?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
lhwhitehead: Bit of a clanger? 
Sorry.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+        }
+    }
+
+    if (foldToPrimaries && !foldToLeadingShowers)
+    {
+        MCParticleSet primarySet;
+        m_pNeutrino = LArHierarchyHelper::GetMCPrimaries(mcParticleList, primarySet);
+        MCParticleList primaries(primarySet.begin(), primarySet.end());
+        if (m_recoCriteria.m_removeNeutrons)
+            primaries.erase(std::remove_if(primaries.begin(), primaries.end(), predicate), primaries.end());
```
lhwhitehead, line 59: Are these first five lines the same in all of the four cases in the if / else statements? Could move them outside if that's the case.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void LArHierarchyHelper::RecoHierarchy::FillHierarchy(const PfoList &pfoList, const bool foldToPrimaries, const bool foldToLeadingShowers)
+{
+    if (foldToPrimaries && !foldToLeadingShowers)
+    {
+        PfoSet primaries;
+        m_pNeutrino = LArHierarchyHelper::GetRecoPrimaries(pfoList, primaries);
```
lhwhitehead, line 495: Can move these two lines out of the if statement as they are common to all cases?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
AndyChappell: Yes, turns out they are - I think there's also one other set of code blocks relating to populating the joint hit lists that probably also warrants some refactoring.

larpandoracontent/LArHelpers/LArPfoHelper.cc
```diff
@@ -173,6 +173,38 @@ void LArPfoHelper::GetAllDownstreamPfos(const ParticleFlowObject *const pPfo, Pf
...
+        for (const ParticleFlowObject *pChild : childPfoList)
+        {
+            if (std::find(outputTrackPfoList.begin(), outputTrackPfoList.end(), pChild) == outputTrackPfoList.end())
+            {
+                const int pdg{std::abs(pChild->GetParticleId())};
+                if (pdg == E_MINUS)
```
MariaBrigida, line 188: Probably missing something, but why is this is only checking for e-?

larpandoracontent/LArHelpers/LArMCParticleHelper.cc
```diff
@@ -248,19 +248,47 @@ const MCParticle *LArMCParticleHelper::GetParentMCParticle(const MCParticle *con
...
         }
     }
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+void LArMCParticleHelper::GetAllDescendentMCParticles(const MCParticle *const pMCParticle, MCParticleList &descendentTrackParticles,
+    MCParticleList &leadingShowerParticles, MCParticleList &leadingNeutrons)
+{
+    const MCParticleList &daughterMCParticleList(pMCParticle->GetDaughterList());
```
MariaBrigida, line 267: If going by Leigh's suggestion in LArPfoHelper.cc, can perhaps do the same here and remove this definition too...

MariaBrigida: However I can see this appears in the function above as well...

larpandoracontent/LArHelpers/LArPfoHelper.cc
AndyChappell: E_MINUS happens to be the slightly awkwardly named PDG code corresponding to the absolute value for electrons

MariaBrigida: I see - thank you! How about photons?

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+
+            /**
+             *  @brief  Copy constructor
+             */
+            ReconstructabilityCriteria(const ReconstructabilityCriteria &obj);
+
+            /**
+             *  @brief  Constructor
+             *
+             *  @param  minHits The total minimumm number of hits for a particle to be considered reconstructable
```
MariaBrigida, line 52: Typo: "minimum"

larpandoracontent/LArHelpers/LArPfoHelper.cc
AndyChappell: PFO convention is muon = track, electron = shower, so it's track/shower distinction rather than actual PID here.

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+
+        virtual ~MCMatches() = default;
+
+        // Might be better to pass in the shared hits themselves
+        /**
+         *  @brief  Add a reconstructed node as a match for this MC node
+         *
+         *  @param  pReco The reconstructed node that matches this MC node
+         */
+        void AddRecoMatch(const RecoHierarchy::Node *pReco, const int nSharedHits);
```
MariaBrigida, line 441: nSharedHits missing description... (unless you plan to change the argument as per your comment above)

AndyChappell: Ah yes, I did omit for that reason, but actually, I'm currently planning to stick with the count, so will need to document this.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+            if (isShower || isNeutron)
+            {
+                if (!m_recoCriteria.m_removeNeutrons)
+                {
+                    LArMCParticleHelper::GetAllDescendentMCParticles(pPrimary, allParticles);
+                }
+                else
+                {   // Throw away neutrons
+                    MCParticleList dummy;
+                    LArMCParticleHelper::GetAllDescendentMCParticles(pPrimary, allParticles, allParticles, dummy);
```
MariaBrigida, line 107: How come the neutron list is sometimes called "dummy", sometimes "neutron", if the contents are discarded anyway? Could the same list be reused?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+        {
+            MCParticleList allParticles{pPrimary};
+            if (!m_recoCriteria.m_removeNeutrons)
+            {
+                LArMCParticleHelper::GetAllDescendentMCParticles(pPrimary, allParticles);
+            }
+            else
+            {   // Collect track-like and shower-like particles together, but throw out neutrons and descendents
+                MCParticleList neutrons;
+                LArMCParticleHelper::GetAllDescendentMCParticles(pPrimary, allParticles, allParticles, neutrons);
```
MariaBrigida, line 70: The neutrons here are just thrown away - later on, an analogous list is labelled as "dummy"; could perhaps make a consistent choice here... (I'd previously put in a similar comment, then cancelled it as the logic wasn't fully clear to me yet. sorry)

AndyChappell: Yes, originally I was always keeping neutrons and just tagging them separately, so this is a legacy of providing an option to throw them away altogether. I'll alter the cases where there is no folding that might still keep them as leading particles to use 'dummy' everywhere. I think Leigh made a similar comment as well, but I don't see it anywhere.

lhwhitehead: Ah, I removed it as it was in the same comment as something that ended up untrue.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+float LArHierarchyHelper::MCMatches::GetCompleteness(const RecoHierarchy::Node *pReco) const
+{
+    auto iter{std::find(m_recoNodes.begin(), m_recoNodes.end(), pReco)};
+    if (iter == m_recoNodes.end())
+        throw StatusCodeException(STATUS_CODE_NOT_FOUND);
+    int index = iter - m_recoNodes.begin();
```
johnmarshall80, line 812: Is there some casting between types going on here? Could make explicit. Const too? Looks OK, though.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+
+    return m_sharedHits[index] / static_cast<float>(m_pMC->GetCaloHits().size());
```
johnmarshall80, line 814: Check for divide by zero.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+            for (const MCParticle *pMC : allParticles)
+            {
+                // Not all MC particles will have hits
+                if (m_mcToHitsMap.find(pMC) != m_mcToHitsMap.end())
+                {
+                    const CaloHitList &caloHits(m_mcToHitsMap.at(pMC));
+                    allHits.insert(allHits.begin(), caloHits.begin(), caloHits.end());
+                }
+            }
+            m_rootNodes.emplace_back(new Node(*this, allParticles, allHits));
```
johnmarshall80, line 82: Worth running this through valgrind, given heap allocations. (Quick and easy to do, these days.)

johnmarshall80 (commented):
Hi @AndyChappell, Apologies for providing a rather boring review! This looks good to me, and I've just added a few musings as I went through. Feel free to ignore many/most as you look through them.

I think the clang format looks like it needs running again, I wondered about running through valgrind, given the heap allocations, and I think Leigh already highlighted the key implementation blocks to aim to simplify and shorten if possible (sometimes there's a limit to how much things can be simplified, of course).

It'll be good to see this in action, if you happen to have a handy XML snippet to test-drive this? I don't see any problems with merging soon, though, given that this is a new strand of analysis code, and not core pattern-recognition code.

Thanks again!

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+        class Node
+        {
+        public:
+            /**
+             *  @brief  Create a node with a primary MC particle
+             *
+             *  @param  hierarchy The parent hierarchy of this node
+             *  @param  pMC The primary MC particle with which this node should be created
+             */
+            Node(const MCHierarchy &hierarchy, const pandora::MCParticle *pMC);
```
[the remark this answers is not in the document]
AndyChappell: pMCParticle changed. Re const pointer to const ref, I considered that, but ultimately, given that we would need access to the pointer again for containers I figured it probably made sense to stick with the form found most often throughout the codebase.

Regarding hierarchy naming, I do like it as is, given that you never find RecoHierarchy being accessed within the MCHierarchy and vice versa, it's only in client code that the two need distinguishing.

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+             */
+            bool IsTestBeamParticle() const;
+
+            /**
+             *  @brief  Check if this is a cosmic ray particle
+             *
+             *  @return Whether or not this is a cosmic ray
+             */
+            bool IsCosmicRay() const;
+
```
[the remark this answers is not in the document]
AndyChappell: Done

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+            /**
+             *  @brief  Produce a string representation of the hierarchy
+             *
+             *  @return The string representation of the hierarchy
+             */
+            const std::string ToString(const std::string &prefix) const;
+
+        private:
+            const MCHierarchy &m_hierarchy;
```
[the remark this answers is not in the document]
AndyChappell: Will add the docs, clang-format hasn't been run on these files at all yet, so that's on the TODO list

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+         *          particles above the shower-like particle will be folded into the primary node, but a new, daughter leaf node will be
+         *          created for the shower-like particle and all of its daughters, and a parent-child relationship will be formed between
+         *          the primary node and shower node.
+         *
+         *  @param  mcParticleList The list of MC particles with which to fill the hierarchy
+         *  @param  caloHitList The list of hits with which to fill the hierarchy
+         *  @param  foldToPrimaries Whether or not to fold daughter particles back to their primary particle
+         *  @param  foldToLeadingShowers Whether or not to fold daughter particles back to their leading shower particle
+         */
+        void FillHierarchy(const pandora::MCParticleList &mcParticleList, const pandora::CaloHitList &caloHitList,
```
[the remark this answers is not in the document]
AndyChappell: Thanks. I figured it would be useful to provide a bit more of a conceptual overview of how everything works.

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+            const bool foldToPrimaries, const bool foldToLeadingShowers);
+
+        // Note - likely want to change this to retrieve the neutrino, the test beam particle, all primaries, all cosmics etc
```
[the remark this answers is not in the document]
AndyChappell: I've now removed this comment. I was imagining a slightly different implementation when I wrote this, which no longer applies. That said, it might nonetheless be useful to provide additional functions to extract only cosmics, or only test beam particles - I'm undecided on this, as it's actually not needed from the perspective of matching, so I've omitted it for now, but could add this in the future if people think it's particularly useful.

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+         */
+        bool IsNeutrinoHierarchy() const { return m_pNeutrino != nullptr; }
+
+        /**
+         *  @brief  Check if this is a test beam hierarchy.
+         *
+         *  @return Whether or not this is a test beam hierarchy.
+         */
+        bool IsTestBeamHierarchy() const { return m_pNeutrino == nullptr; };
+
```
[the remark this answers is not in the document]
AndyChappell: This is really about distinguishing the neutrino case from the test beam case, rather than identifying separate neutrino, testbeam and cosmics hierarchies. The cosmics hierarchy can coexist with both of these cases.

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+        /**
+         *  @brief  Constructor
+         *
+         *  @param  pMC The MCParticle being matched
+         */
+        MCMatches(const MCHierarchy::Node *pMC);
+
+        virtual ~MCMatches() = default;
+
+        // Might be better to pass in the shared hits themselves
```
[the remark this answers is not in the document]
AndyChappell: Not needed, so comment removed

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+        const RecoHierarchy::NodeVector &GetRecoMatches() const { return m_recoNodes; };
+
+        /**
+         *  @brief  Retrieve the number of shared hits in the match
+         *
+         *  @param  pReco The reco node to consider
+         *
+         *  @return The number of shared hits
+         */
+        int GetSharedHits(const RecoHierarchy::Node *pReco) const;
```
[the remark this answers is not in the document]
AndyChappell: Switched to unsigned

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+         *  @brief  Retrieve the completeness of the match
+         *
+         *  @param  pReco The reco node to consider
+         *
+         *  @return The completeness of the match
+         */
+        float GetCompleteness(const RecoHierarchy::Node *pReco) const;
+
+        private:
+            const MCHierarchy::Node *m_pMC; ///< MC node associated with any matches
```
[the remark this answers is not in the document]
AndyChappell: Although the clang format hasn't been applied yet, I think this is actually what clang will do when applied

larpandoracontent/LArHelpers/LArHierarchyHelper.h
```diff
@@ -0,0 +1,629 @@
...
+     *
+     *  @param  mcHierarchy The MC hiearchy
+     *  @param  recoHierarchy The reconstructed hierarchy
+     *  @param  matchInfo The output match information
+     */
+    static void MatchHierarchies(const MCHierarchy &mcHierarchy, const RecoHierarchy &recoHierarchy, MatchInfo &matchInfo);
+
+private:
+    typedef std::set<const pandora::MCParticle*> MCParticleSet;
+    typedef std::set<const pandora::ParticleFlowObject*> PfoSet;
```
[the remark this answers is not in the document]
AndyChappell: No, this is the first such definition (partly why they're private, in case we do want to define them alongside similar types to be more generally available)

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+            primaries.erase(std::remove_if(primaries.begin(), primaries.end(), predicate), primaries.end());
+        for (const MCParticle *pPrimary : primaries)
+        {
+            MCParticleList allParticles{pPrimary};
+            if (!m_recoCriteria.m_removeNeutrons)
+            {
+                LArMCParticleHelper::GetAllDescendentMCParticles(pPrimary, allParticles);
+            }
+            else
+            {   // Collect track-like and shower-like particles together, but throw out neutrons and descendents
```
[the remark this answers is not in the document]
AndyChappell: I like this as an approach to comments about entire blocks of code, so unless clang-format does something horrible with it, I'll probably leave it as is.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+    m_hierarchy(hierarchy),
+    m_mainParticle(nullptr),
+    m_pdg{0}
+{
+    if (!mcParticleList.empty())
+    {
+        m_mainParticle = mcParticleList.front();
+        m_pdg = m_mainParticle->GetParticleId();
+    }
+    m_mcParticles = mcParticleList; m_mcParticles.sort();
```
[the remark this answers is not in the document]
AndyChappell: Not sure why I hadn't done that already

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -0,0 +1,1171 @@
...
+    const bool isShower{pdg == E_MINUS};
+    if (foldToLeadingShowers && isShower)
+        LArPfoHelper::GetAllDownstreamPfos(pRoot, allParticles);
+    else
+        allParticles.emplace_back(pRoot);
+
+    CaloHitList allHits;
+    for (const ParticleFlowObject *pPfo : allParticles)
+    {
+        LArPfoHelper::GetCaloHits(pPfo, TPC_VIEW_U, allHits);
```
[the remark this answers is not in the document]
AndyChappell: Done

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
AndyChappell: Switched to std::distance

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -527,29 +493,19 @@ void LArHierarchyHelper::RecoHierarchy::FillHierarchy(const PfoList &pfoList, co
...
             }
         }
     }
     else if (foldToLeadingShowers)
     {
-        // Identify the primaries as the starting point for the hierarchy
-        // We have special handling for the neutrino, probably want something similar for testbeam particles
-        PfoSet primaries;
-        m_pNeutrino = LArHierarchyHelper::GetRecoPrimaries(pfoList, primaries);
         for (const ParticleFlowObject *pPrimary : primaries)
```
johnmarshall80, line 509: Loop over PfoSet.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -809,17 +750,16 @@ float LArHierarchyHelper::MCMatches::GetCompleteness(const RecoHierarchy::Node *
     auto iter{std::find(m_recoNodes.begin(), m_recoNodes.end(), pReco)};
     if (iter == m_recoNodes.end())
         throw StatusCodeException(STATUS_CODE_NOT_FOUND);
-    int index = iter - m_recoNodes.begin();
+    int index = std::distance(m_recoNodes.begin(), iter);
```
johnmarshall80, line 741: const?

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -33,30 +35,30 @@ LArHierarchyHelper::MCHierarchy::~MCHierarchy()
...
         }
-        catch(const StatusCodeException&)
+        catch (const StatusCodeException &)
         {
         }
     }
 
+    MCParticleSet primarySet;
+    m_pNeutrino = LArHierarchyHelper::GetMCPrimaries(mcParticleList, primarySet);
+    MCParticleList primaries(primarySet.begin(), primarySet.end());
```
johnmarshall80, line 57: Presumably want to sort the MCParticleList, after populating it via an unordered set. Should be a useful sort function in relevant helper classes (of LArContent, maybe even SDK).

johnmarshall80:
Latest commit looks good, thanks @AndyChappell. Closing comment from me is to have a quick think about whether you'd rather `LArHierarchyHelper::GetMCPrimaries` and `LArHierarchyHelper::GetRecoPrimaries` returned sorted containers themselves. So this would take the sorting within their implementations. If not, have a quick look through all current usage of these functions to review use of returned unordered containers. After that think and/or check, then this looks good to go. Thanks again!

AndyChappell:
Both of these functions are only called in one place, and they're private functions too, so there's no risk of client code using unsorted collections, as a result I'm inclined to leave the interface as is.

PR #175, opened 2021-06-11
Add algorithm to clean up 3D hits

AndyChappell:
This is a simple algorithm that removes all 3D hits from PFOs in specified lists and removes the corresponding 3D clusters from their specified lists.

Example XML snippet:

```
<algorithm type = "LArPfoHitCleaning">
    <PfoListNames>TrackParticles3D ShowerParticles3D</PfoListNames>
    <ClusterListNames>TrackClusters3D ShowerClusters3D</ClusterListNames>
</algorithm>
```

### DIFF
--- a/larpandoracontent/LArUtility/PfoHitCleaningAlgorithm.cc
+++ b/larpandoracontent/LArUtility/PfoHitCleaningAlgorithm.cc
@@ -9,6 +9,7 @@
 #include "Pandora/AlgorithmHeaders.h"
 
 #include "larpandoracontent/LArHelpers/LArClusterHelper.h"
+#include "larpandoracontent/LArHelpers/LArPfoHelper.h"
 #include "larpandoracontent/LArUtility/PfoHitCleaningAlgorithm.h"
 
 using namespace pandora;
@@ -20,6 +21,7 @@ StatusCode PfoHitCleaningAlgorithm::Run()
 {
     for (unsigned int i = 0; i < m_pfoListNames.size(); ++i)
     {
+        // ATTN - one-to-one correspondance required between PFO and cluster lists
         const std::string &pfoListName{m_pfoListNames.at(i)};
         const std::string &clusterListName{m_clusterListNames.at(i)};
         const PfoList *pList(nullptr);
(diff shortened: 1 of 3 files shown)


larpandoracontent/LArUtility/PfoHitCleaningAlgorithm.cc
```diff
@@ -0,0 +1,69 @@
...
+#include "larpandoracontent/LArUtility/PfoHitCleaningAlgorithm.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+StatusCode PfoHitCleaningAlgorithm::Run()
+{
+    for (unsigned int i = 0; i < m_pfoListNames.size(); ++i)
```
absolution1, line 21: Would you consider using a more descriptive index?

absolution1 (approved):
Hi @AndyChappell 
Nice PR.  I added a single style comment.

One thing I don't quite get is how this technically hooks up with Ed iterative growing alg set.
At a glance, it looks like this alg would remove all 3D hits in all PFOs in the provided input list.  I can't remember off the top of my head, does Ed's alg return a list of -only- the PFOs that need their hits removing?

AndyChappell:
> At a glance, it looks like this alg would remove all 3D hits in all PFOs in the provided input list. I can't remember off the top of my head, does Ed's alg return a list of -only- the PFOs that need their hits removing?

Yeh, it's not doing anything clever here, you have to remake the 3D hits in the iterative algorithm. There's room to optimise the process to focus only on the parts that actually change in Ed's algorithm, but for a first pass it seemed like simple was best.

johnmarshall80 (approved):
Looks good, thanks @AndyChappell. Might want to add an ATTN comment about the assumed one-to-one relationship between pfo list and the list hosting the associated 3D cluster. Would otherwise have to get into the approaches used by e.g. PfoMopUp algs, and searching a list of cluster list names, to be completely comprehensive. There is a LArPfoHelper function that allows 3D clusters to be extracted from a Pfo list, which may simplify implementation file, lines 32-36?

Thanks again.

PR #176, opened 2021-06-16
Feature/monitoring fix

imawby:
Small PR to fix build errors in standalone pandora build with -DPANDORA_MONITORING=OFF

### DIFF
--- a/larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
+++ b/larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
@@ -340,6 +340,7 @@ void MuonLeadingEventValidationAlgorithm::ProcessOutput(
         ///////////////////////////////
         if (m_visualize && useInterpretedMatching)
         {
+#ifdef MONITORING
             std::cout << "MC COSMIC RAY HITS" << std::endl;
             this->PrintHits(cosmicRayHitList, true);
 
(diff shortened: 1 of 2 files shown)


larpandoracontent/LArMonitoring/MuonLeadingEventValidationAlgorithm.cc
```diff
@@ -269,11 +269,14 @@ void MuonLeadingEventValidationAlgorithm::ProcessOutput(
             continue;
 
         // Cosmic ray parameters
-        int ID_CR, nReconstructableChildCRLs(0);
+        int nReconstructableChildCRLs(0), nCorrectChildCRLs(0);
+
+#ifdef MONITORING
+        int ID_CR;
```
johnmarshall80, line 275: Completely separately to the aims of this PR, would suggest that these variables are initialised with some defined values, anything really is better than being undefined.

johnmarshall80 (commented):
Thanks @imawby - the diffs look fine. I wondered whether this algorithm is actually genuinely functional without PandoraMonitoring? E.g. does it print something useful to terminal still, even if it can't write a tree or produce visualisation?

If not, it might be cleaner for the preprocessor #ifdefs to basically knock out the entire class and all its implementation? (If going for that strategy, would have to check that the way the algs are registered with Pandora is all happy, and this might mean it's ultimately better as is). Hope that makes some sense!

Other than that, I just flagged some variables, which could do with explicit initialisation. Thanks again!

imawby:
> Thanks @imawby - the diffs look fine. I wondered whether this algorithm is actually genuinely functional without PandoraMonitoring? E.g. does it print something useful to terminal still, even if it can't write a tree or produce visualisation?
> 
> If not, it might be cleaner for the preprocessor #ifdefs to basically knock out the entire class and all its implementation? (If going for that strategy, would have to check that the way the algs are registered with Pandora is all happy, and this might mean it's ultimately better as is). Hope that makes some sense!
> 
> Other than that, I just flagged some variables, which could do with explicit initialisation. Thanks again!

This validation alg should be analogous to say the neutrino validation alg. So with all of the monitoring stuff switched off the algorithm will still print an output to the terminal telling you how many DRs were correctly reconstructed and their shared hits etc.

PR #177, opened 2021-07-14
Add processes for QGSP_BERT and EM standard physics lists

AndyChappell:
This PR adds process mappings for the QGSP_BERT and EM standard physics lists.


larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -40,15 +40,39 @@ enum MCProcess
...
+    MC_PROC_ANTI_TRITON_INELASTIC,
+    MC_PROC_DEUTERON_INELASTIC,
+    MC_PROC_ELECTRON_NUCLEAR,
+    MC_PROC_PHOTON_NUCLEAR,
+    MC_PROC_KAON_PLUS_INELASTIC,
+    MC_PROC_KAON_MINUS_INELASTIC,
+    MC_PROC_HAD_BERTINI_CAPTURE_AT_REST,
+    MC_PROC_LAMBDA_INELASTIC,
+    MC_PROC_MU_NUCLEAR,
+    MC_PROC_TRITON_INELASTIC
```
johnmarshall80, line 75: Is it worth adding an extra, final option just to allow algs to handle special cases, etc. Not mandatory, but something we've used in the past, e.g. for initialising variables, e.g. `HIT_CUSTOM `, `MC_CUSTOM` etc. [here](https://github.com/PandoraPFA/PandoraSDK/blob/master/include/Pandora/PandoraEnumeratedTypes.h).

johnmarshall80 (approved):
Thanks. Just a single attached comment.

larpandoracontent/LArObjects/LArMCParticle.h
AndyChappell: There is an MC_PROC_UNKNOWN which acts as a default if we see a process we don't recognise.

PR #179, opened 2021-07-14
Fix bug where tracks won't split at max x position

AndyChappell:
This PR addresses a bug where the TrackSplittingTool would not consider splitting a long transverse track extending beyond the maximal x coordinate of associated clusters in the other views.

### DIFF
--- a/larpandoracontent/LArThreeDReco/LArTransverseTrackMatching/TrackSplittingTool.cc
+++ b/larpandoracontent/LArThreeDReco/LArTransverseTrackMatching/TrackSplittingTool.cc
@@ -161,14 +161,24 @@ bool TrackSplittingTool::PassesChecks(ThreeViewTransverseTracksAlgorithm *const 
         const LArPointingCluster pointingCluster2(pAlgorithm->GetCachedSlidingFitResult(particle.m_pCluster2));
         const LArPointingCluster longPointingCluster(pAlgorithm->GetCachedSlidingFitResult(particle.m_pLongCluster));
 
-        const CartesianVector &position1(
+        const CartesianVector &minPos1{
             pointingCluster1.GetInnerVertex().GetPosition().GetX() < pointingCluster1.GetOuterVertex().GetPosition().GetX()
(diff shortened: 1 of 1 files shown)


larpandoracontent/LArThreeDReco/LArTransverseTrackMatching/TrackSplittingTool.cc
```diff
@@ -161,14 +161,24 @@ bool TrackSplittingTool::PassesChecks(ThreeViewTransverseTracksAlgorithm *const
...
+        const CartesianVector &maxPos1{
+            pointingCluster1.GetInnerVertex().GetPosition().GetX() > pointingCluster1.GetOuterVertex().GetPosition().GetX()
+                ? pointingCluster1.GetInnerVertex().GetPosition()
+                : pointingCluster1.GetOuterVertex().GetPosition()};
+        const CartesianVector &maxPos2{
+            pointingCluster2.GetInnerVertex().GetPosition().GetX() > pointingCluster2.GetOuterVertex().GetPosition().GetX()
+                ? pointingCluster2.GetInnerVertex().GetPosition()
+                : pointingCluster2.GetOuterVertex().GetPosition()};
+        CartesianVector position1(isMinX ? minPos1 : maxPos1);
+        CartesianVector position2(isMinX ? minPos2 : maxPos2);
```
johnmarshall80, line 181: These final variables can be const.

johnmarshall80 (approved):
Thanks @AndyChappell. Looks good. One small const-related comment. Presumably doesn't matter that we check `<`, then `>`, as will always proceed in a defined manner still if x values are constructed as floating-point identical? (Such a state of affairs likely protected against anyway, of course.)

AndyChappell:
> Presumably doesn't matter that we check `<`, then `>`, as will always proceed in a defined manner still if x values are constructed as floating-point identical? (Such a state of affairs likely protected against anyway, of course.)

Yes, I believe that's true on both counts - I don't think the algorithm can get this far if the x values are identical, but if they do I don't think there is any reason to favour one option over another, just that the choice be consistent.

PR #180, opened 2021-08-04
Fix vertex asymmetry tools when no shower clusters are present

etyley:
I recently ran into a bug where the standard (non-mva) vertexing would return a `Failure in algorithm` message. I found the issue was with the way the new `AsymmetryFeatureBaseTool` treats events where there are no shower clusters present, specifically in this case where an empty map is passed an exception would be thrown when `at` is called: https://github.com/PandoraPFA/LArContent/blob/603df26b1049c85ab0a7af746b5d279d2abecbb9/larpandoracontent/LArVertex/EnergyKickVertexSelectionAlgorithm.cc#L57-L60
https://github.com/PandoraPFA/LArContent/blob/603df26b1049c85ab0a7af746b5d279d2abecbb9/larpandoracontent/LArVertex/AsymmetryFeatureBaseTool.cc#L35-L36
I added in some checks to avoid this exception being thrown. 

It seems like this issues slipped through the CI as no experiment uses this algorithm any more, but I think it would be good to have some checks somewhere to ensure `PandoraSettings_Master_Standard.xml` runs.


AndyChappell:
Thanks for the update, Ed. Definitely worth covering off the issue to ensure all possible use cases run correctly.

PR #181, opened 2021-08-13
Vertex Refinement Algorithm

henrylay97:
This PR adds a new algorithm to the vertex reconstruction suite which performs a refinement procedure on a vertex / vertex candidate using the directions of nearby clusters. 

An example xml snippet for utilising this algorithm directly after the candidate creation algorithm would be:

```
<algorithm type = "LArVertexRefinement">
  <InputClusterListNames>ClustersU ClustersV ClustersW</InputClusterListNames>
  <InputVertexListName>CandidateVertices3D</InputVertexListName>
  <OutputVertexListName>RefinedVertices3D</OutputVertexListName>
</algorithm>

```

I presented this algorithm and some results from testing it in SBND at the most recent meeting and the slides can be found [here](https://indico.fnal.gov/event/50252/contributions/221554/attachments/145990/186062/Vertex_Refinement_DUNE_UK.pdf).

### DIFF
--- a/larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
+++ b/larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
@@ -215,9 +215,9 @@ void VertexRefinementAlgorithm::GetBestFitPoint(const CartesianPointVector &inte
         G(3 * i + 2, i + 3) = -directions[i].GetZ();
     }
 
-    if ((G.transpose() * G).determinant() == 0)
+    if ((G.transpose() * G).determinant() < std::numeric_limits<float>::epsilon())
     {
-        bestFitPoint = CartesianVector(-999.f, -999.f, -999.f);
+        bestFitPoint = CartesianVector(std::numeric_limits<float>::max(), std::numeric_limits<float>::max(), std::numeric_limits<float>::max());
         return;
     }
 
--- a/larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
+++ b/larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
@@ -1,0 +1,246 @@
+/**
+ *  @file   larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
+ *
+ *  @brief  Implementation of the vertex refinement algorithm class.
+ *
+ *  $Log: $
+ */
+
+#include "Pandora/AlgorithmHeaders.h"
+
+#include "larpandoracontent/LArHelpers/LArClusterHelper.h"
+#include "larpandoracontent/LArHelpers/LArGeometryHelper.h"
+#include "larpandoracontent/LArHelpers/LArPcaHelper.h"
+
+#include "larpandoracontent/LArVertex/VertexRefinementAlgorithm.h"
+
+#include <Eigen/Dense>
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+VertexRefinementAlgorithm::VertexRefinementAlgorithm() :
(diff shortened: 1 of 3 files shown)


larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
```diff
@@ -0,0 +1,246 @@
...
+        G(3 * i, 0) = weights[i];
+        G(3 * i + 1, 1) = weights[i];
+        G(3 * i + 2, 2) = weights[i];
+
+        G(3 * i, i + 3) = -directions[i].GetX();
+        G(3 * i + 1, i + 3) = -directions[i].GetY();
+        G(3 * i + 2, i + 3) = -directions[i].GetZ();
+    }
+
+    Eigen::VectorXd m = (G.transpose() * G).inverse() * G.transpose() * d;
```
AndyChappell, line 220: Do we need to check that G.transpose() * G  is not singular? What's the behaviour of Eigen if it is?

AndyChappell (commented):
The algorithm looks good to me, thanks Henry. Mostly trivial comments, one more substantive one relating to how Eigen would behave in the event of a singular matrix.

larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
henrylay97: So the `inverse()` function returns a matrix of NaNs in the case of singular matrices. I think this was being implicitly caught by physics requirements on the position of the vertex later on but this should definitely be prevent explicitly so I've added a requirement on the determinant in my new commit. 

Note that this includes a vector initialised with -999.f values not 0.f. Despite the above conversation this is deliberate, I want to be absolutely sure that a return due to singular matrices will cause line 187 to fail. (0,0,0) is a physically "sensible" position in most detectors coordinate systems whereas (-999,-999,-999) shouldn't be.

henrylay97:
Thanks for your review Andy :D I have responded to your comments in a new commit and rebased against [feature/v03_24_02](https://github.com/PandoraPFA/LArContent/tree/feature/larpandoracontent_v03_24_02). All minor comments were fine and I've commented separately on the eigen comment.

larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
```diff
@@ -0,0 +1,250 @@
...
+        G(3 * i, 0) = weights[i];
+        G(3 * i + 1, 1) = weights[i];
+        G(3 * i + 2, 2) = weights[i];
+
+        G(3 * i, i + 3) = -directions[i].GetX();
+        G(3 * i + 1, i + 3) = -directions[i].GetY();
+        G(3 * i + 2, i + 3) = -directions[i].GetZ();
+    }
+
+    if ((G.transpose() * G).determinant() == 0)
```
AndyChappell, line 218: I think we probably want to check that the absolute value of the determinant against a numeric_limits epsilon value here rather than identically zero.

larpandoracontent/LArVertex/VertexRefinementAlgorithm.cc
AndyChappell: Is this latter scenario a reason to opt for initialising this particular vector with a set of std::numeric_limits<float>::max() values instead. I agree -999 x 3 is a very unlikely coordinate, but numeric limits should be safer still if we want to avoid doing this with exceptions.

henrylay97: Yep that's a good shout!

PR #185, opened 2021-11-01
Updates required for the Range Loop Optimisation warnings from modern compilers

steverdennis:
Natsumi Taniuchi reported being unable to build with Clang 13 because of it adds a number of warnings when using range-based for.

I've fixed this on both clang 12 and gcc 11.2, but because I don't have a Mac I'm unable to test with the same version as her.


AndyChappell:
I've cherry-picked the Clang update in another pull request, and the GCC updates are now redundant due to changes in that same PR, so I'm closing this PR.

PR #193, opened 2022-05-30
Integration of vertex finding neural network

AndyChappell:
This PR replaces the existing candidate vertex creation and BDT-based selection with a neural network-based approach. The change represents a significant improvement in the vertex resolution in the context of accelerator neutrinos and leads to downstream improvements in particle reconstruction efficiency across almost all interaction modes.

The move away from the BDT to the network-based approach also yields improvements in computation time, with the complete Pandora pattern recognition step seeing a 30% improvement in reconstruction time, though at a somewhat higher peak memory requirement, up to ~1.1GB from ~0.9GB maximum resident set size (this is still comfortably within the 2GB grid slot size).


larpandoracontent/LArMonitoring/VertexAssessmentAlgorithm.cc
```diff
@@ -0,0 +1,209 @@
...
+            PANDORA_MONITORING_API(VisualizeCaloHits(this->GetPandora(), pCaloHitListU, "hits_u", BLACK));
+        }
+        if (pCaloHitListV && !pCaloHitListV->empty())
+        {
+            PANDORA_MONITORING_API(VisualizeCaloHits(this->GetPandora(), pCaloHitListV, "hits_v", BLACK));
+        }
+        if (pCaloHitListW && !pCaloHitListW->empty())
+        {
+            PANDORA_MONITORING_API(VisualizeCaloHits(this->GetPandora(), pCaloHitListW, "hits_w", BLACK));
+        }
```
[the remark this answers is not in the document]
AndyChappell: No, it's the PANDORA_MONITORING_API macro, which evaluates to nothing if monitoring is switched off in the build, so you need to protect against that possibility.

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
```diff
@@ -0,0 +1,932 @@
...
+
+//-----------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode DlVertexingAlgorithm::CompleteMCHierarchy(const LArMCParticleHelper::MCContributionMap &mcToHitsMap, MCParticleList& mcHierarchy)
+    const
+{
+    try
+    {
+        for (const auto [ mc, hits ] : mcToHitsMap)
+        {   (void) hits;
```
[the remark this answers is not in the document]
AndyChappell: Right, older versions of GCC moan about this (it'll be able to go when we stop supporting these older versions).

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
```diff
@@ -0,0 +1,932 @@
...
+{
+    if (m_trainingMode)
+        return this->Train();
+    else
+        return this->Infer();
+
+    return STATUS_CODE_SUCCESS;
+}
+
+StatusCode DlVertexingAlgorithm::Train()
```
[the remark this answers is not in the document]
AndyChappell: Could be renamed

larpandoracontent/LArHelpers/LArVertexHelper.cc
```diff
@@ -58,4 +59,19 @@ LArVertexHelper::ClusterDirection LArVertexHelper::GetClusterDirectionInZ(
...
+    {
+        const float x{vertex.GetX()};
+        const float y{vertex.GetY()};
+        const float z{vertex.GetZ()};
+        return -310 < x && x < 310 && -550 < y && y < 550 && 50 < z && z < 1244;
+    }
+    else
+        return false;
+}
+
```
CrossR, line 76: I'm guessing this is hard-coded because the fiducial volume cuts are less well-defined (and mostly exist in random analysis modules, rather than actual geometry level properties)?

Would it be difficult to swap to include a status code, or something here? Returning `false` right now means "not in fiducial volume or the detector string wasn't valid and a silent failure occurred", which could cause issues either for other detectors in the future, or just cause some annoying debugging later due to a dodgy typo in that string.

AndyChappell: Good points all round, I can update it to provide a more flexible return value (and it would be nice to have FV definition come from an upstream source in the future - I'm not aware of any such source right now).

larpandoracontent/LArHelpers/LArVertexHelper.cc
```diff
@@ -58,4 +59,21 @@ LArVertexHelper::ClusterDirection LArVertexHelper::GetClusterDirectionInZ(
     throw StatusCodeException(STATUS_CODE_FAILURE);
 }
 
+//-----------------------------------------------------------------------------------------------------------------------------------------
+
+bool LArVertexHelper::IsInFiducialVolume(const CartesianVector &vertex, const std::string &detector)
+{
+    if (detector == "dune_fd_hd")
```
absolution1, line 66: `==` uses `compare(string)` under the hood so you could save 0.000001 CPU cycles by using that.

larpandoracontent/LArHelpers/LArVertexHelper.cc
```diff
@@ -58,4 +59,21 @@ LArVertexHelper::ClusterDirection LArVertexHelper::GetClusterDirectionInZ(
...
+    {
+        const float x{vertex.GetX()};
+        const float y{vertex.GetY()};
+        const float z{vertex.GetZ()};
+        return -310 < x && x < 310 && -550 < y && y < 550 && 50 < z && z < 1244;
```
absolution1, line 71: Can these be defined against the detector boundaries?  This would defend against using a larger geometry.

Also I guess the magic numbers should be floats.

larpandoracontent/LArMonitoring/VertexAssessmentAlgorithm.cc
```diff
@@ -0,0 +1,209 @@
...
+                        PANDORA_MONITORING_API(FillTree(this->GetPandora(), "vertex"));
+                    }
+                }
+            }
+        }
+        else
+        {
+            if (truthAvailable && isReconstructable)
+            {   // Failed to find a vertex, apply large distance penalty
+                const float dr{5000.f};
```
absolution1, line 165: why {}

larpandoracontent/LArMonitoring/VertexAssessmentAlgorithm.cc
```diff
@@ -0,0 +1,209 @@
...
+                PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "vertex", "dr", dr));
+                PANDORA_MONITORING_API(FillTree(this->GetPandora(), "vertex"));
+            }
+        }
+    }
+    else
+    {
+        if (truthAvailable && isReconstructable)
+        {   // Failed to find a vertex, apply large distance penalty
+            const float dr{5000.f};
```
absolution1, line 175: why {}

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
```diff
@@ -0,0 +1,932 @@
...
+    m_trainingOutputFile{""},
+    m_event{-1},
+    m_pass{1},
+    m_height{256},
+    m_width{256},
+    m_maxHitAdc{500.f},
+    m_regionSize{32.f},
+    m_visualise{false},
+    m_writeTree{false},
+    m_rng(static_cast<std::mt19937::result_type>(0))
```
absolution1, line 38: Does the (0) here mean the seed is always 0?

larpandoracontent/LArMonitoring/VertexAssessmentAlgorithm.cc
absolution1: Actually I can see you've used this everywhere

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
```diff
@@ -0,0 +1,932 @@
...
+        inputs.push_back(input);
+        LArDLHelper::TorchOutput output;
+        if (isU)
+            LArDLHelper::Forward(m_modelU, inputs, output);
+        else if (isV)
+            LArDLHelper::Forward(m_modelV, inputs, output);
+        else
+            LArDLHelper::Forward(m_modelW, inputs, output);
+
+        double thresholds[]{0., 0.00275, 0.00825, 0.01925, 0.03575, 0.05775, 0.08525, 0.12375, 0.15125, 0.20625, 0.26125, 0.31625, 0.37125,
```
absolution1, line 185: Do these need to be hardcoded?  If so, could they be stored in another file so the maintenance is (a little bit) easier.

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
```diff
@@ -0,0 +1,932 @@
...
+        const float x{pCaloHit->GetPositionVector().GetX()};
+        const float z{pCaloHit->GetPositionVector().GetZ()};
+        if (x < xMin)
+            xMin = x;
+        if (x > xMax)
+            xMax = x;
+        if (z < zMin)
+            zMin = z;
+        if (z > zMax)
+            zMax = z;
```
absolution1, line 620: Does std::min/std::max work here?

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
```diff
@@ -0,0 +1,932 @@
...
+
+    // Avoid unreasonable rescaling of very small hit regions
+    const float xRange{xMax - xMin}, zRange{zMax - zMin};
+    if (2 * xRange < m_width)
+    {
+        const float padding{0.5f * (0.5f * m_width - xRange)};
+        xMin -= padding;
+        xMax += padding;
+    }
+    if (2 * zRange < m_height)
```
absolution1, line 662: 2.f?

absolution1 (commented):
Looks good to me Andy.  I think my only decent comments are related to magic numbers.

larpandoracontent/LArMonitoring/VertexAssessmentAlgorithm.cc
AndyChappell: I generally prefer {} initialisation to protect against narrowing conversions

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
AndyChappell: It does

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
AndyChappell: It ought to be possible to make these configurable and read them from XML

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
AndyChappell: It ensures training is reproducible (whether this is genuinely important... maybe?)

larpandoradlcontent/LArVertex/DlVertexingAlgorithm.cc
absolution1: Is all of the training produced in a single job?  Or are multiple jobs used where the seed is the same?

AndyChappell: Good point, it's multiple jobs, so there will be some seed sharing going on here

PR #199, opened 2023-01-16
Use the ordered list directly for GetClosestPosition.

CrossR:
John asked me to have a look over getting Pandora running inside of Intel VTune, to perform a bit of profiling and see where we are at (I'll write up some instructions on that soon, as I was running it locally). As part of that, I spotted this easy to change function that is taking up a weirdly large amount of the cosmic ray reco time (at least in the PDSP data I was using).

Previously, this function could use around 20% of the CPU time in CR reco (on a small sample, but still). This is because calling `FillCaloHitList` means we copy over every calo hit in the cluster into a new list, which in my tests took around 15% of the total CPU time for 10 events.

Similarly, that list is then thrown away straight after, with all the deallocations taking another 5%.

Since this function doesn't actually need that copy at all, just add a variant of the function that uses the ordered calo hit list directly, avoiding the issue here, and helping speed up the CR reco.

I've kept both versions of the function in (calo hit list vs ordered calo hit list) as many places (I assume) are using the old function, but I've changed the `Cluster` based function to stop all the copying, which should help in a few places.


CrossR:
Oh, and if anyone is interested what the current bottlenecks are / what sort of output you can get out of VTune:

vTune (1)

That shows a list of the current hotspots (before this fix is applied), for 10 PDSP events running in Cosmic Ray reco only. In the GUI you can then dig into the exact lines calling those bits of code etc.

Issue #197, opened 2022-12-05
Bug in LArControlFlow/NeutrinoIdTool.cc (or maybe better phrased as sub-optimal way of failing?)


brucehoward-physics:
Hi all!

I found a bit of a bug in LArControlFlow/NeutrinoIDTool.cc (or, maybe really this should be described as a sub-optimal way of failing).

When it goes to persist the slice ID features in the metadata, it calls `sliceFeaturesVector.at(sliceIndex).GetFeatureMap(featureMap);`

In the case where the features are not properly filled, then it throws `StatusCodeException(STATUS_CODE_NOT_FOUND)`.

Some of the code has a better handling of this issue, for example the `GetNeutrinoProbability` function has this:
```
if (!this->IsFeatureVectorAvailable())
        return 0.f;
```

When we go to fill the metadata, can we change the if statements to actually check if the feature vector is available, and just not fill the metadata for that if so, rather than throwing an exception? Does this cause any other issues? Something like
```
if (m_persistFeatures)
```
becoming
```
if (m_persistFeatures && sliceFeaturesVector.at(sliceIndex).IsFeatureVectorAvailable())
```

brucehoward-physics:
Just checking in -- any thoughts here?

johnmarshall80:
Hi @brucehoward-physics - Thanks for the reminder about this! I don't think the overall processing should differ whether `m_persistFeatures` is set to `true` or set to `false`, excepting (of course) for the accompanying feature persistency. So, you should then be good to make any associated changes so that the processing/outcome is unchanged by `m_persistFeatures` _and_ so that you're actually happy with what is persisted, in terms of the values provided as a matter of course, vs. some placeholder/default/identifiable value, etc. that you persist if the feature calculation had somehow failed. I hope that makes sense to you. Thanks again.

brucehoward-physics:
Hi @johnmarshall80 - I believe what you say makes sense to me, though I suppose I am also double checking that others agree this can be a problem that we should address and/or if folks had a better solution in mind than what I propose.

I suppose in my potential "solution," there is no alternative default that is saved, but one could pick this up at the analysis level by noting the lack of a saved value. Unless this behavior is not desired broadly? (in which case maybe any solution should save some default values where needed)

AndyChappell:
Resolved in larpandoracontent PR [50](https://github.com/LArSoft/larpandoracontent/pull/50).

PR #200, opened 2023-01-20
Add a check on feature vector being available before trying to get th…

brucehoward-physics:
…e map and persist.

Related to https://github.com/PandoraPFA/LArContent/issues/197. Continuing discussion here.

Perhaps we need defaults to fill the metadata instead?

### DIFF
--- a/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
+++ b/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
@@ -257,7 +257,7 @@ void NeutrinoIdTool<T>::SelectPfosByProbability(const pandora::Algorithm *const 
             object_creation::ParticleFlowObject::Metadata metadata;
             metadata.m_propertiesToAdd["NuScore"] = nuProbability;
 
-            if (m_persistFeatures)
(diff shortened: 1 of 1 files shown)


AndyChappell:
Hi Bruce, thanks for the update. My initial reaction is that the proposed solution looks reasonable. As for default values, is there a scenario in which you would look up values for properties when a feature vector is not properly filled?

brucehoward-physics:
I think we (SBN) handle this at analysis file creation, where we fill the value for the property if it can find it in the map, otherwise it should remain at a default set in our code.

Not sure if there are considerations with other experiments, use cases, or within Pandora itself? I had talked to @absolution1 and he was thinking a little about the possibility for values to be used in different purposes where a default might or might not be useful, not sure if he has had further thoughts.

brucehoward-physics:
checking in on status/thinking :)

AndyChappell:
> checking in on status/thinking :)

Did @absolution1 have anything to add here re default values?

absolution1:
Hello.
My feelings on default values was mostly related to cases where you'd want all of the values to be provided (such as BDT training) but it is possible to handle/create default values in user scripts.

AndyChappell:
Issue addressed in another PR

PR #203, opened 2023-03-07
Basic Pandora GitHub Actions Config

CrossR:
This implements three things:

 - Automated building of LArContent (and all required deps) on any PR, commit to the main branch, or manual running. It is built using either GCC or Clang, with or without PandoraMonitoring and with or without LibTorch.
 - Once a week, build and upload a coverity result.
 - Verify that `clang-format` is happy on commits to main, PRs, manual runs etc. For now, I've disabled this one, as parts of the repo still aren't compliant, so we can do manual runs only. Once that is fixed, we can enable it and have automated checks for code style in all PRs.

Steps following the merge:

 - GitHub Actions needs enabling per-repo from my understanding. It would perhaps be useful if @AndyChappell could merge this PR to his fork? That would allow checking this for sure, as GitHub do have a habit of changing defaults over time (would also allow the setup to be tested outside the main repo).
 - Once enabled, there is a second "Actions Secret" tab in the repo setting that needs the coverity email / repo and scan token adding in there. I believe John should have access to that in the Coverity pages.
 - Fix the build issues. There is currently a few build issues, mostly around clang and builds without Pandora Monitoring. Andy has a branch that fixes most of them, so would be good once this is merged to get that brought in. The first initial builds will fail because of this, but that isn't the end of the world and we will be able to see the change from failure to success.
 - At some point, should work out if we want to clang-format the full repo and get automated checking added.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,182 @@
+name: Coverity
+
+# Controls when the workflow will run
+on:
+  # Trigger the workflow based on cron.
+  schedule:
+    - cron: '0 0 * * 6'
+  workflow_dispatch:
+
+# Workflow with 1 stage: Build.
+# Each of the sub-stages in the build job run in parallel.
+jobs:
+  build:
+
+    # Use the latest ubuntu image: https://github.com/actions/runner-images
+    runs-on: ubuntu-latest
+    name: LArContent - Coverity
+
+    # Defines the build matrix, so what combinatorics of compiler etc. to test.
+    strategy:
+      fail-fast: false # Don't quit other jobs if one job fails.
+      matrix:
+        compiler: [ {cpp: g++-9, c: gcc-9} ]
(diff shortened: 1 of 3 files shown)


CrossR:
Annoyingly, looks like GitHub have changed their stance on "You should opt into the CI", as can be seen by the fact its started running straight away here. Seems a bit of an odd choice, a stranger could open a PR against any repo and start using up their CI allowance (2000 mins per month but still).

So, before this is merged, I'll need to add an `if: github.repository == 'PandoraPFA/LArContent'` check to the various stages, to stop it running outside of this repo.

CrossR:
The conditional to only run in the PandoraPFA repo has been added now, and I've confirmed with the SciSoft team there shouldn't be any issue with them eventually picking this up (they'll get a "Skip Job" UI, as seen here: https://github.com/CrossR/LArContent/pull/9, but nothing actually runs etc).

PR #207, opened 2023-04-19
Make default 'NuScore' configurable

henrylay97:
Original message in slack

> I've been digging into an issue in SBND where we were missing persisted slice ID variables for "neutrino" labelled slices. I've traced it down to [these](https://github.com/LArSoft/larpandoracontent/blob/develop/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc#L482-L484) lines in the NeutrinoID tool which gives slices for which not all MVA variables could be calculated a default value of 0. Because SBND runs with the default minimum probability of 0 [this](https://github.com/LArSoft/larpandoracontent/blob/develop/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc#L289) line still fails and the slice is still persisted as a "neutrino slice", however, unlike all other neutrino slices we don't get any persisted input variables.
>
>Clearly there are catches I can add for this in our use cases. But I wondered whether there was any particular reason we wanted it to be 0.f ? Would a value of -1.f work in this scenario? It would nicely indicate to analysers that something went wrong with the calculation of the score, and would simultaneously ensure that it wasn't saved as a neutrino slice? (edited)

This PR aims to introduce the ability to set a different default score for the "neutrino probability" or nuscore created in the slice identification stage. Standard behaviour will not change for any experiment. SBND will introduce an xml change to activate this.

### DIFF
--- a/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
+++ b/larpandoracontent/LArControlFlow/NeutrinoIdTool.cc
@@ -33,6 +33,7 @@ NeutrinoIdTool<T>::NeutrinoIdTool() :
     m_minPurity(0.9f),
     m_minCompleteness(0.9f),
     m_minProbability(0.0f),
+    m_defaultProbability(0.0f),
     m_maxNeutrinos(1),
     m_persistFeatures(false),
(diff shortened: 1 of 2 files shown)


AndyChappell:
Looks like a sound change to me thanks, Henry.

PR #214, opened 2023-07-12
Feature/lhw extract base classes: extract various classes and base classes from other header files

lhwhitehead:
This pull request contains a number of changes that were required for developing some algorithms for the near detector. The `Slice` class was originally defined in the slicing algorithm header file, but the ND requires a modified version of the slice, so the decision was taken to move the `Slice` class into its own header file `LArSlice.h`. Various similar changes have been made along these lines, moving various base classes out from algorithm header files in order to make things more easily extendable in the future.


AndyChappell:
Thanks for the update Leigh. Everything looks good to me. We do have another PR going through the LArSoft release cycle at the moment, so once that's done we can rebase this and merge it into a release branch for a follow up PR.

PR #215, opened 2023-08-09
Allow DLVertexing to use different driftStep & volumeType for the ND

jback08:
The default values for the driftStep and volumeType for the Far Detector are 5 mm & "dune_fd_hd", respectively. These can be changed using the "DriftStep" & "VolumeType" LArDLVertexing xml settings.

Modified LArVertexHelper::IsInFiducialVolume to expect "dune_nd" for the Near Detector, which just checks that the vertex point is inside the TPC.


johnmarshall80 (approved):
Hi @jback08 - this looks good to me, thanks, with a thought that we should consider adding a string to (optionally) hold names for `pandora::LArTPC` instances. These are created via the Pandora app (`LArRecoND` for ND usage, of course) and so could feed in some name that could then be used to trigger the correct behaviour in the `LArVertexHelper` and the calling algorithm, rather then needing to specify geometry type as an algorithm parameter. What do you think @AndyChappell? Thanks again!

AndyChappell:
> What do you think @AndyChappell? Thanks again!

Yes, it would be good to eliminate the XML specification in a future release.

PR #218, opened 2024-02-01
Slice Metrics

CrossR:
I've spun this out into its own PR, mostly to make it easier on Matt, but it could be generally useful to get in anyways.

That said...its marked as a draft as I currently have to do a few things, some iffy, some **very very bad**.
Essentially, to make the slicing metrics somewhat useful, you need to have the full event context:

 - You can't have hits missing due to clear cosmic removal.
 - You can't only run inside your singular slice (i.e. Nu or CR pass)

For that reason, it necessitates some dumb decisions (well.....it was easiest to do it in a dumb way at least).
I think its fine for this to be used like this, but it is obviously not good to keep it around, so we should work out a) if we want this feature of keeping every hit around and b) how best to implement that.

### DIFF
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -50,6 +50,7 @@ MasterAlgorithm::MasterAlgorithm() :
     m_pSliceCRWorkerInstance(nullptr),
     m_fullWidthCRWorkerWireGaps(true),
     m_passMCParticlesToWorkerInstances(false),
+    m_passAllCaloHitsToWorkers(false),
     m_filePathEnvironmentVariable("FW_SEARCH_PATH"),
     m_inTimeMaxX0(1.f)
 {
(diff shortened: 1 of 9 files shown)


larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -654,6 +682,10 @@ StatusCode MasterAlgorithm::Copy(const Pandora *const pPandora, const CaloHit *c
     }
     LArCaloHitParameters parameters;
     pLArCaloHit->FillParameters(parameters);
+
+    if (ignoredHit)
+        parameters.m_nCellInteractionLengths = -999.f;
+
```
CrossR, line 688: This is the core of why this is implemented in a very hacky way right now...

I want to pass over every single calo hit, but I don't want the downstream algorithms to actually use all these hits, I just want them to carry on as normal.

My "fix" was to hijack a seemingly unused var and use that to flag in the pre-processing "Hey this is a hit you should not actually use, just store it to one side".

larpandoracontent/LArMonitoring/SliceMonitoringTool.cc
```diff
@@ -0,0 +1,361 @@
...
+    LArMCParticleHelper::MCContributionMap mcToTrueHitListMap;
+
+    // Since the slice hits, and the full, unfiltered hits may not match up, first
+    // perform a quick match between these two sets of hits.
+    std::map<const std::tuple<float, float, float>, const CaloHit*> caloHitListMatchMap;
+    auto getHitKey = [](const CaloHit* pCaloHit) -> std::tuple<float, float, float> {
+        const auto pos = pCaloHit->GetPositionVector();
+        return {pos.GetX(), pos.GetZ(), pCaloHit->GetHadronicEnergy()};
+    };
+    CaloHitList matchedCaloHitList;
```
CrossR, line 74: This is also a knock on impact of that strange hack.

The "full hit list" and the "actual hit list" are different underlying hits.
Since there isn't a way for Pandora to know they are the same (say if the hits had a built in `uuid` or something), I have to do a somewhat iffy map to go from "hit in full hit list" to "hit in actual hit list".

Its awkward, and ideally people wouldn't need to do that (assuming again, we want to keep a "full hit list" around).

PR #223, opened 2024-06-10
Update sliding cone fits for mopup

AndyChappell:
Updated sliding cone algorithm to allow for greater configurability in cone structure.

### DIFF
--- a/larpandoracontent/LArObjects/LArThreeDSlidingConeFitResult.cc
+++ b/larpandoracontent/LArObjects/LArThreeDSlidingConeFitResult.cc
@@ -132,8 +132,8 @@ void ThreeDSlidingConeFitResult::GetSimpleConeList(
     TrackStateLinkedList trackStateList;
     CartesianVector directionSum(0.f, 0.f, 0.f);
     const float clusterLength((trackStateMap.begin()->second.GetPosition() - trackStateMap.rbegin()->second.GetPosition()).GetMagnitude());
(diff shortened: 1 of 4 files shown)


larpandoracontent/LArObjects/LArThreeDSlidingConeFitResult.cc
```diff
@@ -131,6 +132,8 @@ void ThreeDSlidingConeFitResult::GetSimpleConeList(
     TrackStateLinkedList trackStateList;
     CartesianVector directionSum(0.f, 0.f, 0.f);
     const float clusterLength((trackStateMap.begin()->second.GetPosition() - trackStateMap.rbegin()->second.GetPosition()).GetMagnitude());
+    const float lengthStep{0.5f * clusterLength / (nCones - 1)};
+    const float angleStep{0.5f * tanHalfAngle / (nCones - 1)};
```
absolution1, line 136: Is there any defence against nCones==1 in the calling function?

absolution1 (commented):
Nice one @AndyChappell 
I only had a single trivial comment.

larpandoracontent/LArObjects/LArThreeDSlidingConeFitResult.cc
AndyChappell: Actually no... the default value in the main algorithm is 5, but this can be overridden in XML, so I should add a check there

PR #225, opened 2024-08-06
HierarchyValidationAlgorithm: handle events with missing vertices & other updates

jback08:
Changes to the HierarchyValidationAlgorithm:

Do not skip events if there are problems accessing the PFO vertex or interaction code information (handles exceptions using try-catch).

Allow both the Event and MC hierarchy ROOT files to be created together if enabled, instead of only being either one or the other; they both use the same hierarchy matching info. Added the following xml parameters:

`ValidateEvent (false), WriteEventTree (false), EventFileName, EventTreeName`
`ValidateMC (false), WriteMCTree (false), MCFileName, MCTreeName`

Allow the reconstructability criteria to be modified by xml parameters:
`MinRecoHits (30), MinRecoHitsPerView (10), MinRecoGoodViews (2) & RemoveRecoNeutrons (true)`

Add reco and true MC vertex coordinates to the ROOT output.


AndyChappell (approved):
Thanks for the updates @jback08 . I'll look to get this merged into a release branch for the next LArSoft cycle.

PR #227, opened 2024-08-18
 LArHierarchyHelper: allow selecting reco hits that overlap with the MC particle hits

jback08:
The MicroBooNE [NeutrinoEventValidationAlgorithm](https://github.com/PandoraPFA/LArContent/blob/feature/larpandoracontent_v04_10_00/larpandoracontent/LArMonitoring/NeutrinoEventValidationAlgorithm.cc) requires that the reco hits from the PFOs overlap with the hits from the MC particles for the reco-MC matching. This is done by the [LArMCParticleHelper::SelectReconstructableMCParticles()](https://github.com/PandoraPFA/LArContent/blob/feature/larpandoracontent_v04_10_00/larpandoracontent/LArHelpers/LArMCParticleHelper.cc#L546) function, which eventually calls [LArMCParticleHelper::CollectReconstructable2DHits()](https://github.com/PandoraPFA/LArContent/blob/feature/larpandoracontent_v04_10_00/larpandoracontent/LArHelpers/LArMCParticleHelper.cc#L1064). Any reco hits with Ids that are not in the MC hit list are left out, which generally increases the match purities by reducing their denominators.

This functionality has been added to `LArHierarchyHelper` to give an equivalent MicroBooNE-like MC-selection of reco hits for the hierarchy tools. The `LArHierarchyHelper::MatchInfo::Match()` process has been modified here to use the `GetSelectedRecoHits()` function to get the list of reco hits with parent addresses that overlap with those from the MC hit list, which is then stored by `AddRecoMatch`. The `SelectedRecoHitsMap` stores the selected hit list for each reco node; this map is needed since the nodes are constant, making it is not possible to update them to include the selected hits during the reco-MC matching process.

The `SelectRecoHits` xml boolean in `HierarchyValidationAlgorithm` enables or disables this procedure, being passed as an extra `QualityCuts` parameter. When set to false (currently the default), all hits in a given reco node are considered and not just those that are within the MC hit list, which generally gives lower purities (especially for particles with secondary interactions).

Presentation file showing the issue: [PR227.pdf](https://github.com/user-attachments/files/16661993/PR227.pdf)

### DIFF
--- a/larpandoracontent/LArHelpers/LArHierarchyHelper.cc
+++ b/larpandoracontent/LArHelpers/LArHierarchyHelper.cc
@@ -1316,7 +1316,7 @@ void LArHierarchyHelper::MatchInfo::Match()
         MCHierarchy::NodeVector mcNodes;
         m_mcHierarchy.GetFlattenedNodes(pRootMC, mcNodes);
 
-	// Get all of the hits from the MC nodes (for selecting reco hits)
+        // Get all of the hits from the MC nodes (for selecting reco hits)
         CaloHitList allMCHits;
         for (const MCHierarchy::Node *pMCNode : mcNodes)
         {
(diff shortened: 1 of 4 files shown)


larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -1301,21 +1315,35 @@ void LArHierarchyHelper::MatchInfo::Match()
...
+        }
+
         for (const ParticleFlowObject *const pRootPfo : rootPfos)
         {
             RecoHierarchy::NodeVector recoNodes;
             m_recoHierarchy.GetFlattenedNodes(pRootPfo, recoNodes);
 
             std::sort(mcNodes.begin(), mcNodes.end(),
-                [](const MCHierarchy::Node *lhs, const MCHierarchy::Node *rhs)
+	        [](const MCHierarchy::Node *lhs, const MCHierarchy::Node *rhs)
```
AndyChappell, line 1333: Looks like this might have been switched to hard tabs, can this be switched back to soft tabs.

larpandoracontent/LArHelpers/LArHierarchyHelper.cc
```diff
@@ -1301,21 +1315,35 @@ void LArHierarchyHelper::MatchInfo::Match()
...
                 { return lhs->GetCaloHits().size() > rhs->GetCaloHits().size(); });
             std::sort(recoNodes.begin(), recoNodes.end(),
-                [](const RecoHierarchy::Node *lhs, const RecoHierarchy::Node *rhs)
+		[](const RecoHierarchy::Node *lhs, const RecoHierarchy::Node *rhs)
```
AndyChappell, line 1336: Looks like this might have been switched to hard tabs, can this be switched back to soft tabs.

AndyChappell (commented):
Aside from some minor formatting requests, this looks good to me, thanks John.

PR #230, opened 2024-11-01
Fix Mac Build + Update CI.

CrossR:
Few small changes to make the Mac build happy:

 - Fix some unused variables.
 - PyTorch/LibTorch on Mac doesn't seem to like the use of `long`, so I've swapped to `int64_t` there....I've been using this locally on mac for the past 12 months or so, but I (and ideally someone else) need to test this on Linux.
 - Link LArDLContent against LArContent, to access the helpers etc. Again, I've used this on Mac for a while, but would be good to test on Linux.

To help prevent this again, I've also bumped the CI to more modern versions (GCC 12 to match LArSoft v10_00_X, and the latest Clang in Ubuntu 24.04), as well as fixing a few other bits in the CI (i.e. bumping the tools we use in the actions to their latest versions, to prevent some deprecation warnings).

The ROOT and LibTorch versions used here should match the versions used by LArSoft V10_00_02.

Whilst testing this in my local branch, I found that when jumping to the newer compiler versions, I got even more "set but not used variable" warnings, so I've fixed them too.

There is maybe also a question of if we want to bother adding a Mac config to the CI....that would catch these errors quickly, but would be an extra thing running each time (and make the CI config a touch more complicated to support picking the right versions of ROOT / LibTorch, setup steps etcetc).

### DIFF
--- a/.github/workflows/formatting.yml
+++ b/.github/workflows/formatting.yml
@@ -32,10 +32,10 @@ jobs:
           - 'larpandoradlcontent'
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Check clang-format
-        uses: jidicula/clang-format-action@v4.10.2
+        uses: jidicula/clang-format-action@v4.13.0
         with:
-          clang-format-version: '13'
+          clang-format-version: '18'
           check-path: ${{ matrix.path }}
 
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -13,7 +13,7 @@ jobs:
   build:
 
     # Use the latest ubuntu image: https://github.com/actions/runner-images
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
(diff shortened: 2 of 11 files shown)


.github/workflows/formatting.yml
```diff
@@ -32,10 +32,10 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Check clang-format
-        uses: jidicula/clang-format-action@v4.10.2
+        uses: jidicula/clang-format-action@v4.13.0
         with:
-          clang-format-version: '13'
+          clang-format-version: '18'
```
CrossR, line 39: This isn't actually actively used...if we want to, we could get the whole code base under `clang-format` compliance, then enable this on PRs.

CrossR:
(PR was done purposely against main here, since I don't believe the CI runs against PRs to other branches in its current form....we could look at changing that to remove an extra step for reviewers? I can change the target after the CI runs at least, I just want to verify everything is happy.)

PR #231, opened 2024-11-19
DL Vertexing Slicing + MicroBooNE Considerations

CrossR:
Two main changes here:

 - In a slicing context, it is technically possible for the DL Vertexing to either not return anything, or return something useless (this will be even more the case if/when we add filter to the DL Vertexing to skip cosmic hits etc). This can happen more frequently for cosmic ray slices. For that reason, its nice to have the existing, SVM-based candidate selection work alongside the newer vertexing. This change simply allows a specific vertexing list to be loaded (not just the current, as that could be set already), and also to stop processing if either a pass 1 vertex wasn't made for the DL vertexing, or if there is any vertex already made, the older vertexing just quits early.

 - The only other change is to make some of the DL Master algorithms methods be protected, not private. This is required for the MicroBooNE wrapper, and matches the functionality of the non-DL master algorithm I think.

### DIFF
--- a/larpandoracontent/LArVertex/CandidateVertexCreationAlgorithm.cc
+++ b/larpandoracontent/LArVertex/CandidateVertexCreationAlgorithm.cc
@@ -47,10 +47,20 @@ StatusCode CandidateVertexCreationAlgorithm::Run()
 {
     try
     {
+        // INFO: See if there is already a vertex, and quit early if there is.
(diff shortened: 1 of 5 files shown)


imawby:
After some discussion with Ryan about the details of the corresponding PandoraSettings xml I'm happy with the PR.

PR #238, opened 2025-02-17
Small updates to 3D reclustering tools

AlexWilkinsonnn:
Some small updates following my testing of the code in PR #228 :

- Handle expected exceptions - the shower pfo list may be empty and the truth matching may not be present for some hits, this was consistently throwing exceptions previously
- Small refactor to ensure old pfos (no longer having any cluster associated with them) do not remain in the pfo list. Track the pfos that have been reclustered, delete them at the end of the reclustering loop and move all the remaining original pfos into the new list.

PR #228, opened 2024-08-30
Add functionality to run 3D clusters reclustering tools

MariaBrigida:
This PR introduces the 3D clusters reclustering mechanics with **ThreeDReclusteringAlgorithm**. This algorithm:

- Reads the list of shower pfos and examines their associated 3D clusters. At the moment tracks are not considered.
- Checks if new clustering outcomes should be proposed (at the moment, this is done via a cheated figure of merit, which is equivalent to an "impurity": it is given by the fraction of 3D hits with parent 2D hits that are not created by the "main" MC particle)
- If the decision to reclusters is positive, loops over a series of algorithm tools, specified via XML, which take the calo hits in the 3D cluster as input, and propose a new set of calo hit lists
- if a new set of calo hit lists contains at least one list that yields a figure of merit that is lower than the initial one, then persist this outcome: create new 3D clusters from this new set of lists.
- Also creates new 2D clusters and new Pfos matching the new 3D clusters
- If no better set of calo hit lists was found, persists the original clusters/pfos
- Does list book-keeping.

At the moment this PR does not support DL features. [A separate branch](https://github.com/MariaBrigida/LArContent/blob/feature/ReclusteringFramework_preparePR_dl/larpandoracontent/LArReclustering/CheatedThreeDClusteringTool.h) has been created with classes in larpandoracontent that enable those.

This PR is to be used with XMLs contained in [this LArReco branch](https://github.com/PandoraPFA/LArReco/compare/master...MariaBrigida:LArReco:feature/ReclusteringFramework_preparePR_2). These are set up to run these tool configurations:
* Only **cheated reclustering tool**: uses truth information on what particle contributed the largest energy to the 2D parent hit of each 3D calo hit, to split calo hits into a new set of lists
* Only **PCA-based tool**: this splits the calo hit lists in two, based on hits projections on secondary axis. The PCA is run twice, first on the original calo hit list, then on the two new lists, to refine the splitting.

PR #240, opened 2025-03-04
Reworked 3D reclustering framework(?)

AlexWilkinsonnn:
Here is the PR for the 3D reclustering algorithm as discussed in today's meeting (https://indico.fnal.gov/event/68520/contributions/311068/attachments/185738/255775/3dreclustering_update_Mar4_25.pdf) . 

`ThreeDMultiReclusteringAlgorithm` is the builk of the PR and is where any comments are most appreciated.

`ThreeDReclusteringFigureOfMeritBaseTool.h` just defines the interface for the figure of merit alg tool to follow and the contents of `LArExamples/` are implementations of random clustering and figure of merit for reference.

This is how the xml settings block may look:
```
    <algorithm type = "LArThreeDMultiReclustering">
        <PfoListName>TrackParticles3D</PfoListName>
        <Cluster3DListName>TrackClusters3D</Cluster3DListName>
        <ClusterUListName>ClustersU</ClusterUListName>
        <ClusterVListName>ClustersV</ClusterVListName>
        <ClusterWListName>ClustersW</ClusterWListName>
        <ClusteringAlgorithms>
            <algorithm type = "LArRandomClustering" instance = "Reclustering1">
                <NumNewClusters>1</NumNewClusters>
            </algorithm>
            <algorithm type = "LArRandomClustering" instance = "Reclustering2">
                <NumNewClusters>4</NumNewClusters>
            </algorithm>
        </ClusteringAlgorithms>
        <tool type = "LArRandomFigureOfMeritTool" description = "FomAlgorithmTool">
            <MaxFomToRecluster>0.5</MaxFomToRecluster>
        </tool>
    </algorithm>
```

As discussed, after getting this merged we will do a clean up/porting over parts of the existing `ThreeDReclusteringAlgorithm` PR (#238 , #228 )

Thanks!


larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.h
```diff
@@ -0,0 +1,150 @@
...
+     *
+     *  @param view hit type
+     *  @param listName output list name
+     */
+    pandora::StatusCode GetClusterListName(const pandora::HitType &view, std::string &listName);
+
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+
+    std::string                              m_pfoListName;       ///< Name of list of pfos to consider for reclustering
+    std::string                              m_cluster3DListName; ///< Name of list of 3D clusters that comprise the pfos
```
lhwhitehead, line 116: Would a map of <HitType, std::string> work a bit more nicely for storing the cluster list names? I think it would at least simplify the GetClusterListName function.

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+                continue;
+            PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, MopUpCaloHits(caloHits2D, *pOldClusters2D, true));
+            continue;
+        }
+        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, MopUpCaloHits(caloHits2D, viewToNewClusters2D.at(view), true));
+    }
+
+    // Create the new Pfos
+    // ATTN New pfos will: have no vertex, not be characterised as track/shower, not be part of any particle hierarchy.
+    //      If desired, these these things will need to be added in succeeding algs.
```
AndyChappell, line 219: Probably worth adding this as a comment against the class definition and/or the Run function to make it more likely to be seen.

AndyChappell (changes requested):
Thanks Alex, this is looking nice. Minor change requests, but otherwise I think this is good to go into the next release.

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    // Ask the FOM alg tool which pfos to recluster
+    PfoList pfosToRecluster;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, m_pFomAlgTool->GetPfosToRecluster(pPfos, pfosToRecluster));
+    if (pfosToRecluster.empty())
+        return STATUS_CODE_SUCCESS;
+
+    // Remove clusters from the pfos, keeping track of their original pfo
+    std::map<HitType, ClusterList> viewToFreeClusters = { {TPC_3D, {}}, {TPC_VIEW_U, {}}, {TPC_VIEW_V, {}}, {TPC_VIEW_W, {}} };
+    std::map<const Pfo *const, ClusterList> pfoToFreeClusters;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, FreeClustersFromPfos(pfosToRecluster, viewToFreeClusters, pfoToFreeClusters));
```
imawby, line 54: I see that the only non STATUS_CODE_SUCCESS is given if the cluster cannot be removed from the pfo - i think this suggests that something has gone quite wrong. Should this result in a throw instead i.e. in FreeClustersFromPfo and then the function can be void.

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode ThreeDMultiReclusteringAlgorithm::FreeClustersFromPfos(const PfoList &pfos,
+                                                                  std::map<HitType, ClusterList> &viewToFreeClusters,
+                                                                  std::map<const Pfo *const, ClusterList> &pfoToFreeClusters)
+{
+    for (const Pfo *const pPfo : pfos)
+    {
```
imawby, line 128: If your logic of your alg (and I dont know if it does) always ignores pfos with no 3D hits, you could just skip over them at this point. Theres a function in PfoHelper that returns this info for you.

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+        }
+    }
+
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::TemporarilyReplaceCurrentList<Cluster>(*this, bestReclusterListName));
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::EndReclustering(*this, bestReclusterListName));
+
+    if (bestReclusterListName == originalClusterListName) // Return the clusters to the original pfos as if nothing happened...
+    {
+        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, AddClustersToPfos(pfoToFreeClusters));
+        return STATUS_CODE_SUCCESS;
```
imawby, line 95: could you not just do `return AddClustersToPfos(pfoToFreeClusters) `

larpandoracontent/LArReclustering/LArExample/RandomClusteringAlgorithm.cc
```diff
@@ -0,0 +1,83 @@
...
+
+RandomClusteringAlgorithm::RandomClusteringAlgorithm()
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+RandomClusteringAlgorithm::~RandomClusteringAlgorithm()
+{
+}
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Added

larpandoracontent/LArReclustering/LArExample/RandomClusteringAlgorithm.cc
```diff
@@ -0,0 +1,83 @@
...
+    for (const CaloHit *const pCaloHit : *pCaloHits)
+    {
+        if (splitItr != splitIdxs.cend() && cntr++ > *splitItr)
+        {
+            const Cluster *pCluster {nullptr};
+            PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::Cluster::Create(*this, params, pCluster));
+            params.m_caloHitList.clear();
+            splitItr++;
+        }
+        params.m_caloHitList.push_back(pCaloHit);
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Changed

larpandoracontent/LArReclustering/LArExample/RandomClusteringAlgorithm.h
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     *  @brief  Default constructor
+     */
+    RandomClusteringAlgorithm();
+
+    /**
+    *  @brief  Destructor
+    */
+    ~RandomClusteringAlgorithm();
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Added

larpandoracontent/LArReclustering/LArExample/RandomFigureOfMeritTool.cc
```diff
@@ -0,0 +1,66 @@
...
+RandomFigureOfMeritTool::RandomFigureOfMeritTool() :
+    m_maxFomToRecluster {0.5f}
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+RandomFigureOfMeritTool::~RandomFigureOfMeritTool()
+{
+}
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Added

larpandoracontent/LArReclustering/LArExample/RandomFigureOfMeritTool.cc
```diff
@@ -0,0 +1,66 @@
...
+
+StatusCode RandomFigureOfMeritTool::GetPfosToRecluster(const PfoList *pPfos, PfoList &pfosToRecluster)
+{
+    if (!pfosToRecluster.empty())
+        return STATUS_CODE_FAILURE;
+
+    for (const Pfo *const pPfo : *pPfos)
+    {
+        if (pPfo->GetNClusters() == 0)
+            continue;
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Ideally no but I did see this occasionally. You are allowed to have an empty pfo so I suppose some preceding algs have either created a pfo and not added anything to it or removed clusters from a pfo and not bothered to delete it off the list :cry:

larpandoracontent/LArReclustering/LArExample/RandomFigureOfMeritTool.cc
```diff
@@ -0,0 +1,66 @@
...
+        if (GetRandomFom() < m_maxFomToRecluster)
+            pfosToRecluster.push_back(pPfo);
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Changed

larpandoracontent/LArReclustering/LArExample/RandomFigureOfMeritTool.cc
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode RandomFigureOfMeritTool::CalcClusteringFom([[maybe_unused]] const ClusterList clusters, float &fom)
```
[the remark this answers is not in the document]
AlexWilkinsonnn: This is the first time I have ever had the chance to use a cpp attribute lol :mage:

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    m_pFomAlgTool {nullptr},
+    m_clusterUListName {"ClustersU"},
+    m_clusterVListName {"ClustersV"},
+    m_clusterWListName {"ClustersW"}
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+ThreeDMultiReclusteringAlgorithm::~ThreeDMultiReclusteringAlgorithm()
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Yep, changed

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+{
+    const PfoList *pPfos {nullptr};
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_INITIALIZED, !=,
+                                    PandoraContentApi::GetList(*this, m_pfoListName, pPfos));
+    if (!pPfos || pPfos->empty())
+        return STATUS_CODE_SUCCESS;
+
+    // Ask the FOM alg tool which pfos to recluster
+    PfoList pfosToRecluster;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, m_pFomAlgTool->GetPfosToRecluster(pPfos, pfosToRecluster));
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Changed, thanks

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
AlexWilkinsonnn: To throw or to return an status code is something I don't really have a good grasp on. The `RemovePfo` SDK call returns a `STATUS_CODE_FAILURE` so I am just having this function pass the status code along if that happens. Is there a compelling reason to have this as a throw? Execution gets stopped either way I think/hope

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    if (pfosToRecluster.empty())
+        return STATUS_CODE_SUCCESS;
+
+    // Remove clusters from the pfos, keeping track of their original pfo
+    std::map<HitType, ClusterList> viewToFreeClusters = { {TPC_3D, {}}, {TPC_VIEW_U, {}}, {TPC_VIEW_V, {}}, {TPC_VIEW_W, {}} };
+    std::map<const Pfo *const, ClusterList> pfoToFreeClusters;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, FreeClustersFromPfos(pfosToRecluster, viewToFreeClusters, pfoToFreeClusters));
+    if (viewToFreeClusters.at(TPC_3D).empty())
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Good point! As in the case of the new clustering being worse than the original, I changed it to reattach the clusters to their pfos and _then_ return. We just leave the pfos as they are if there is no 3D hits we can work with.

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+        return STATUS_CODE_SUCCESS;
+
+    // Run reclustering algs over the free 3D clusters to see if there is a superior 3D clustering
+    // NOTE Temporarily replacing the current list doesn't seem to work
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::ReplaceCurrentList<Cluster>(*this, m_cluster3DListName));
+    std::string originalClusterListName;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=,
+                             PandoraContentApi::InitializeReclustering(*this, TrackList(), viewToFreeClusters.at(TPC_3D), originalClusterListName));
+
+    float bestReclusterFom;
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Changed to initialise to lowest as small FOM is worse clustering

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    for (const std::string &clusteringAlg : m_clusteringAlgs)
+    {
+        std::string reclusterListName;
+        const ClusterList *pReclusterList {nullptr};
+        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=,
+                                 PandoraContentApi::RunClusteringAlgorithm(*this, clusteringAlg, pReclusterList, reclusterListName));
+        if (!pReclusterList || pReclusterList->empty())
+            continue;
+
+        float reclusterFom;
```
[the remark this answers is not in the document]
AlexWilkinsonnn: changed

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
AlexWilkinsonnn: I think the best thing to do is leave it to the figure of merit tool `GetPfosToRecluster` to tell us to recluster only pfos with 3D hits if that is desired. There may be some weird use case where you have a pfo with pretty bad 2D clusters and no 3D cluster which you want to throw into the reclustering with other pfos that do have 3D hits so that its 2D clusters can be reclustered according to the new 3D clustering... If that makes sense :grin:

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    {
+        ClusterList thisPfoClusters;
+        for (auto &[view, freeClusters] : viewToFreeClusters)
+        {
+            ClusterList clusters;
+            LArPfoHelper::GetClusters(pPfo, view, clusters);
+            for (const Cluster *const pCluster : clusters)
+            {
+                freeClusters.push_back(pCluster);
+                thisPfoClusters.push_back(pCluster);
```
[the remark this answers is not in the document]
AlexWilkinsonnn: changed

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode ThreeDMultiReclusteringAlgorithm::AddClustersToPfos(std::map<const Pfo *const, ClusterList> &pfoToClusters)
+{
+    for (const auto &[pPfo, clusters] : pfoToClusters)
+        for (const Cluster *const pCluster : clusters)
+            PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::AddToPfo(*this, pPfo, pCluster));
```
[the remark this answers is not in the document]
AlexWilkinsonnn: No this method only gets used when reattaching clusters to _original_ pfos. So this is only called when you are giving up on reclustering. I changed the name to `RelinkClustersToOriginalPfos` and changed the docstring to reflect this better. If this were to be used to only reattach the clusters to _some_ of the original pfos then yes we would need to update the free clusters map.

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+    std::map<HitType, ClusterList> viewToNewClusters2D = { {TPC_VIEW_U, {}}, {TPC_VIEW_V, {}}, {TPC_VIEW_W, {}} };
+    std::vector<ClusterList> newPfoClusters;
+    for (const Cluster *const pCluster3D : clusters)
+    {
+        ClusterList newClusters;
+        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, Build2DClustersFrom3D(pCluster3D, viewToFreeCaloHits2D, newClusters));
+        for (const Cluster *const pCluster : newClusters)
+            viewToNewClusters2D.at(LArClusterHelper::GetClusterHitType(pCluster)).push_back(pCluster);
+        newClusters.push_back(pCluster3D);
+        newPfoClusters.push_back(newClusters);
```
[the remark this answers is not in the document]
AlexWilkinsonnn: changed

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
AlexWilkinsonnn: Yep agree, added

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,360 @@
...
+        "ClusterVListName", m_clusterVListName));
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle,
+        "ClusterWListName", m_clusterWListName));
+
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ProcessAlgorithmList(*this, xmlHandle,
+        "ClusteringAlgorithms", m_clusteringAlgs));
+    AlgorithmTool *pAlgTool {nullptr};
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ProcessAlgorithmTool(*this, xmlHandle,
+        "FomAlgorithmTool", pAlgTool));
+    m_pFomAlgTool = dynamic_cast<ThreeDReclusteringFigureOfMeritBaseTool *>(pAlgTool);
```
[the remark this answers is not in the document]
AlexWilkinsonnn: yep thanks, added a check

AlexWilkinsonnn:
Thanks everyone for the quick review! The majority were easily resolved and I left some replies on others :grin:

AlexWilkinsonnn:
@imawby I expect it would be run at least before any hierarchy building, but maybe you would still want vertices + track/shower ID to happen beforehand so that the usual algs can use that extra information to attempt the initial 3D clustering.

The way I would envision it being used is to have some `LArListMerging ` and `LArListDeletion` blocks in the xml before this alg to get all the 3D clusters in one list and all the pfos in one list, then run the reclustering, then run any vertex building, track/shower ID and separation into track and shower lists, and hierarchy building afterwards.

Does that make sense?

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
AndyChappell: Yes, in principle these things behave in the same way (the main difference would be in behaviour of a helper function, where the throw variant doesn't allow a calling function to quiertly/unintentionally ignore a status code), the main reason for the two approaches relates to whether or not the function can/does return a StatusCode. If you don't have that option, the only way to bail the function and flag a problem is with an exception, which is where the THROW_RESULT_IF variant comes in. I think the RETURN variant is fine here, because it isn't ignored by the calling Run function.

AndyChappell (approved):

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
AlexWilkinsonnn: changed, thanks!

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,355 @@
...
+#include "larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+ThreeDMultiReclusteringAlgorithm::ThreeDMultiReclusteringAlgorithm() :
+    m_clusterListNames {{TPC_3D, {}}, {TPC_VIEW_U, "ClustersU"}, {TPC_VIEW_V, "ClustersV"}, {TPC_VIEW_W, "ClustersW"}},
+    m_pFomAlgTool {nullptr}
```
absolution1, line 26: A few data members are missing from the initialisation list

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,355 @@
...
+        return STATUS_CODE_SUCCESS;
+
+    // Ask the FOM alg tool which pfos to recluster
+    PfoList pfosToRecluster;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, m_pFomAlgTool->GetPfosToRecluster(pPfos, pfosToRecluster));
+    if (pfosToRecluster.empty())
+        return STATUS_CODE_SUCCESS;
+
+    // Remove clusters from the pfos, keeping track of their original pfo
+    std::map<HitType, ClusterList> viewToFreeClusters = { {TPC_3D, {}}, {TPC_VIEW_U, {}}, {TPC_VIEW_V, {}}, {TPC_VIEW_W, {}} };
```
absolution1, line 47: very minor (can be ignored) but this should probably be functionised as this default initialisation happens in a couple of places (e.g. the initialisation list of the class)

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,355 @@
...
+    std::map<const Pfo *const, ClusterList> pfoToFreeClusters;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, FreeClustersFromPfos(pfosToRecluster, viewToFreeClusters, pfoToFreeClusters));
+    if (viewToFreeClusters.at(TPC_3D).empty()) // No 3D clusters to work with, rebuild original pfos as if nothing happened...
+        return RelinkClustersToOriginalPfos(pfoToFreeClusters);
+
```
absolution1, line 52: Does this ever fire?
This seems like a tear down and immediate re-build.  Is that correct?

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,355 @@
...
+{
+    const PfoList *pPfos {nullptr};
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_INITIALIZED, !=,
+                                    PandoraContentApi::GetList(*this, m_pfoListName, pPfos));
+    if (!pPfos || pPfos->empty())
+        return STATUS_CODE_SUCCESS;
+
+    // Ask the FOM alg tool which pfos to recluster
+    PfoList pfosToRecluster;
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, m_pFomAlgTool->GetPfosToRecluster(pPfos, pfosToRecluster));
```
absolution1, line 42: Do we only ever want to calculate a single FOM?

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,355 @@
...
+        }
+    }
+
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::TemporarilyReplaceCurrentList<Cluster>(*this, bestReclusterListName));
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::EndReclustering(*this, bestReclusterListName));
+
+    if (bestReclusterListName == originalClusterListName) // No 3D clusters to work with, rebuild original pfos as if nothing happened...
+    {
+        return RelinkClustersToOriginalPfos(pfoToFreeClusters);
+    }
```
absolution1, line 90: Does this ever fire?

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
```diff
@@ -0,0 +1,355 @@
...
+            const ClusterList *pOldClusters2D {nullptr};
+            PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::GetList(*this, clusterListName, pOldClusters2D));
+            // This means no hits in this view made any 3D hits, this happens very rarely for events with few hits.
+            // In this case, just leave these hits unclustered. Could also consider putting them in their own isolated hit-only cluster?
+            if (pOldClusters2D->empty())
+                continue;
+            PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, MopUpCaloHits(caloHits2D, *pOldClusters2D, true));
+            continue;
+        }
+        PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, MopUpCaloHits(caloHits2D, viewToNewClusters2D.at(view), true));
```
absolution1, line 207: I'd like to understand the motivation for the 'framework' being in charge of mopping up hit that were failed to be reclustered.
To me this feels like a physics/end user decision and not something the framework should decide.

absolution1 (commented):
Thanks Alex.
I've added a few comments and questions.  Most are trivial.
My main quandary is the motvation for the framework being completely in charge of mopping up hits that did not get reclustered, as this feels more like an end-user/physics decision rather than machinery.

larpandoracontent/LArReclustering/LArExample/RandomFigureOfMeritTool.cc
```diff
@@ -0,0 +1,66 @@
...
+{
+    fom = GetRandomFom();
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Is this the style when only for when the definition of the member function is not in the current file or just for all member function calls?

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
absolution1: I was more thinking that we might want to run multiple tools, each of which calculates a single FOM.  Is that not ever going to be the case?

larpandoracontent/LArReclustering/ThreeDMultiReclusteringAlgorithm.cc
absolution1: Personally I don't think the framework should do the mopping up and it should be left to a future alg.
But, I think this is a bit of a technical detail that shouldn't hold up the PR.

I think the halfway house solution is to add a flag to the host class that enables/disables the mop up as you have it now.  Have the XML read in the value of the flag and set it to true for now.

PR #241, opened 2025-03-05
Rand index helper function + 2d clustering validation alg

AlexWilkinsonnn:
Thought it might be useful to get the adjusted rand index (a measure of the similarity of two clusterings) calculation into a helper function. The helper function calculates the index from a contingency table (a map of map which encodes all non-zero intersections between the two clusterings) which the user should make. I included additional helpers to make the contingency table for the case of `Cluster` and `MCParticle` clusterings which I think would be the most common use case.

I also included my 2D clustering monitoring algorithm that computes the rand index (and other aggregate metrics) for the 2D clustering in each view of an event.

### DIFF
--- a/larpandoracontent/LArHelpers/LArMonitoringHelper.h
+++ b/larpandoracontent/LArHelpers/LArMonitoringHelper.h
@@ -80,8 +80,9 @@ public:
 
     /**
      *  @brief  Calculate the adjusted Rand Index for a given contingency table that summarises two clusterings A and B.
-     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical)
-     *          and is 0 when the two clusterings are as similar as two random clusterings with the same number of clusters.
+     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical),
+     *          is 0 when the two clusterings are as similar as two random clusterings with the same number of clusters,
+     *          and has no lower bound. An index < 0 means very discordant clusterings and is rare in most cases.
      *          Reference - https://link.springer.com/article/10.1007/BF01908075
      *
      *  @param[in] cTable Contingency table as a map of the object that defines the clustering A (e.g. pandora::Cluster)
@@ -93,8 +94,9 @@ public:
 
     /**
      *  @brief  Calculate the adjusted Rand Index for the clustering defined by MCPartices and by CaloHits.
-     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical)
-     *          and is 0 when the two clusterings are as similar as two random clusterings with the same number of clusters.
+     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical),
+     *          is 0 when the two clusterings are as similar as two random clusterings with the same number of clusters,
+     *          and has no lower bound. An index < 0 means very discordant clusterings and is rare in most cases.
      *          Reference - https://link.springer.com/article/10.1007/BF01908075
      *
      *  @param[in] caloHits List of hits
--- a/larpandoracontent/LArHelpers/LArMonitoringHelper.h
+++ b/larpandoracontent/LArHelpers/LArMonitoringHelper.h
@@ -75,6 +75,46 @@ public:
      */
     static void PrintMatchingTable(const pandora::PfoVector &orderedPfoVector, const pandora::MCParticleVector &orderedMCParticleVector,
         const LArMCParticleHelper::MCParticleToPfoHitSharingMap &mcParticleToPfoHitSharingMap, const unsigned int nMatches);
+
+    // -- Rand index helpers
+
+    /**
+     *  @brief  Calculate the adjusted Rand Index for a given contingency table that summarises two clusterings A and B.
+     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical)
+     *          and is 0 when the two clusterings are as similar as two random clusterings with the same number of clusters.
+     *          Reference - https://link.springer.com/article/10.1007/BF01908075
+     *
+     *  @param[in] cTable Contingency table as a map of the object that defines the clustering A (e.g. pandora::Cluster)
+     *                    to a map of the object that defines the clustering B (e.g. pandora::MCParticle)
+     *                    to the value of the table at this entry (which is the intersection of the two clusters)
+     */
+    template<typename Ti, typename Tj>
+    static float CalcRandIndex(const std::map<const Ti, std::map<const Tj, int>> &cTable);
+
+    /**
+     *  @brief  Calculate the adjusted Rand Index for the clustering defined by MCPartices and by CaloHits.
+     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical)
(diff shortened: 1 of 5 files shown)


larpandoracontent/LArHelpers/LArMonitoringHelper.h
```diff
@@ -75,6 +75,46 @@ class LArMonitoringHelper
      */
     static void PrintMatchingTable(const pandora::PfoVector &orderedPfoVector, const pandora::MCParticleVector &orderedMCParticleVector,
         const LArMCParticleHelper::MCParticleToPfoHitSharingMap &mcParticleToPfoHitSharingMap, const unsigned int nMatches);
+
+    // -- Rand index helpers
+
+    /**
+     *  @brief  Calculate the adjusted Rand Index for a given contingency table that summarises two clusterings A and B.
+     *          Adjusted Rand Index is a measure of the similarity of two clusterings that is bounded above by 1 (identical)
+     *          and is 0 when the two clusterings are as similar as two random clusterings with the same number of clusters.
```
AndyChappell, line 84: Probably worth noting the lack of a lower bound (unless we enforce one), for the "worse than random" case.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.h
```diff
@@ -0,0 +1,125 @@
...
+        const pandora::MCParticle *m_pMainMC;
+        const pandora::Cluster *m_pCluster;
+    };
+
+    /**
+    *  @brief  Default constructor
+    */
+    EventClusterValidationAlgorithm();
+
+    virtual ~EventClusterValidationAlgorithm();
```
AndyChappell, line 55: Base Process destructor is already virtual, so don't want to redeclare as virtual here.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -0,0 +1,337 @@
...
+
+#include <numeric>
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+EventClusterValidationAlgorithm::CaloHitParents::CaloHitParents() :
+    m_pMainMC {nullptr},
```
AndyChappell, line 24: Don't need to change this, but just a note that clang format will remove the space between the name and the opening brace here.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -0,0 +1,337 @@
...
+        mcSumHits.at(pMainMC)++;
+    }
+
+    for (auto it = hitParents.begin(), itNext = it; it != hitParents.end(); it = itNext)
+    {
+      itNext++;
+      if (mcSumHits.at(it->second.m_pMainMC) < m_minMCHitsPerView)
+        hitParents.erase(it);
+    }
+}
```
AndyChappell, line 187: I thnk my preference here would be for

```
    for (auto it = hitParents.begin(), it != hitParents.end();)
    {
        if (mcSumHits.at(it->second.m_pMainMC) < m_minMCHitsPerView)
            it = hitParents.erase(it);
        else
            ++it;
    }
```

Alternatively, I think this conditional probably allows you to do this with the std::erase function and a lambda function and avoid the for loop altogether.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -0,0 +1,337 @@
...
+    std::map<const CaloHit *const, CaloHitParents> hitParentsValid;
+    for (const auto &[pCaloHit, parents] : hitParents)
+    {
+        const MCParticle *const pMainMC {parents.m_pMainMC};
+        if (valType != ValidationType::ALL)
+        {
+            const int pdg {std::abs(pMainMC->GetParticleId())};
+            if (
+                (valType == ValidationType::SHOWER && (pdg != 22 && pdg != 11)) ||
+                (valType == ValidationType::TRACK && (pdg == 22 || pdg == 11))
```
AndyChappell, line 204: As a note, among the Pandora SDK definitions (Pandora/PdgTable.h) are the defintions PHOTON and E_MINUS

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -0,0 +1,337 @@
...
+        for (const auto &[pMC, nHits] : mainMCParticleHits)
+        {
+            if (nHits > maxHits)
+            {
+                pMainMC = pMC;
+                maxHits = nHits;
+            }
+        }
+        metrics.m_purities.push_back(maxHits / totalHits);
+        metrics.m_nRecoHits.push_back(totalHits);
```
AndyChappell, line 266: Typically prefer emplace_back over push_back.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -0,0 +1,337 @@
...
+{
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), m_treeName, branchPrefix + "n_hits", metrics.m_nHits ));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), m_treeName, branchPrefix + "n_clusters", metrics.m_nClusters));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), m_treeName, branchPrefix + "n_mainMCs", metrics.m_nMainMCs));
+    double meanPurity = -1.0, meanCompleteness = -1.0, meanNRecoHits = -1.0;
+    if (metrics.m_nClusters > 0)
+    {
+        meanPurity = std::accumulate(metrics.m_purities.begin(), metrics.m_purities.end(), 0.0) / metrics.m_purities.size();
+        meanCompleteness = std::accumulate(metrics.m_completenesses.begin(), metrics.m_completenesses.end(), 0.0) / metrics.m_completenesses.size();
+        meanNRecoHits = std::accumulate(metrics.m_nRecoHits.begin(), metrics.m_nRecoHits.end(), 0.0) / metrics.m_nRecoHits.size();
```
AndyChappell, line 310: I couldn't quite convince myself that the denomenator here can never be zero.  I think it's probably ok, but can the code get here if hitParents ends up being empty?

AndyChappell (changes requested):
Thanks Alex, this looks good. A couple of minor requests and a couiple of queries, otherwise this should be good to go out in the next release.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: Yeah this method is cleaner, hadnt realised `erase` returns the next element. Could use `erase_if` for maps then could do this with a lambda but that is C++20 so maybe safer to leave it

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: I didnt have any problems with this on a 50k sample but I think your suspicion is correct as if truth matching was completely missing hitParents would be empty. I just checked everything to be safe

AndyChappell (approved):

PR #243, opened 2025-03-31
2d clustering validation bug

AlexWilkinsonnn:
Pretty bad bug I made in calculation of a hit's main MC particle :disappointed:  . In the majority of cases where only one MC particle contributes to a hit, this makes no difference, but in the other cases the weight was not being taken into account and the last MC particle in the weight map was being taken as the main MC particle.


AndyChappell (approved):
Thanks, Alex. No worries. This week's release is already flagged for merging, so I'll get this one into a release branch for next week.

PR #246, opened 2025-05-12
Use Collection plane in ICARUS for track/shower discrimination

rtriozzi:
This PR (based off `v04_07_00`) introduces some changes to the MVA track/shower discrimination modules and charge-based tools (`ConeChargeFeatureTool` and `ThreeDChargeFeatureTool`) specifically for ICARUS. The change consists in selecting the views that are mapped to the Collection wire plane in ICARUS, instead of defaulting to the W view (which is mapped to the Induction-1 wire plane in ICARUS). The choice of views for Collection depends on whether the reconstructed PFP crosses the cathode (in which case a combination of U and V is needed) or not (in which case we just need to pick either U or V).

As a starting point, the changes are made directly to the standard track/shower MVA classes and are optional, enabled by a `UseICARUSCollectionPlane` XML parameter set to `false` by default.

For context, attached is a scheme of the wire planes for one of the two ICARUS modules:
ICARUSWirePlanesScheme

Tagging @absolution1, @brucehoward-physics, and @cerati: thanks!

### DIFF
--- a/larpandoracontent/LArContent.cc
+++ b/larpandoracontent/LArContent.cc
@@ -410,10 +410,12 @@
     d("LArTwoDVertexDistanceFeatureTool",       TwoDVertexDistanceFeatureTool)                                                  \
     d("LArThreeDVertexDistanceFeatureTool",     ThreeDVertexDistanceFeatureTool)                                                \
     d("LArThreeDChargeFeatureTool",             ThreeDChargeFeatureTool)                                                        \
+    d("LArThreeDChargeFeatureTool_ICARUS",      ThreeDChargeFeatureTool_ICARUS)                                                 \
     d("LArThreeDPCAFeatureTool",                ThreeDPCAFeatureTool)                                                           \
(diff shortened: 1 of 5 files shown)


rtriozzi:
Hi @AndyChappell, thank you very much for the review! 

As discussed, I’ve moved all the ICARUS-specific changes into new, parallel tools following the suggested naming convention (i.e., introduced `ConeChargeFeatureTool_ICARUS`, `ThreeDChargeFeatureTool_ICARUS`, and re-named `LocatePointInCryostat_ICARUS`). 
The original code is now untouched, and ICARUS-specific changes can be enabled simply by selecting the corresponding tools in the XML (I've also kept the XML flags in place for the ICARUS tools).

Let me know if anything else needs adjustments. Thanks again!

AndyChappell (approved):
Hi Riccardo, the latest updates look good to me, thanks. I'll look to get this into next week's LArSoft release cycle.

PR #247, opened 2025-05-31
Add LANGUAGES CXX in the project call in CMakeLists.txt

jmcarcell:
This disables a check for a C compiler since the default is both C and C++


AndyChappell:
Rejected. We prefer to avoid changes to the build configuration that are not strictly necessary.

PR #248, opened 2025-06-23
Update to validation alg

AlexWilkinsonnn:
This is a consolidation of updates I made to this validation alg from various branches (hence the single big commit, sorry...)

It now does this for the 2D clusters in each view:
- Purity and completeness averaged over clusters or hits
- Adjusted rand index
- Split the above two into all/track/shower only hits/clusters
- Match clusters to MCParticles and write out the values needed to calculate purity+completeness for these
- In the target clusters, options to fold showers to leading particle + prevent overlapping delta-rays
- Handle hits with no cluster
- Visualisation of target clusters
- For bayesian optimisation work there are also the options to:
  - Cheat the merging of shower clusters when filling the contingency table for the rand index calculation
  - Option to only compute the rand index for all hits to reduce processing time

### DIFF
--- a/larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
+++ b/larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
@@ -698,8 +698,10 @@ void EventClusterValidationAlgorithm::GetMatchedParticleMetrics(
     else
     {
         std::map<const Cluster *const, MCParticleList> clusterOrderedMCs;
-        for (const auto &[pCluster, mcNHits] : clusterMCNHits)
+        for (auto it = clusterMCNHits.begin(); it != clusterMCNHits.end(); it++)
         {
+            const Cluster *const pCluster = it->first;
+            const std::map<const MCParticle *const, int> &mcNHits = it->second;
             MCParticleList orderedMCs;
             for (const auto &[pMC, nHits] : mcNHits)
             {


larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -129,20 +198,49 @@ StatusCode EventClusterValidationAlgorithm::Run()
...
+                }
+                else
+                {
+                    pFoldedMC = this->FoldMCTo(pMC);
+                    mcFoldTo.insert({pMC, pFoldedMC});
+                }
+                foldedWeightMap[pFoldedMC] += weight;
+            }
+            weightMap = foldedWeightMap;
+        }
```
AndyChappell, line 226: Could you compute an MC to MC map at the start of the alg and just use this to look up the approriate MC? i.e. an unfolded map would just have each MC point to itself, while the folded shower map would have MC point to itself for non-shower MC, and then point to the leading MC for the shower MC? That way you separate the folding logic from the hit to MC logic.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -151,181 +249,681 @@ void EventClusterValidationAlgorithm::GetHitParents(
         }
     }
 
+    if (m_handleDeltaRays)
+    {
+        for (auto &[pCaloHit, parents] : hitParents)
+        {
+            parents.m_pMainMC = this->FoldPotentialDeltaRayTo(pCaloHit, parents.m_pMainMC);
```
AndyChappell, line 256: This might be more complex given the hit level decision making that I think this involves, rather than a pure MC consideration, but might this also fit into a precomputed folding?

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -151,181 +249,681 @@ void EventClusterValidationAlgorithm::GetHitParents(
...
         const CaloHitList &isolatedHits{pCluster->GetIsolatedCaloHitList()};
         CaloHitList clusterCaloHits;
         pCluster->GetOrderedCaloHitList().FillCaloHitList(clusterCaloHits);
         clusterCaloHits.insert(clusterCaloHits.end(), isolatedHits.begin(), isolatedHits.end());
+
         for (const CaloHit *const pCaloHit : clusterCaloHits)
         {
-            // Ignoring the calo hit if truth matching is missing
             if (hitParents.find(pCaloHit) == hitParents.end())
+            {
```
AndyChappell, line 270: The now deleted comment notes that hits with MC are ignored (which makes sense), but, if I understand correctly, hitParents can now have hits removed due to other criteria, is it still correct to skip over these hits, or do they need to be accounted for in the metrics (e.g. if I've understood correctly, a hit with a true PDG code that doesn't fit the category you are looking at would be omitted from the metrics, when in practice I think you'd want this to turn up in the error accounting. Have I understood this correctly?)

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -151,181 +249,681 @@ void EventClusterValidationAlgorithm::GetHitParents(
...
+                hasAncestorElectron = true;
+            }
+            pCurrentMC = pParentMC;
+            continue;
+        }
+        pLeadingMC = pCurrentMC;
+        break;
+    }
+
+    // Don't fold "showers" that consist of only compton scatters
```
AndyChappell, line 307: LArMCParticles do store process information, so in principle you could explicitly look for "only compton scatters" directly. Would this fit your aim here?

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -151,181 +249,681 @@ void EventClusterValidationAlgorithm::GetHitParents(
...
+    }
+    else
+    {
+        return pMC;
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool EventClusterValidationAlgorithm::CausesShower(const MCParticle *const pMC, int nDescendentElectrons) const
```
AndyChappell, line 325: Does this behave logically if showers are folded at sim-level? Do we care?

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
```diff
@@ -151,181 +249,681 @@ void EventClusterValidationAlgorithm::GetHitParents(
...
+    // Did not come from a track-like particle -> not a delta ray -> do nothing
+    const MCParticle *const pParentMC{*(pMC->GetParentList().begin())};
+    const int parentPdg{std::abs(pParentMC->GetParticleId())};
+    if (parentPdg == PHOTON || parentPdg == E_MINUS || PdgTable::GetParticleCharge(parentPdg) == 0)
+    {
+        return pMC;
+    }
+
+    // Delta ray that does not start a shower and is short -> fold into parent particle
+    if (!this->CausesShower(pMC, 0) && (pMC->GetVertex() - pMC->GetEndpoint()).GetMagnitudeSquared() < m_deltaRayLengthThresholdSquared)
```
AndyChappell, line 374: As per earlier comment on wire pitch, at the moment I think this will fold back a lot of delta ray hits? ~2.5 cm before something is distinct?

AndyChappell (commented):
Thanks Alex. A couple of minor and a couple of more substantive comments/queries.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: Yes I see what you mean, I considered a similar implementation but I think I prefer the way it is currently.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: Yes there are more configurable reasons to drop hits from hitParents map now (hit is in a null reco cluster, minimum required number of MC hits in a view). These hits are only skipped over if configured to do so which I think is ok.  The only ones that are completely ignored is when the hit has no MC weight.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: Good to know, I didnt realise we have access to process information. So I think I could replace this ancestor + child electron counting with explicitly checking the end process type, but I validated this behaves sensibly with event displays so dont think its worth.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: It doesnt matter what this does for already folded MC as the two options for folding, pMC and pLeadingMC, will just be the same particle. Ideally you wouldnt configure folding if you know your sample is already folded but I think it behaves normally (does nothing) even if you do.

larpandoracontent/LArMonitoring/EventClusterValidationAlgorithm.cc
AlexWilkinsonnn: All of these are in mm, so the threshold is 0.25cm which I think is small enough to ignore

AlexWilkinsonnn:
I added a flag for the `GetMatchedParticleMetrics` method that does a slightly different mc particle -> cluster matching

PR #256, opened 2025-10-10
Introduce infrastructure to take in external hit-level "track" and "shower" predictions

rtriozzi:
This PR (based off of `larpandoracontent v04_15_01`) adds the interface to take in hit-level track (`m_pTrack`) and shower (`m_pShower`) probabilities from external inputs. This PR is linked to [larpandora PR #35](https://github.com/PandoraPFA/larpandora/pull/35) (where a bit more detail is provided).

Tagging @cerati, @AndyChappell, and @absolution1, thank you very much!

### DIFF
--- a/larpandoracontent/LArObjects/LArCaloHit.h
+++ b/larpandoracontent/LArObjects/LArCaloHit.h
@@ -278,6 +278,8 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
     // ATTN: To receive this call-back must have already set file reader mc particle factory to this factory
     unsigned int larTPCVolumeId(std::numeric_limits<unsigned int>::max());
     unsigned int daughterVolumeId(0);
+    float pTrack(-1.);
+    float pShower(-1.);
 
     if (pandora::BINARY == fileReader.GetFileType())
     {
@@ -285,6 +287,11 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(larTPCVolumeId));
         if (m_version > 1)
             PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(daughterVolumeId));
+        if (m_version > 2) {
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(pTrack));
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(pShower));  
+        }
+        
     }
     else if (pandora::XML == fileReader.GetFileType())
     {
@@ -292,6 +299,10 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("LArTPCVolumeId", larTPCVolumeId));
         if (m_version > 1)
             PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("DaughterVolumeId", daughterVolumeId));
+        if (m_version > 2) {
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("PTrack", pTrack));
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("PShower", pShower));
+        }
     }
     else
     {
@@ -301,6 +312,8 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
     LArCaloHitParameters &larCaloHitParameters(dynamic_cast<LArCaloHitParameters &>(parameters));
     larCaloHitParameters.m_larTPCVolumeId = larTPCVolumeId;
     larCaloHitParameters.m_daughterVolumeId = daughterVolumeId;
+    larCaloHitParameters.m_pTrack = pTrack;
+    larCaloHitParameters.m_pShower = pShower;
 
     return pandora::STATUS_CODE_SUCCESS;
 }
--- a/larpandoracontent/LArObjects/LArCaloHit.h
+++ b/larpandoracontent/LArObjects/LArCaloHit.h
@@ -30,8 +30,8 @@ public:
     pandora::InputUInt m_larTPCVolumeId;   ///< The lar tpc volume id
     pandora::InputUInt m_daughterVolumeId; ///< The daughter volume id
 
-    pandora::InputFloat m_pTrack;          ///< Probability the hit is track-like
-    pandora::InputFloat m_pShower;         ///< Probability the hit is shower-like
+    std::vector<float> m_hitScores;             ///< Hit scores
+    std::vector<std::string> m_hitScoreLabels;  ///< Labels for the hit scores
 };
 
 //------------------------------------------------------------------------------------------------------------------------------------------
@@ -71,6 +71,20 @@ public:
     void FillParameters(LArCaloHitParameters &parameters) const;
 
     /**
+     *  @brief  Get scores for each hit
+     *
+     *  @return a vector of scores for each category
+     */
+    std::vector<float> GetHitScores() const;
+
+    /**
+     *  @brief  Get scores for each hit
+     *
+     *  @return a vector of scores for each category
+     */
+    std::vector<std::string> GetHitScoreLabels() const;
+
+    /**
      *  @brief  Get the probability that the hit is track-like
      *
      *  @return the probability the hit is track-like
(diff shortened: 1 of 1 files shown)


AndyChappell (changes requested):
Hi @rtriozzi. Thanks for the update. The core of the update looks fine to me (one open question as to what the "best" initialisation value is), but if external inputs are going to be made available to LArContent algorithms we'd want to ensure those values can be persisted in PNDRs so that we can replicate the behaviour when running in standalone versions of Pandora. I've added links to examples showing how to do this (this would be accompanied by a suitable increase in the LArCaloHitVersion to 3 in XML (similar to [this](https://github.com/PandoraPFA/LArReco/blob/625776fb4655619e7b8f1bc1da90c9298a553414/settings/PandoraSettings_Master_DUNEFD_VD.xml#L27)). Thanks!

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -158,7 +161,9 @@ class LArCaloHitFactory : public pandora::ObjectFactory<object_creation::CaloHit
 inline LArCaloHit::LArCaloHit(const LArCaloHitParameters &parameters) :
     object_creation::CaloHit::Object(parameters),
     m_larTPCVolumeId(parameters.m_larTPCVolumeId.Get()),
-    m_daughterVolumeId(parameters.m_daughterVolumeId.IsInitialized() ? parameters.m_daughterVolumeId.Get() : 0)
+    m_daughterVolumeId(parameters.m_daughterVolumeId.IsInitialized() ? parameters.m_daughterVolumeId.Get() : 0),
+    m_pTrack(parameters.m_pTrack.IsInitialized() ? parameters.m_pTrack.Get() : -1.),
+    m_pShower(parameters.m_pShower.IsInitialized() ? parameters.m_pShower.Get() : -1.)
```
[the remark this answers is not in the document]
rtriozzi: thanks Andy! I think I would prefer something like -1 (or another "invalid" float value) rather than zero (which is technically among the possible values for `m_pTrack`, e.g.). Let me know if you have a preference!

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -29,6 +29,9 @@ class LArCaloHitParameters : public object_creation::CaloHit::Parameters
 public:
     pandora::InputUInt m_larTPCVolumeId;   ///< The lar tpc volume id
     pandora::InputUInt m_daughterVolumeId; ///< The daughter volume id
+
+    std::vector<float> m_hitScores;             ///< Hit scores
+    std::vector<std::string> m_hitScoreLabels;  ///< Labels for the hit scores
```
AndyChappell: Pandora has typedefs pandora::FloatVector and pandora::StringVector.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -67,6 +70,20 @@ class LArCaloHit : public object_creation::CaloHit::Object
...
+     *  @return a vector of scores for each category
+     */
+    std::vector<float> GetHitScores() const;
+
+    /**
+     *  @brief  Get scores for each hit
+     *
+     *  @return a vector of scores for each category
+     */
+    std::vector<std::string> GetHitScoreLabels() const;
```
AndyChappell: pandora::StringVector

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -271,20 +308,36 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
     // ATTN: To receive this call-back must have already set file reader mc particle factory to this factory
     unsigned int larTPCVolumeId(std::numeric_limits<unsigned int>::max());
     unsigned int daughterVolumeId(0);
+    unsigned int nLabels(0);
+    std::vector<float> hitScores;
+    std::vector<std::string> hitScoreLabels;
```
AndyChappell: typedefs

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -271,20 +308,36 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
...
 
     if (pandora::BINARY == fileReader.GetFileType())
     {
         pandora::BinaryFileReader &binaryFileReader(dynamic_cast<pandora::BinaryFileReader &>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(larTPCVolumeId));
         if (m_version > 1)
             PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(daughterVolumeId));
+        if (m_version > 2) {
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, binaryFileReader.ReadVariable(hitScores));
```
AndyChappell: This won't work. You'll need to do something similar to what you've done with hitScoreLabels. Under the hood, ReadVariable checks the size of the type being passed in, but for a std::vector, this will not factor in the space allocated to the elements, only the address, current size and capacity information. The safest option here is to have an integer representing the number of elements in the vector that's stored in the file, and then read in N floats and populate the vector that way.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -294,6 +347,8 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
     LArCaloHitParameters &larCaloHitParameters(dynamic_cast<LArCaloHitParameters &>(parameters));
     larCaloHitParameters.m_larTPCVolumeId = larTPCVolumeId;
     larCaloHitParameters.m_daughterVolumeId = daughterVolumeId;
+    larCaloHitParameters.m_hitScores = hitScores;
+    larCaloHitParameters.m_hitScoreLabels = hitScoreLabels;
```
AndyChappell: There's a need for the corresponding Write function to be updated to conform to what's expected in the Read function above.

AndyChappell (changes requested):
Hi @rtriozzi , thanks for the PR. Conceptually this all looks ok. I think the main issues here are around persisting PNDR files. As described in some specific comments, the reading of the scores vector won't behave in the way we would want for the binary file case, so you'd end up with an invalid vector - I think you'd have a vector with the correct size and capacity, but with the elements unpopulated. Additionally, the Write function doesn't appear to have been updated to match. If you can address those issues, I think this is looking reasonable. Thanks!

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -178,6 +199,20 @@ inline unsigned int LArCaloHit::GetDaughterVolumeId() const
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+inline std::vector<float> LArCaloHit::GetHitScores() const
```
AndyChappell: It looks like this could return

```
const FloatVector &
```

rtriozzi:
Hi @AndyChappell, thanks a lot for the review! 
Let me know if the new code fixes the handling of PNDR files.

AndyChappell (commented):
Pending testing, I think these updates looks good thanks @rtriozzi

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -271,20 +308,44 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
...
     }
     else if (pandora::XML == fileReader.GetFileType())
     {
         pandora::XmlFileReader &xmlFileReader(dynamic_cast<pandora::XmlFileReader &>(fileReader));
         PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("LArTPCVolumeId", larTPCVolumeId));
         if (m_version > 1)
             PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("DaughterVolumeId", daughterVolumeId));
+        if (m_version > 2) {
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("NHitLabels", nLabels));
+            PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("HitScores", hitScores));
```
AndyChappell: I think here we probably want a conditional that only attempts to read HitScores if nLabels > 0 - otherwise I think we don't want the tag in the file at all, analagous to the behaviour of the PNDR format.

larpandoracontent/LArObjects/LArCaloHit.h
```diff
@@ -271,20 +308,44 @@ inline pandora::StatusCode LArCaloHitFactory::Read(Parameters &parameters, pando
...
+            for (unsigned int i = 0; i < nLabels; ++i) {
+                std::string label;
+                PANDORA_RETURN_RESULT_IF(pandora::STATUS_CODE_SUCCESS, !=, xmlFileReader.ReadVariable("HitScoreLabels", label));
```
AndyChappell: Minor, but as this is reading one-by-one, we probably want the label to be HitScoreLabel.

AndyChappell (changes requested):
Hi Riccardo, I think a couple of updates are needed here for consistency of the file processing at read and write time. Otherwise I think this is nearly ready.

AndyChappell (approved):
Updates look good thanks Riccardo. Subject to testing, this looks ready to go.

PR #257, opened 2025-10-21
Feature/ingratta larpandoracontent v04 14 00 add zstitching algorithm

GianfrancoIngratta:
This PR implements the functions to stitch pfos in the Z direction, enabling rock muons tagging. The current version allows activating the stitching only if the related flag is set to true in the cosmic XML.


AndyChappell (commented):
Thanks for the PR Gianfranco. I've not had a chance to have a detailed look at this yet, but a quick glance through highlights a few things that might want some broader consideration in the interim.

On the more trivial side, the couts and the blocks of commented out code should be removed for a release (or for some prints you may want to wrap them in the standard pandora optional printout structures to ensure they are only printed when requesting verbose logging).

There's a broader discussion to be had, both within the ND group and with the wider Pandora development team about structure here. We have some competing considerations regarding what's ND, what's FD, and what's generic, and we need to figure out how to handle this to avoid creating maintenance problems moving forward.

Most immediately, I think there needs to be some consideration of how to isolate, or cleanly integrate what are currently ND-specific considerations. The current extension of the Stitching tool suggests this might be better as a completely independent tool, rather than adding this functionality to the existing tool, where it seems in an FD, or non-DUNE context, a lot of overhead would be added where the use case isn't present. In general we like to try to follow the open-closed principle - open for extension, closed for modification. That's not always possible of course, but anything that modifies existing code runs the risk of unintended consequences for other workflows or experiments.

Stylistically, we do have .clang-format to enforce Pandora style (though please be sure to isolate any code formatting to their own commits), as it's important that we maintain consistency of style, and also variable naming conventions (e.g variables are camelcase, starting with lower case, while functions do start with an uppercase first character).

I'll try to have a more detailed look over the next week or two, but just some provisional things to highlight. Thanks again.

GianfrancoIngratta:
Progress on this PR [here](https://docs.dunescience.org/cgi-bin/sso/DisplayMeeting?conferenceid=1662) (scroll down after CAF-related slides). This feature branch was validated using 2x2 simulation MiniRun6.4 with chaeted workflow. See presentation for details.

PR #271, opened 2026-02-20
VisualMonitoring - option to inspect hit TPC+Daughter vols

AlexWilkinsonnn:
I think this is useful. Especially as the Pandora geometry doesn't know about the Daughter volumes afaik, so the only way to inspect them is through the CaloHits. 

Screenshot from 2026-02-20 17-59-44

### DIFF
--- a/larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
+++ b/larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
@@ -214,7 +214,7 @@ void VisualMonitoringAlgorithm::VisualizeCaloHitList(const std::string &listName
     for (const CaloHit *const pCaloHit : caloHitList)
     {
         const LArCaloHit *const pLArCaloHit{dynamic_cast<const LArCaloHit *>(pCaloHit)};
-        const std::pair<unsigned int, unsigned int> volId = { pLArCaloHit->GetLArTPCVolumeId(), pLArCaloHit->GetDaughterVolumeId() };
+        const std::pair<unsigned int, unsigned int> volId{ pLArCaloHit->GetLArTPCVolumeId(), pLArCaloHit->GetDaughterVolumeId() };
         volIdToCaloHits[volId].push_back(pCaloHit);
     }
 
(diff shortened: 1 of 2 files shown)


larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
```diff
@@ -201,8 +203,31 @@ void VisualMonitoringAlgorithm::VisualizeCaloHitList(const std::string &listName
...
+
+    [[maybe_unused]] int colorIter{0};
+    for (const auto &[volId, caloHitListPartition] : volIdToCaloHits)
+    {
+        const std::string label{(listName.empty() ? "CurrentCaloHits" : listName) +
+            ", volId ("  + std::to_string(volId.first) + ", " + std::to_string(volId.second) + ")"};
+        PANDORA_MONITORING_API(VisualizeCaloHits(this->GetPandora(), &caloHitListPartition, label.c_str(),
+            static_cast<Color>(++colorIter) < AUTO          ? static_cast<Color>(colorIter)
+                : (colorIter % static_cast<int>(AUTO)) == 0 ? static_cast<Color>(++colorIter % static_cast<int>(AUTO))
+                                                            : static_cast<Color>(colorIter % static_cast<int>(AUTO))));
```
AndyChappell: So, if the incremented value of the colour is in the set prior to the AUTO value, just give it the colour. If that incremented value is AUTO, increment the colour again and use that (this bit confuses me), and if it's not auto, set it to whatever the cyclic colour is?

I feel like the setting of colorIter should probably be independent of the function call, I could imagine that being both easier to comprehend, and shorter - a few too many side effects for me on this one.

AndyChappell (changes requested):
Additional feature looks good - requesting a change in the modification of colorIter.

larpandoracontent/LArMonitoring/VisualMonitoringAlgorithm.cc
AlexWilkinsonnn: Yes, the extra condition == 0 is to skip the WHITE color when cycling back round. It is all stuffed in the PANDORA_MONITORING_API macro to stop the monitoring off build complaining about the Color enum not existing.

I refactored it to use  #ifdef MONITORING with some (hopefully) less opaque logic outside the PANDORA_MONITORING_MACRO.

AndyChappell (approved):

PR #274, opened 2026-03-27
Refactor of roll-up code & DL shower growing hit feature vector construction

AlexWilkinsonnn:
I put all the code related to roll-up of EM activity in a RollUp utility as there is growing duplication of the same methods. This also included moving a couple of functions to the MC particle helper class.

The refactored roll up uses a controller class `RollUp` that you initialise with the desired roll up policy class (classes that inherit from the policy interface `IRollUpPolicy`). So to use the roll up you would have something like this in your code
```
RollUpper rollUpper = RollUpper(std::make_unique<RollUpEMPolicy>());
MCParticle *pMC1 = rollUpper.RollUpCaloHit(pCaloHit);
MCParticle *pMC2 = rollUpper.RollUpMC(pMC);
```
`RollUp` contains a cache of previously computed mappings while the policy classes define the roll-up logic. `RollUp::Reset()` is used to clear the cache, this must be done at the start of a new event.

@AndyChappell Does this follow the your suggested design?

### DIFF
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -587,6 +587,8 @@ public:
      */
     static bool IsPairProduction(const pandora::MCParticle *const pMCParticle);
 
+    static bool CausesShower(const pandora::MCParticle *const pMC, int nDescendantElectrons = 0);
+
     /**
      *  @brief  Determine if two MC particles are topologically continuous within a given tolerance.
      *          If the parent does not travel any distance, a travelling parent is sought and the comparison made between this and the child.
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -245,6 +245,15 @@ public:
     static const pandora::MCParticle *GetParentMCParticle(const pandora::MCParticle *const pMCParticle);
 
     /**
+     *  @brief  Get the next parent mc particle one tier below the current mc particle
+     *
+     *  @param  pMCParticle the input mc particle
+     *
+     *  @return address of the parent mc particle one tier away
+     */
+    static const pandora::MCParticle *GetNextParentMCParticle(const pandora::MCParticle *const pMCParticle);
+
+    /**
      *  @brief  Get all descendent mc particles
      *
      *  @param  pMCParticle the input mc particle
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -587,6 +587,15 @@ public:
      */
     static bool IsPairProduction(const pandora::MCParticle *const pMCParticle);
 
+    /**
+     *  @brief  Check whether an MC particle's decay tree contains more than one electron, indicating an EM shower.
+     *
+     *  @param  pMC                   The MC particle that defines the root of the subtree to check
+     *  @param  nDescendantElectrons  The running count of electrons seen so far,
+     *                                this is a recursive parameter that should be 0 at the top level
+     *
+     *  @return  Whether the MC particle's subtree contains more than one electron
+     */
     static bool CausesShower(const pandora::MCParticle *const pMC, int nDescendantElectrons = 0);
 
     /**
--- a/larpandoracontent/LArUtility/RollUp.cc
+++ b/larpandoracontent/LArUtility/RollUp.cc
@@ -1,0 +1,193 @@
+/**
+ *  @file   
+ *
+ *  @brief  
+ *
+ *  $Log: $
+ */
+
+#include "larpandoracontent/LArUtility/RollUp.h"
+
+#include "larpandoracontent/LArHelpers/LArMCParticleHelper.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+RollUpper::RollUpper(std::unique_ptr<IRollUpPolicy> policy) :
+    m_policy{std::move(policy)}
+{
+}
+
+const MCParticle *RollUpper::RollUpMC(const MCParticle *const pMC)
+{
+    if (m_mcCache.find(pMC) != m_mcCache.end())
+        return m_mcCache.at(pMC);
+
+    const MCParticle *pLeadingMC{pMC};
+    while (!pLeadingMC->IsRootParticle())
+    {
+        if (!m_policy->ShouldRollUpMC(pLeadingMC))
+            break;
+        pLeadingMC = LArMCParticleHelper::GetNextParentMCParticle(pLeadingMC);
+    }
+
+    m_mcCache.insert({pMC, pLeadingMC});
+
+    return pLeadingMC;
+}
+
+const MCParticle *RollUpper::RollUpCaloHit(const CaloHit *const pCaloHit)
+{
+    if (m_caloHitCache.find(pCaloHit) != m_caloHitCache.end())
+        return m_caloHitCache.at(pCaloHit);
+
+    const MCParticleWeightMap &weightMap{pCaloHit->GetMCParticleWeightMap()};
+    MCParticleWeightMap rollUpWeightMap;
+    for (const auto &[pMC, weight] : weightMap)
+        rollUpWeightMap[this->RollUpMC(pMC)] += weight;
+
+    const MCParticle *pMainMC{nullptr};
+    float maxWeight{0.f};
+    for (const auto &[pMC, weight] : rollUpWeightMap)
+    {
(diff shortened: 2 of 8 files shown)


larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -578,6 +587,8 @@ class LArMCParticleHelper
      */
     static bool IsPairProduction(const pandora::MCParticle *const pMCParticle);
 
+    static bool CausesShower(const pandora::MCParticle *const pMC, int nDescendantElectrons = 0);
```
AndyChappell, line 599: Needs doxygen

larpandoracontent/LArUtility/RollUp.h
```diff
@@ -0,0 +1,180 @@
...
+
+private:
+    float m_deltaRayParentWeightThreshold;                               ///< Minimum parent weight for a hit to be folded to the parent track
+    std::map<pandora::HitType, float> m_deltaRayLengthThresholdsSquared; ///< Squared per-view length thresholds for short delta ray folding
+};
+
+/**
+ *  @brief  The RollUpEMWithComptonFilterAndAmbiguousDeltaRayHitsPolicy class
+ *
+ *  A concisely named class that extends RollUpEMAndAmbiguousDeltaRayHitsPolicy with a Compton scatter filter.
```
AndyChappell, line 117: I chuckled

larpandoracontent/LArUtility/RollUp.h
```diff
@@ -0,0 +1,180 @@
...
+ *  Pure Compton chains (photon → electron(s) with no bremms) are not rolled up,
+ *  preventing potential associations been distant and diffuse hits.
```
AndyChappell: typo "been" -> "between"

larpandoracontent/LArUtility/RollUp.h
```diff
@@ -0,0 +1,180 @@
...
+private:
+    float m_deltaRayParentWeightThreshold;                               ///< Minimum parent weight for a hit to be folded to the parent track
+    std::map<pandora::HitType, float> m_deltaRayLengthThresholdsSquared; ///< Squared per-view length thresholds for short delta ray folding
+};
+
+/**
+ *  @brief  The RollUpEMWithComptonFilterAndAmbiguousDeltaRayHitsPolicy class
+ *
+ *  A concisely named class that extends RollUpEMAndAmbiguousDeltaRayHitsPolicy with a Compton scatter filter.
+ *  Pure Compton chains (photon → electron(s) with no bremms) are not rolled up,
```
AndyChappell: Might want to clarify the behaviour when there are Brems (also, typo "bremms")

larpandoracontent/LArUtility/RollUp.cc
```diff
@@ -0,0 +1,231 @@
...
+
+void RollUpper::Reset() const
+{
+    m_mcCache.clear();
+    m_caloHitCache.clear();
+}
+
+//-----------------------------------------------------------------------------------------------------------------------------------------
+
+const MCParticle *RollUpper::RollUpMC(const MCParticle *const pMC) const
```
AndyChappell, line 40: I think the parameter here can be null, right? Probably want an explicit case to handle this.

larpandoracontent/LArUtility/RollUp.cc
```diff
@@ -0,0 +1,231 @@
...
+    return pLeadingMC;
+}
+
+//-----------------------------------------------------------------------------------------------------------------------------------------
+
+bool RollUpEMPolicy::ShouldFoldCaloHit(
+    [[maybe_unused]] const CaloHit *const pCaloHit, [[maybe_unused]] const MCParticle *const pRolledUpMainMC) const
+{
+    return false;
+}
```
AndyChappell: So, to make sure I understand the intent here, essentially, the GetRollupTargetMC function determines the base level rollup for a calo hit, and this ShouldFoldCaloHit function is then applied in the case where you have some additional partial rollup case, like for the delta ray hits that sit directly along the parent muon, for example. So "false" here, can still imply that rollup is in play, just not at this extra level?

If that's an accurate summary of the situation, I wonder if the name of this function should change to make the intention more explicit? ApplyExtraHitRollup or something like that?

larpandoracontent/LArUtility/RollUp.cc
```diff
@@ -0,0 +1,231 @@
...
+    while (!pLeadingMC->IsRootParticle())
+    {
+        const MCParticle *const pParentMC{LArMCParticleHelper::GetNextParentMCParticle(pLeadingMC)};
+        if (!LArMCParticleHelper::IsEM(pParentMC))
+        {
+            // Is electron and parent is a charged track -> jump up the chain by one to roll-up delta ray
+            const bool parentIsCharged{PdgTable::GetParticleCharge(pParentMC->GetParticleId()) != 0};
+            if (pLeadingMC->GetParticleId() == E_MINUS && parentIsCharged)
+                pLeadingMC = pParentMC;
+            break;
```
AndyChappell, line 149: So this one is simply folding delta rays into their parent (charged) tracks, no further conditions (essentially like the standard simulation rollup)?

larpandoracontent/LArUtility/RollUp.cc
```diff
@@ -0,0 +1,231 @@
...
+    const CaloHit *const pCaloHit, const MCParticle *const pRolledUpMainMC) const
+{
+    // Can't be a delta ray? -> don't roll-up this hit
+    if (pRolledUpMainMC->IsRootParticle() || pRolledUpMainMC->GetParticleId() != E_MINUS)
+        return false;
+
+    const MCParticle *const pParentMC{LArMCParticleHelper::GetNextParentMCParticle(pRolledUpMainMC)};
+
+    // Is not a delta ray? -> don't roll-up this hit
+    const bool parentIsCharged{PdgTable::GetParticleCharge(pParentMC->GetParticleId()) != 0};
```
AndyChappell, line 180: Here

AndyChappell (changes requested):
Thanks Alex. This looks good to me, and in line with what I was imagining re the strategy/policy approach. A couple of clarifying queries and then some minor edits, and this looks good.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
AlexWilkinsonnn: I missed this one, added

larpandoracontent/LArUtility/RollUp.cc
AlexWilkinsonnn: Yeah I think atm it would fall over if passed null. I went with returning null if passed null but maybe it should throw. Im not really sure.

larpandoracontent/LArUtility/RollUp.cc
AlexWilkinsonnn: Yes that is correct. I agree the name could be better, I changed it to `ShouldFurtherRollUpCaloHit` .

larpandoracontent/LArUtility/RollUp.cc
AlexWilkinsonnn: Yes

larpandoracontent/LArUtility/RollUp.cc
AlexWilkinsonnn: What is here? 😁

AlexWilkinsonnn (commented):
Thanks Andy, I addressed all your comments.

AlexWilkinsonnn:
Hey @AndyChappell, I just noticed a minor bug in my cluster validation algorithm. I am going to sneak the fix through in this PR in a moment

AlexWilkinsonnn:
The latest commit ([0c5a288](https://github.com/PandoraPFA/LArContent/pull/274/commits/0c5a2881641dd92164b988cdcc112d2e45cb2ce5)) is pushing the limit of what should be in this PR but I though it was easier to sneak through here. 

It refactors the way I build the hit feature vector for the DL shower growing to be cleaner (I think) and have the distance to X gap feature be optional, which is necessary for the reduced workspace VD geometry as I just found out.

larpandoracontent/LArUtility/RollUp.cc
AndyChappell: My initial thought is that null is ok, because it would be consistent with the case where hits have no associated MC? (I think that is the case that could generate the null input here).

larpandoracontent/LArUtility/RollUp.cc
AndyChappell: Ha, that was just my keeping tracking of where I was in the review.

AndyChappell (approved):

PR #275, opened 2026-03-30
Adding functions to tag rock muons inside CosmicRayTaggingTool

GianfrancoIngratta:
In this PR, I added two functions inside CosmicRayTaggingTool to tag rock muons. The DUNE ND will be exposed to a high rate of rock muons (~100 per spill), so we would like to give users the possibility, when running cosmic-ray reconstruction, to tag “clear rock muons,” i.e., PFOs whose endpoints fall outside the fiducial volume specified in the input XML.

A parallel draft PR was created in LArRecoND https://github.com/PandoraPFA/LArRecoND/pull/59 to handle the rock muons reconstruction within the cosmic pass. The PR also includes a few slides for a more detailed explanation.

### DIFF
--- a/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.cc
+++ b/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.cc
@@ -30,9 +30,9 @@ CosmicRayTaggingTool::CosmicRayTaggingTool() :
     m_inTimeMargin(5.f),
     m_inTimeMaxX0(1.f),
     m_tagRockMuons(false),
-    m_marginX(5.f), // [cm]
-    m_marginY(5.f),
-    m_marginZ(5.f),
+    m_marginX(20.f),
+    m_marginY(10.f),
+    m_marginZ(10.f),
     m_maxNeutrinoCosTheta(0.2f),
     m_minCosmicCosTheta(0.6f),
     m_maxCosmicCurvature(0.04f),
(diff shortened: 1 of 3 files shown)


CrossR:
I've not reviewed this yet, but would it not make sense to have all of this completely in LArRecoND?

It would stop some ND specific code being added to the master algorithm, but also means the "Cosmic Ray Tagging Tool", could stay as a cosmic ray tagger, and we could add a new tool, built on top of the CR Tagger specifically called "Rock Muon Tagging Tool".

That approach may still need a few small changes to LArContent to enable overriding functions etc, but it could end up better overall for both LArContent + LArRecoND: Less ND specific code here, and we'd have better, more appropriate names and separation of concerns for the CR vs Rock Muon tools, which we will likely need both of.

From a quick look, it seems like we'd basically have to only override the `FindAmbiguousPfos` method, and the rest would be fine as new methods. It could potentially help with the reco/tagging too, as things like `CheckIfTopToBottom` / `CheckIfInTime` etc may not be completely appropriate for Rock Muons (They aren't really top-to-bottom, and they are broadly in time with the beam since they came from it), so we could have more / less specific checks for each, without worrying about the overlap.

Some light refactoring of `FindAmbiguousPfos` could be useful though, just to tidy it up and enable it to be re-used outside more easily.

Happy to chat though, and of course see Andy's and others opinions.

johnmarshall80:
Hi, I'll just add some quick words of agreement to the thoughts from @CrossR, above. It would be ideal to use careful inheritance structures to minimise code duplication, and then to have separate tools for separate tasks, avoiding coupling of implementation for different use-cases/tasks.

AndyChappell:
Also agreeing with prior comments. A new tool for rock muons, inheriting from the cosmic ray tool seems like a good option here. As noted, it avoids adding ND-specific code to the other experiment use cases, and in particular, avoids changes to default parameters like m_marginY(10.f), which would impact all of our other experimental use cases.

I would also echo Ryan's comments re refactoring. Where we start finding that we essentially want to reuse significant parts of logic from one algorithm/tool in another, but can't trivially inherit due to notable detector differences, that would tend to suggest an opportunity for some refactoring, and potentially new helper functions that can provide that logic to multiple clients.

GianfrancoIngratta:
Thanks everyone for providing these comments! Sounds good! I'll leave this PR open for the moment, while I move stuff to LArRecoND. I guess this one gave us a good reference on how to operate when deciding whether to add new tools/duplicate code or refactor things in LArContent.

PR #278, opened 2026-04-14
CosmicRayTaggingTool attributes from private to protected

GianfrancoIngratta:
Following the discussion on [PR#275](https://github.com/PandoraPFA/LArContent/pull/275), I've created a new tool in LArRecoND called RockMuonTaggingTool, which is a derived class from CosmicRayTaggingTool. However to access the parent class attributes (such as [CRCandidates ](https://github.com/PandoraPFA/LArContent/blob/bd0903d3253508cb86b9a6899eb651e516ee0c6b/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h#L49), [dector boundaries](https://github.com/PandoraPFA/LArContent/blob/bd0903d3253508cb86b9a6899eb651e516ee0c6b/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h#L212-L217), etc. )) these need to be declared protected.

### DIFF
--- a/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h
+++ b/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h
@@ -32,7 +32,7 @@ public:
     pandora::StatusCode Initialize();
     void FindAmbiguousPfos(const pandora::PfoList &parentCosmicRayPfos, pandora::PfoList &ambiguousPfos, const MasterAlgorithm *const pAlgorithm);
 
-private:
+protected:
     /**
      *  @brief  Class to encapsulate the logic required determine if a Pfo should or shouldn't be tagged as a cosmic ray
      */
--- a/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h
+++ b/larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h
@@ -68,6 +68,8 @@ protected:
 
     typedef std::list<CRCandidate> CRCandidateList;
 
+private:
+
     /**
      *  @brief  Get the 3D calo hit cluster associated with a given Pfo, and check if it has sufficient hits
      *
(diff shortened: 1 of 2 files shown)


AndyChappell:
Hi @GianfrancoIngratta. Can I just clarify, what's the complete set of classes and attributes for which access is needed? The change as made increases the access boundary to a broad range of attributes and functions - is a more targeted increase in the access boundary possible?

CrossR:
From looking, I think the bits that do need to be publicly accessible are:

 - `CRCandidate` + `std::vector<CRCandidate>`
 - The detector boundary definitions
 - A couple of other random typedefs (UIntSet etc)

The rest is not used / not needed I don't think, so there could likely be a reduction in how many things are being marked `protected`.

GianfrancoIngratta:
> From looking, I think the bits that do need to be publicly accessible are:
> 
> * `CRCandidate` + `std::vector<CRCandidate>`
> * The detector boundary definitions
> * A couple of other random typedefs (UIntSet etc)
> 
> The rest is not used / not needed I don't think, so there could likely be a reduction in how many things are being marked `protected`.

Hi @CrossR, I believe I've addressed your comment. Let me know if there's anything more I should change.

larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h
AndyChappell: I think I would move the function and variables into the protected block above, rather than switching between protected and private multiple times. This will need an update to the corresponding CC file to ensure the constructor initialisation list ordering is correct.

AndyChappell (changes requested):
Minor update requested, otherwise I think this looks good. Thanks Gianfranco.

larpandoracontent/LArControlFlow/CosmicRayTaggingTool.h
GianfrancoIngratta: Thanks Andy. Yeah the reason I alternated private and protected was just to keep the cc file as it is and have as little changes as possible. But sure, will do!

PR #279, opened 2026-04-17
Validation algorithms

imawby:
Hello there!

Here is the metrics code. The metrics take the form of a master algorithm (ValidationAlgorithm.cc) which calls a series of tools (Event, Hierarchy, PFP, Shower and Track) that each fill a respective tree.

I've added a few things to the MCParticle:
- end direction
- trajectory points 
- visible energy

The only thing I need to do is to change the implementation so that the metrics are ran via a steering parameter rather than by an xml snippet. 

This PR is accompanied by a larpandora PR which fills the new MCParticle parameters.

### DIFF
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -152,6 +152,9 @@ void MasterAlgorithm::StitchPfos(
 
 StatusCode MasterAlgorithm::Run()
 {
+    if (m_pSliceNuWorkerInstance)
+        PandoraApi::SetEventInformation(*m_pSliceNuWorkerInstance, this->GetPandora().GetRun(), this->GetPandora().GetSubrun(), this->GetPandora().GetEvent());
+
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->Reset());
 
     if (!m_workerInstancesInitialized)
@@ -915,6 +918,8 @@ const Pandora *MasterAlgorithm::CreateWorkerInstance(
     PANDORA_THROW_RESULT_IF(
         STATUS_CODE_SUCCESS, !=, PandoraApi::SetLArTransformationPlugin(*pPandora, new lar_content::LArRotationalTransformationPlugin));
     PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->RegisterCustomContent(pPandora));
+    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraApi::SetEventInformation(*pPandora, this->GetPandora().GetRun(), 
+        this->GetPandora().GetSubrun(), this->GetPandora().GetEvent()));
     MultiPandoraApi::AddDaughterPandoraInstance(&(this->GetPandora()), pPandora);
 
     // The LArTPC
@@ -990,6 +995,8 @@ const Pandora *MasterAlgorithm::CreateWorkerInstance(
     PANDORA_THROW_RESULT_IF(
         STATUS_CODE_SUCCESS, !=, PandoraApi::SetLArTransformationPlugin(*pPandora, new lar_content::LArRotationalTransformationPlugin));
     PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->RegisterCustomContent(pPandora));
+    PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraApi::SetEventInformation(*pPandora, this->GetPandora().GetRun(), 
+        this->GetPandora().GetSubrun(), this->GetPandora().GetEvent()));
     MultiPandoraApi::AddDaughterPandoraInstance(&(this->GetPandora()), pPandora);
 
     // The Parent LArTPC
--- a/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
+++ b/larpandoracontent/LArControlFlow/MasterAlgorithm.cc
@@ -152,9 +152,6 @@ void MasterAlgorithm::StitchPfos(
 
 StatusCode MasterAlgorithm::Run()
 {
-    if (m_pSliceNuWorkerInstance)
-        PandoraApi::SetEventInformation(*m_pSliceNuWorkerInstance, this->GetPandora().GetRun(), this->GetPandora().GetSubrun(), this->GetPandora().GetEvent());
-
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->Reset());
 
     if (!m_workerInstancesInitialized)
@@ -163,6 +160,9 @@ StatusCode MasterAlgorithm::Run()
     if (m_passMCParticlesToWorkerInstances)
         PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->CopyMCParticles());
 
+    if (m_pSliceNuWorkerInstance)
+        PandoraApi::SetEventInformation(*m_pSliceNuWorkerInstance, this->GetPandora().GetRun(), this->GetPandora().GetSubrun(), this->GetPandora().GetEvent());
+    
     PfoToFloatMap stitchedPfosToX0Map;
     VolumeIdToHitListMap volumeIdToHitListMap;
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->GetVolumeIdToHitListMap(volumeIdToHitListMap));
--- a/larpandoracontent/LArHelpers/LArMCParticleHelper.h
+++ b/larpandoracontent/LArHelpers/LArMCParticleHelper.h
@@ -130,6 +130,15 @@ public:
     static bool IsNeutrino(const pandora::MCParticle *const pMCParticle);
 
     /**
+     *  @brief  Whether a mc particle is part of a CC interaction
+     *
+     *  @param[in]  pMCParticle the input mc particle
+     *
+     *  @return boolean
+     */
+    static bool IsCCInteraction(const pandora::MCParticle *const pMCParticle);
+    
+    /**
      *  @brief  Whether a provided mc particle matches the implemented definition of being primary
      *
      *  @param  pMCParticle the address of the mc particle
--- a/larpandoracontent/LArMetrics/BaseValidationTool.h
+++ b/larpandoracontent/LArMetrics/BaseValidationTool.h
@@ -22,23 +22,35 @@ namespace lar_content
 class BaseValidationTool : public pandora::AlgorithmTool
 {
 public:
+
+    /**
+     *  @brief  Default constructor
+     */
+    BaseValidationTool();
+
     virtual pandora::StatusCode Run(const pandora::Algorithm *const pAlgorithm, const pandora::MCParticle *const pMCNu, 
         const LArHierarchyHelper::MCMatchesVector &mcMatchesVec, const pandora::MCParticleVector &targetMC, 
         const pandora::PfoVector &bestRecoMatch) = 0;
 
 protected:
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+
     /**
      *  @brief  Pick out the hits of a specified type from an input list of hits
      *
-     *  @param  inputList the input hit list
-     *  @param  hitType the hit type
-     *  @param  outputVector the output vector of hits
-     *  @param  totalEnergy the total charge of the filtered hits
-     *
-     *  @return boolean whether the clusters are associated
+     *  @param[in]   inputList the input hit list
+     *  @param[in]   hitType the hit type
+     *  @param[out]  outputVector the output vector of hits
+     *  @param[out]  totalEnergy the total charge of the filtered hits
      */
     void GetHitsOfType(const pandora::CaloHitList &inputList, const pandora::HitType hitType, pandora::CaloHitVector &outputVector, 
         float &totalEnergy);
+
+    float m_maxMichelSep;      ///< the maximum separation between a michel vertex and parent muon endpoint    
+    float m_invalidSmallFloat; ///< small float value used for failure case
+    float m_invalidLargeFloat; ///< large float value used for failure case    
+    int m_invalidInt;          ///< int value used for failure case    
+    float m_invalidAngle;      ///< angle value used for failure case
 };
 
 } // namespace lar_content
--- a/larpandoracontent/LArMetrics/BaseValidationTool.h
+++ b/larpandoracontent/LArMetrics/BaseValidationTool.h
@@ -47,8 +47,8 @@ protected:
         float &totalEnergy);
 
     float m_maxMichelSep;      ///< the maximum separation between a michel vertex and parent muon endpoint    
-    float m_invalidSmallFloat; ///< small float value used for failure case
-    float m_invalidLargeFloat; ///< large float value used for failure case    
+    float m_invalidSmallFloat; ///< small float value used for failure case (visible on plots)
+    float m_invalidLargeFloat; ///< large float value used for failure case (used when small value is a valid value)
     int m_invalidInt;          ///< int value used for failure case    
     float m_invalidAngle;      ///< angle value used for failure case
 };
--- a/larpandoracontent/LArMetrics/BaseValidationTool.h
+++ b/larpandoracontent/LArMetrics/BaseValidationTool.h
@@ -1,0 +1,46 @@
+/**
+ *  @file   larpandoracontent/LArMetrics/BaseValidationTool.h
+ *
+ *  @brief  Header file for the base validation tool class.
+ *
+ *  $Log: $
+ */
+#ifndef BASE_VALIDATION_TOOL_H
+#define BASE_VALIDATION_TOOL_H 1
+
+#include "Pandora/Algorithm.h"
+#include "Pandora/AlgorithmTool.h"
+
+#include "larpandoracontent/LArHelpers/LArHierarchyHelper.h"
+
+namespace lar_content
+{
+
+/**
+ *  @brief  BaseValidationTool class
+ */
+class BaseValidationTool : public pandora::AlgorithmTool
+{
+public:
+    virtual pandora::StatusCode Run(const pandora::Algorithm *const pAlgorithm, const pandora::MCParticle *const pMCNu, 
+        const LArHierarchyHelper::MCMatchesVector &mcMatchesVec, const pandora::MCParticleVector &targetMC, 
+        const pandora::PfoVector &bestRecoMatch) = 0;
+
+protected:
+    /**
+     *  @brief  Pick out the hits of a specified type from an input list of hits
+     *
+     *  @param  inputList the input hit list
+     *  @param  hitType the hit type
+     *  @param  outputVector the output vector of hits
+     *  @param  totalEnergy the total charge of the filtered hits
+     *
+     *  @return boolean whether the clusters are associated
+     */
+    void GetHitsOfType(const pandora::CaloHitList &inputList, const pandora::HitType hitType, pandora::CaloHitVector &outputVector, 
+        float &totalEnergy);
+};
+
+} // namespace lar_content
+
+#endif // #ifndef BASE_VALIDATION_TOOL_H
--- a/larpandoracontent/LArMetrics/EventValidationTool.cc
+++ b/larpandoracontent/LArMetrics/EventValidationTool.cc
@@ -27,8 +27,8 @@ EventValidationTool::EventValidationTool() :
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 StatusCode EventValidationTool::Run(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, 
-    const LArHierarchyHelper::MCMatchesVector &/*mcMatchesVec*/, const MCParticleVector &targetMC, 
-    const PfoVector &/*bestRecoMatch*/)
+    [[maybe_unused]] const LArHierarchyHelper::MCMatchesVector &mcMatchesVec, const MCParticleVector &targetMC, 
+    [[maybe_unused]] const PfoVector &bestRecoMatch)
 {
     if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
         std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
@@ -51,10 +51,10 @@ StatusCode EventValidationTool::Run(const Algorithm *const pAlgorithm, const MCP
 void EventValidationTool::GetVertexVariables(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, EventTreeVars &eventTreeVars)
 {
     eventTreeVars.m_trueNuVertex = pMCNu->GetVertex();
-    eventTreeVars.m_recoNuVertexPass1 = CartesianVector(-9999.f, -9999.f, -9999.f);
-    eventTreeVars.m_recoNuVertexPass2 = CartesianVector(-9999.f, -9999.f, -9999.f);
-    eventTreeVars.m_recoNuVertexAccPass1 = -1.f;
-    eventTreeVars.m_recoNuVertexAccPass2 = -1.f;
+    eventTreeVars.m_recoNuVertexPass1 = CartesianVector(m_invalidLargeFloat, m_invalidLargeFloat, m_invalidLargeFloat);
+    eventTreeVars.m_recoNuVertexPass2 = CartesianVector(m_invalidLargeFloat, m_invalidLargeFloat, m_invalidLargeFloat);
+    eventTreeVars.m_recoNuVertexAccPass1 = m_invalidSmallFloat;
+    eventTreeVars.m_recoNuVertexAccPass2 = m_invalidSmallFloat;
 
     const VertexList *pNuVertexList_pass1(nullptr), *pNuVertexList_pass2(nullptr);
     PandoraContentApi::GetList(*pAlgorithm, m_nuVertexPass1ListName, pNuVertexList_pass1);
@@ -79,35 +79,12 @@ void EventValidationTool::GetInteractionTypeVariables(const MCParticle *const pM
 {
     eventTreeVars.m_nuPDG = pMCNu->GetParticleId();
     eventTreeVars.m_nuEnergy = pMCNu->GetEnergy(); 
+    eventTreeVars.m_isCC = LArMCParticleHelper::IsCCInteraction(pMCNu);
+    
     const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle *>(pMCNu));
-    eventTreeVars.m_nuVisEnergy = pLArMCParticle->GetVisibleEnergy();
+    if (!pLArMCParticle) { throw StatusCodeException(STATUS_CODE_FAILURE); }
 
-    // CC or NC?
-    eventTreeVars.m_isCC = 0;
-    if (std::abs(pMCNu->GetParticleId()) == 12)
-    {
-        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
-        {
-            if (std::abs(pMCParticle->GetParticleId()) == 11)
-                eventTreeVars.m_isCC = 1;
-        }
-    }
-    else if (std::abs(pMCNu->GetParticleId()) == 14)
-    {
-        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
-        {
-            if (std::abs(pMCParticle->GetParticleId()) == 13)
-                eventTreeVars.m_isCC = 1;
-        }
-    }
-    else if (std::abs(pMCNu->GetParticleId()) == 16)
-    {
-        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
-        {
-            if (std::abs(pMCParticle->GetParticleId()) == 15)
-                eventTreeVars.m_isCC = 1;
-        }
-    }
+    eventTreeVars.m_nuVisEnergy = pLArMCParticle->GetVisibleEnergy();
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
@@ -146,7 +123,7 @@ StatusCode EventValidationTool::ReadSettings(const TiXmlHandle xmlHandle)
     PANDORA_RETURN_RESULT_IF_AND_IF(
         STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "NuVertexPass2ListName", m_nuVertexPass2ListName));
 
-    return STATUS_CODE_SUCCESS;
+    return BaseValidationTool::ReadSettings(xmlHandle);
 }
 
 } // namespace lar_content
--- a/larpandoracontent/LArMetrics/EventValidationTool.cc
+++ b/larpandoracontent/LArMetrics/EventValidationTool.cc
@@ -1,0 +1,152 @@
+/**
+ *  @file   larpandoracontent/LArMetrics/EventValidationTool.cc
+ *
+ *  @brief  Implementation of the event validation tool class.
+ *
+ *  $Log: $
+ */
+
+#include "Pandora/AlgorithmHeaders.h"
+
+#include "larpandoracontent/LArHelpers/LArMCParticleHelper.h"
+#include "larpandoracontent/LArHelpers/LArInteractionTypeHelper.h"
+
+#include "larpandoracontent/LArMetrics/EventValidationTool.h"
+
+using namespace pandora;
+
+namespace lar_content
+{
+
+EventValidationTool::EventValidationTool() :
+    m_nuVertexPass1ListName("NeutrinoVertices3D_Pass1"),
+    m_nuVertexPass2ListName("NeutrinoVertices3D")
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode EventValidationTool::Run(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, 
+    const LArHierarchyHelper::MCMatchesVector &/*mcMatchesVec*/, const MCParticleVector &targetMC, 
+    const PfoVector &/*bestRecoMatch*/)
+{
+    if (PandoraContentApi::GetSettings(*pAlgorithm)->ShouldDisplayAlgorithmInfo())
+        std::cout << "----> Running Algorithm Tool: " << this->GetInstanceName() << ", " << this->GetType() << std::endl;
+
+    EventTreeVars eventTreeVars;
+    eventTreeVars.m_run = this->GetPandora().GetRun();
+    eventTreeVars.m_subrun = this->GetPandora().GetSubrun();
+    eventTreeVars.m_event = this->GetPandora().GetEvent();
+    eventTreeVars.m_nTargets = targetMC.size();
+
+    this->GetVertexVariables(pAlgorithm, pMCNu, eventTreeVars);
+    this->GetInteractionTypeVariables(pMCNu, eventTreeVars);
+    this->FillTree(eventTreeVars);
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void EventValidationTool::GetVertexVariables(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, EventTreeVars &eventTreeVars)
+{
+    eventTreeVars.m_trueNuVertex = pMCNu->GetVertex();
+    eventTreeVars.m_recoNuVertexPass1 = CartesianVector(-9999.f, -9999.f, -9999.f);
+    eventTreeVars.m_recoNuVertexPass2 = CartesianVector(-9999.f, -9999.f, -9999.f);
+    eventTreeVars.m_recoNuVertexAccPass1 = -1.f;
+    eventTreeVars.m_recoNuVertexAccPass2 = -1.f;
+
+    const VertexList *pNuVertexList_pass1(nullptr), *pNuVertexList_pass2(nullptr);
+    PandoraContentApi::GetList(*pAlgorithm, m_nuVertexPass1ListName, pNuVertexList_pass1);
+    PandoraContentApi::GetList(*pAlgorithm, m_nuVertexPass2ListName, pNuVertexList_pass2);
+
+    if (pNuVertexList_pass1 && !pNuVertexList_pass1->empty())
+    {
+        eventTreeVars.m_recoNuVertexPass1 = pNuVertexList_pass1->front()->GetPosition();
+        eventTreeVars.m_recoNuVertexAccPass1 = (eventTreeVars.m_trueNuVertex - eventTreeVars.m_recoNuVertexPass1).GetMagnitude();
+    }
+
+    if (pNuVertexList_pass2 && !pNuVertexList_pass2->empty())
+    {
+        eventTreeVars.m_recoNuVertexPass2 = pNuVertexList_pass2->front()->GetPosition();
+        eventTreeVars.m_recoNuVertexAccPass2 = (eventTreeVars.m_trueNuVertex - eventTreeVars.m_recoNuVertexPass2).GetMagnitude();
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void EventValidationTool::GetInteractionTypeVariables(const MCParticle *const pMCNu, EventTreeVars &eventTreeVars)
+{
+    eventTreeVars.m_nuPDG = pMCNu->GetParticleId();
+    eventTreeVars.m_nuEnergy = pMCNu->GetEnergy(); 
+    const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle *>(pMCNu));
+    eventTreeVars.m_nuVisEnergy = pLArMCParticle->GetVisibleEnergy();
+
+    // CC or NC?
+    eventTreeVars.m_isCC = 0;
+    if (std::abs(pMCNu->GetParticleId()) == 12)
+    {
+        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
+        {
+            if (std::abs(pMCParticle->GetParticleId()) == 11)
+                eventTreeVars.m_isCC = 1;
+        }
+    }
+    else if (std::abs(pMCNu->GetParticleId()) == 14)
+    {
+        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
+        {
+            if (std::abs(pMCParticle->GetParticleId()) == 13)
+                eventTreeVars.m_isCC = 1;
+        }
+    }
+    else if (std::abs(pMCNu->GetParticleId()) == 16)
+    {
+        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
+        {
+            if (std::abs(pMCParticle->GetParticleId()) == 15)
+                eventTreeVars.m_isCC = 1;
+        }
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void EventValidationTool::FillTree(EventTreeVars &eventTreeVars)
+{
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "Run", eventTreeVars.m_run));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "Subrun", eventTreeVars.m_subrun));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "Event", eventTreeVars.m_event));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCEvent_NTargets", eventTreeVars.m_nTargets));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCNu_PDG", eventTreeVars.m_nuPDG));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCNu_Energy", eventTreeVars.m_nuEnergy));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCNu_VisEnergy", eventTreeVars.m_nuVisEnergy));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCInt_IsCC", eventTreeVars.m_isCC));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCNu_VertexX", eventTreeVars.m_trueNuVertex.GetX()));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCNu_VertexY", eventTreeVars.m_trueNuVertex.GetY()));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "MCNu_VertexZ", eventTreeVars.m_trueNuVertex.GetZ()));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "EventTree", "RecoNu_VertexX_Pass1", eventTreeVars.m_recoNuVertexPass1.GetX()));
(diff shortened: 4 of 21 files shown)


larpandoracontent/LArControlFlow/MasterAlgorithm.cc
```diff
@@ -152,6 +152,9 @@ void MasterAlgorithm::StitchPfos(
 
 StatusCode MasterAlgorithm::Run()
 {
+    if (m_pSliceNuWorkerInstance)
+        PandoraApi::SetEventInformation(*m_pSliceNuWorkerInstance, this->GetPandora().GetRun(), this->GetPandora().GetSubrun(), this->GetPandora().GetEvent());
+
```
AndyChappell: Before Reset?

larpandoracontent/LArMetrics/BaseValidationTool.h
```diff
@@ -0,0 +1,46 @@
...
+protected:
+    /**
+     *  @brief  Pick out the hits of a specified type from an input list of hits
+     *
+     *  @param  inputList the input hit list
+     *  @param  hitType the hit type
+     *  @param  outputVector the output vector of hits
+     *  @param  totalEnergy the total charge of the filtered hits
+     *
+     *  @return boolean whether the clusters are associated
```
AndyChappell: No return value.

larpandoracontent/LArMetrics/BaseValidationTool.h
```diff
@@ -0,0 +1,46 @@
...
+        const pandora::PfoVector &bestRecoMatch) = 0;
+
+protected:
+    /**
+     *  @brief  Pick out the hits of a specified type from an input list of hits
+     *
+     *  @param  inputList the input hit list
+     *  @param  hitType the hit type
+     *  @param  outputVector the output vector of hits
+     *  @param  totalEnergy the total charge of the filtered hits
```
AndyChappell: Suggest noting this is an output variable as well. We've started making use of the `@param[in,out]` or `@param[out]` annotations to signify this, as appropriate.

larpandoracontent/LArMetrics/HierarchyValidationTool.h
```diff
@@ -0,0 +1,97 @@
...
+     *
+     *  @param  pBestMatch the best-matched pfo
+     *  @param  bestMatches the vector of best-matched pfos
+     *  @param  hierarchyTreeVars the hierarchy tree variables
+     */
+    void FillRecoVariables(const pandora::Pfo *const pBestMatch, const pandora::PfoVector &bestMatches, 
+        HierarchyTreeVars &hierarchyTreeVars);
+
+    /**
+     *  @brief  Fill ther hierarchy tree
```
AndyChappell: Typo

larpandoracontent/LArMetrics/PFPValidationTool.h
```diff
@@ -0,0 +1,171 @@
...
+    pandora::FloatVector m_purityW;    ///< purity of the pfo considering W-view hits
+    pandora::FloatVector m_purityADC;  ///< adc-weighted purity of the pfo considering all 2D hits
+    pandora::FloatVector m_purityADCU; ///< adc-weighted purity of the pfo considering U-view hits
+    pandora::FloatVector m_purityADCV; ///< adc-weighted purity of the pfo considering V-view hits
+    pandora::FloatVector m_purityADCW; ///< adc-weighted purity of the pfo considering W-view hits
+    pandora::FloatVector m_altCompleteness;      ///< completeness of the thief particle wrt the target particle
+    pandora::FloatVector m_altPurity;            ///< purity of the thief particle wrt the target particle
+    pandora::IntVector m_altPDG;                 ///< PDG of the MCParticle the thief best matches
+    pandora::IntVector m_altIsUpstreamHierarchy; ///< is the MCParticle that the thief best matches, an ancestor of this MCParticle?
+    pandora::IntVector m_altIsSameMC;            ///< is the MCParticle that the thief best matches, an ancestor of this MCParticle?
```
AndyChappell: Presumably the comment associated with the second variable is not correct?

larpandoracontent/LArMetrics/PFPValidationTool.h
```diff
@@ -0,0 +1,171 @@
...
+     *  @param  pfpTreeVars the pfo tree variables
+     */
+    void GetMatchingInfo(const LArHierarchyHelper::MCMatchesVector &mcMatchesVec,
+        const pandora::MCParticle *const pMCTarget, const pandora::Pfo *const pBestMatch, PFPTreeVars &pfpTreeVars);
+
+    /**
+     *  @brief  Fill the variables that describe the second-best MCParticle-pfo matches
+     *
+     *  @param  mcMatchesVec the mapping of true nodes (MCParticles) to reco node (pfo) matches
+     *  @param  pMCTarget the MCParticle
```
AndyChappell: Missing a parameter here

larpandoracontent/LArMetrics/ValidationAlgorithm.h
```diff
@@ -0,0 +1,59 @@
...
+
+    ~ValidationAlgorithm();
+
+private:
+    typedef std::vector<BaseValidationTool *> ValidationToolVector;
+
+    pandora::StatusCode Run();
+    pandora::StatusCode ReadSettings(const pandora::TiXmlHandle xmlHandle);
+
+    bool IsReconstructable(const LArHierarchyHelper::MCHierarchy::Node *pMCNode);
```
AndyChappell, line 46: doxygen

larpandoracontent/LArMetrics/ValidationAlgorithm.h
```diff
@@ -0,0 +1,59 @@
...
+
+    std::string m_caloHitListName;
+    std::string m_mcParticleListName;
+    pandora::StringVector m_pfoListNames;     ///< The pfo list names
+    std::string m_fileName;
+    std::string m_treeName;
```
AndyChappell: doxygen

larpandoracontent/LArMetrics/ValidationAlgorithm.h
```diff
@@ -0,0 +1,59 @@
...
+    float m_minPurity;                 ///< Minimum purity to tag a node as being of good quality
+    float m_minCompleteness;           ///< Minimum completeness to tag a node as being of good quality
+    unsigned int m_minRecoHits;        ///< Minimum number of reconstructed primary good hits
+    unsigned int m_minRecoHitsPerView; ///< Minimum number of reconstructed hits for a good view
+    unsigned int m_minRecoGoodViews;   ///< Minimum number of reconstructed primary good views
+    bool m_removeRecoNeutrons;         ///< Whether to remove reconstructed neutrons and their downstream particles
+    bool m_selectRecoHits;             ///< Whether to select reco hits that overlap with the MC particle hits
+
+    ValidationToolVector m_validationToolVector;
```
AndyChappell: doxygen

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -123,9 +129,57 @@ class LArMCParticle : public object_creation::MCParticle::Object
...
+
+    /**
+     *  @brief  Get the number of trajectory points
+     *
+     *  @return the number of trajectory points
+     */
+    int GetNTrajPoints() const;
+
+    /**
+     *  @brief  Get the x positions
```
AndyChappell: I might expand the description here. Note that it's along the trajectory, with the expectation that it's sparsified trajectories that only update when something changes.

larpandoracontent/LArMetrics/EventValidationTool.cc
```diff
@@ -0,0 +1,152 @@
...
+    m_nuVertexPass1ListName("NeutrinoVertices3D_Pass1"),
+    m_nuVertexPass2ListName("NeutrinoVertices3D")
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode EventValidationTool::Run(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, 
+    const LArHierarchyHelper::MCMatchesVector &/*mcMatchesVec*/, const MCParticleVector &targetMC, 
+    const PfoVector &/*bestRecoMatch*/)
```
AndyChappell: Suggest `[[maybe_unused]]` annotation for the two unused variables over comments

e.g. `[[maybe_unused]] const PfoVector &bestRecoMatch`

larpandoracontent/LArMetrics/EventValidationTool.cc
```diff
@@ -0,0 +1,152 @@
...
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void EventValidationTool::GetVertexVariables(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, EventTreeVars &eventTreeVars)
+{
+    eventTreeVars.m_trueNuVertex = pMCNu->GetVertex();
+    eventTreeVars.m_recoNuVertexPass1 = CartesianVector(-9999.f, -9999.f, -9999.f);
```
AndyChappell: Maybe define a constant in the base tool for INVALID_VALUE. Reduces the risk of a typo that might change behaviour.

larpandoracontent/LArMetrics/EventValidationTool.cc
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void EventValidationTool::GetInteractionTypeVariables(const MCParticle *const pMCNu, EventTreeVars &eventTreeVars)
+{
+    eventTreeVars.m_nuPDG = pMCNu->GetParticleId();
+    eventTreeVars.m_nuEnergy = pMCNu->GetEnergy(); 
+    const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle *>(pMCNu));
```
AndyChappell, line 84: Worth checking `if (pLArMCParticle)` before continuing here, in case the dynamic cast fails (a subsequent dereference would segfault).

larpandoracontent/LArMetrics/EventValidationTool.cc
```diff
@@ -0,0 +1,152 @@
...
+    eventTreeVars.m_nuVisEnergy = pLArMCParticle->GetVisibleEnergy();
+
+    // CC or NC?
+    eventTreeVars.m_isCC = 0;
+    if (std::abs(pMCNu->GetParticleId()) == 12)
```
AndyChappell: Prefer NU_MU from Pandora/PdgTable.h

larpandoracontent/LArMetrics/EventValidationTool.cc
```diff
@@ -0,0 +1,152 @@
...
+    {
+        for (const MCParticle *const pMCParticle : pMCNu->GetDaughterList())
+        {
+            if (std::abs(pMCParticle->GetParticleId()) == 11)
```
AndyChappell: Prefer E_MINUS from Pandora/PdgTable.h, etc.

larpandoracontent/LArMetrics/EventValidationTool.cc
```diff
@@ -0,0 +1,152 @@
...
+
+void EventValidationTool::GetInteractionTypeVariables(const MCParticle *const pMCNu, EventTreeVars &eventTreeVars)
+{
+    eventTreeVars.m_nuPDG = pMCNu->GetParticleId();
+    eventTreeVars.m_nuEnergy = pMCNu->GetEnergy(); 
+    const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle *>(pMCNu));
+    eventTreeVars.m_nuVisEnergy = pLArMCParticle->GetVisibleEnergy();
+
+    // CC or NC?
+    eventTreeVars.m_isCC = 0;
```
AndyChappell: Actually, for this entire block, can we just propagate the CC flag from LArSoft into Pandora? It's a little overkill because of the common MCParticle for both neutrinos and other particles, but could just add an m_isCC flag to the LArMCParticle and populate it from that, giving a nice clean implementation (can argue whether to only set a true flag againts neutrinos). I'm actually not sure this will work for the tau, because I believe Pythia handles the tau decay, in which case the tau itself may not appear in the MC hierarchy, only its subsequently leptonic or hadronic decay, so tau's might not ever register as CC under this method.

larpandoracontent/LArMetrics/HierarchyValidationTool.cc
```diff
@@ -0,0 +1,172 @@
...
+{
+
+HierarchyValidationTool::HierarchyValidationTool()
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode HierarchyValidationTool::Run(const Algorithm *const pAlgorithm, const MCParticle *const pMCNu, 
+    const LArHierarchyHelper::MCMatchesVector &/*mcMatchesVec*/, const MCParticleVector &targetMC,
```
AndyChappell: As previously `[[mayby_unused]]`

larpandoracontent/LArMetrics/HierarchyValidationTool.cc
```diff
@@ -0,0 +1,172 @@
...
+        // If reco parent is not in the best-match list, the particle has been split
+        // I don't want this to feed into a hierarchy building 'failure' so mark as -1 and remove from downstream metrics
+        if (bestMatchesIter == bestMatches.end())
+        {
+            hierarchyTreeVars.m_recoParentIndex.push_back(-1);
+        }
+        else
+        {
+            hierarchyTreeVars.m_recoParentIndex.push_back(bestMatchesIter - bestMatches.begin());
+        }
```
AndyChappell: Can drop the braces where every branch of the conditional is a one-liner.

larpandoracontent/LArMetrics/HierarchyValidationTool.cc
```diff
@@ -0,0 +1,172 @@
...
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "HierarchyTree", "MC_HierarchyTier", &trueTier));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "HierarchyTree", "MC_ParentIndex", &trueParentIndex));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "HierarchyTree", "BM_HierarchyTier", &recoTier));
+    PANDORA_MONITORING_API(SetTreeVariable(this->GetPandora(), "HierarchyTree", "BM_ParentIndex", &recoParentIndex));
+    PANDORA_MONITORING_API(FillTree(this->GetPandora(), "HierarchyTree"));
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode HierarchyValidationTool::ReadSettings(const TiXmlHandle /*xmlHandle*/)
```
AndyChappell: `[[maybe_unused]]`

larpandoracontent/LArMetrics/ValidationAlgorithm.cc
```diff
@@ -0,0 +1,227 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+ValidationAlgorithm::~ValidationAlgorithm()
+{
+    PANDORA_MONITORING_API(SaveTree(this->GetPandora(), "EventTree", m_fileName.c_str(), "UPDATE"));
+    PANDORA_MONITORING_API(SaveTree(this->GetPandora(), "HierarchyTree", m_fileName.c_str(), "UPDATE"));
+    PANDORA_MONITORING_API(SaveTree(this->GetPandora(), "PFPTree", m_fileName.c_str(), "UPDATE"));
+    PANDORA_MONITORING_API(SaveTree(this->GetPandora(), "TrackTree", m_fileName.c_str(), "UPDATE"));
+    PANDORA_MONITORING_API(SaveTree(this->GetPandora(), "ShowerTree", m_fileName.c_str(), "UPDATE"));
```
AndyChappell: c_str isn't necessary here.

larpandoracontent/LArMetrics/ValidationAlgorithm.cc
```diff
@@ -0,0 +1,227 @@
...
+        for (BaseValidationTool *const pValidationTool : m_validationToolVector)
+            pValidationTool->Run(this, nuParticlesVec.at(i), mcMatchesVec, targetMC, bestRecoMatch);
+    }
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool ValidationAlgorithm::IsReconstructable(const LArHierarchyHelper::MCHierarchy::Node *pMCNode)
```
AndyChappell, line 150: I think you're just replicating

https://github.com/PandoraPFA/LArContent/blob/bd0903d3253508cb86b9a6899eb651e516ee0c6b/larpandoracontent/LArHelpers/LArHierarchyHelper.cc#L707

?

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+
+        this->GetMCParticleInfo(pMCNu, pMC, pfpTreeVars);
+
+        if (pBestMatch)
+        {
+            this->GetRecoParticleInfo(pMC, pBestMatch, pfpTreeVars);
+        }
+        else
+        {
+            pfpTreeVars.m_recoVertexX.push_back(-9999.f);
```
AndyChappell: As previosuly mentioned re -9999 versus a constant defined in the base tool.

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void PFPValidationTool::GetMCParticleInfo(const MCParticle *const pMCNu, const MCParticle *const pMCTarget, PFPTreeVars &pfpTreeVars)
+{
+    // Energy
+    pfpTreeVars.m_trueEnergy.push_back(pMCTarget->GetEnergy());
+    const LArMCParticle *const pLArMCParticle(dynamic_cast<const LArMCParticle *>(pMCTarget));
```
AndyChappell, line 82: Check that the cast succeeded.

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+    pfpTreeVars.m_trueVisEnergy.push_back(pLArMCParticle->GetVisibleEnergy());
+
+    // Vertex
+    const CartesianVector &trueVertex((pMCTarget->GetParticleId() == 22) ? pMCTarget->GetEndpoint() : pMCTarget->GetVertex());
```
AndyChappell: PHOTON

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+    // Length
+    pfpTreeVars.m_trueLength.push_back((trueVertex - trueEnd).GetMagnitude());
+
+    // Initial direction
+    const CartesianVector &mcMom(pMCTarget->GetMomentum());
+    const float momMagSq(mcMom.GetMagnitudeSquared());
+    if (momMagSq < std::numeric_limits<float>::epsilon())
+    {
+        pfpTreeVars.m_trueThetaXZ.push_back(-4.f);
+        pfpTreeVars.m_trueThetaYZ.push_back(-4.f);
```
AndyChappell: -4 ?

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+        pfpTreeVars.m_trueDirX.push_back(-9999.f);
+        pfpTreeVars.m_trueDirY.push_back(-9999.f);
+        pfpTreeVars.m_trueDirZ.push_back(-9999.f);
+    }
+    else
+    {
+        pfpTreeVars.m_trueThetaXZ.push_back(atan2(mcMom.GetX(), mcMom.GetZ()));
```
AndyChappell, line 116: Is this angle with respect to the z axis? Basically, just checking that parameter ordering is correct, given the potentially counterintuitive definition of atan2 as atan2(y, x) yielding a quadrant aware arc tangent of y/x.

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+    else
+    {
+        pfpTreeVars.m_trueEndDirX.push_back(trueEndDir.GetX());
+        pfpTreeVars.m_trueEndDirY.push_back(trueEndDir.GetY());
+        pfpTreeVars.m_trueEndDirZ.push_back(trueEndDir.GetZ());
+    }
+
+    // TruePDG
+    int truePDG(pMCTarget->GetParticleId());
+    if (abs(pMCTarget->GetParticleId()) == 22)     // 111 = photon from pi0
```
AndyChappell: PHOTON etc

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+            truePDG = 111;
+    }
+    else if (std::abs(pMCTarget->GetParticleId()) == 11)    // 777 = michel electron
+    {
+        const MCParticle *const pMCParent(pMCTarget->GetParentList().front());
+
+        if ((abs(pMCParent->GetParticleId()) == 13) || (abs(pMCParent->GetParticleId()) == 211))
+        {
+            if ((pMCParent->GetEndpoint() - pMCTarget->GetVertex()).GetMagnitude() < m_maxMichelSep)
+                truePDG = 777;
```
AndyChappell: Suggest defining a named constant

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+    if (abs(pMCTarget->GetParticleId()) == 22)     // 111 = photon from pi0
+    {
+        if (pMCTarget->GetParentList().front()->GetParticleId() == 111)
+            truePDG = 111;
+    }
+    else if (std::abs(pMCTarget->GetParticleId()) == 11)    // 777 = michel electron
+    {
+        const MCParticle *const pMCParent(pMCTarget->GetParentList().front());
+
+        if ((abs(pMCParent->GetParticleId()) == 13) || (abs(pMCParent->GetParticleId()) == 211))
```
AndyChappell: I think you might also want to check the MC process for decay here (available in the LArMCParticle), as this presumably also succeeds where we have delta rays that aren't rolled up in sim.

larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -0,0 +1,544 @@
...
+    if (altMCIndex != -1)
+    {
+        // Is upstream hierarchy
+        const MCParticle *pSeed(pMCTarget);
+        while (pSeed->GetParentList().size() != 0)
+        {
+            if (pSeed->GetParentList().front() == targetMC.at(altMCIndex))
+            {
+                isUpstreamHierarchy = 1;
+            }
```
AndyChappell: No need for braces

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+
+ShowerValidationTool::ShowerValidationTool() :
+    m_trueLengthEnergyFrac(0.90),
+    m_initialRegion3D(14.f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode ShowerValidationTool::Run(const Algorithm *const pAlgorithm, const MCParticle *const /*pMCNu*/,
```
AndyChappell: `[[maybe_unused]]`

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ShowerValidationTool::GetTrueLength(const LArHierarchyHelper::MCMatchesVector &mcMatchesVec, const MCParticle *const pMCParticle,
+     ShowerTreeVars &showerTreeVars)
+{
+    // Get 3D positions/directions
+    const CartesianVector trueShrVtx(pMCParticle->GetVertex());
```
AndyChappell: Still going through code, so my sense of what's going on here might alter as I read more. But, under any definition, does it make sense to start the shower at the vertex for photons?

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+            const CartesianVector positionB(pCaloHitB->GetPositionVector() - trueShrVtx2D);
+            
+            const float lA(trueShrDir2D.GetDotProduct(positionA));
+            const float lB(trueShrDir2D.GetDotProduct(positionB));
+
+            return lA < lB;
+        });
+
+        // Calculate 'true length'
+        FloatVector &trueLength(hitType == TPC_VIEW_U ? showerTreeVars.m_coreTrueLengthFromU :
```
AndyChappell, line 124: Mainly annotating here, but I'm imagining that core length is going to turn out to be some function of interaction/radiation length (which makes sense to me, just retaining that open question of where we start measuring).

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+                                hitType == TPC_VIEW_V ? showerTreeVars.m_coreTrueLengthFromV : showerTreeVars.m_coreTrueLengthFromW);
+
+        float runningEnergySum(0.f), endpointL(-9999.f);
+        for (const CaloHit *const pHit2D : viewMCHits)
+        {
+            const float hitEnergy(std::fabs(pHit2D->GetElectromagneticEnergy()));
+            runningEnergySum += hitEnergy;
+
+            if ((totalEnergy > std::numeric_limits<float>::epsilon()) && 
+                ((runningEnergySum / totalEnergy) > m_trueLengthEnergyFrac))
```
AndyChappell, line 134: Yep, so basically the shower length is defined with respect to two interaction lengths?

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+            {
+                const CartesianVector displacement(pHit2D->GetPositionVector() - trueShrVtx2D);
+                endpointL = trueShrDir2D.GetDotProduct(displacement);
+                break;
+            }
+        }
+
+        if (endpointL > -9990.f)
```
AndyChappell: Guessing that's a typo - likely not a consequential one, but, constant in base class.

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void ShowerValidationTool::GetInitialRegionVars(const LArHierarchyHelper::MCMatchesVector &mcMatchesVec, const MCParticle *const pMCParticle, 
+    const Pfo *const pPfo, ShowerTreeVars &showerTreeVars)
+{
+    // Loop over matches to get MCNode
+    CaloHitList mcHits;
+    for (const LArHierarchyHelper::MCMatches &mcMatches : mcMatchesVec)
+    {
+        if (mcMatches.GetMC()->GetMCParticles().front() == pMCParticle)
```
AndyChappell, line 168: Presumably you could break out of the loop here

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -0,0 +1,523 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode ShowerValidationTool::ReadSettings(const TiXmlHandle xmlHandle)
+{
+    PANDORA_RETURN_RESULT_IF_AND_IF(
+        STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "InitialRegion", m_initialRegion3D));
+
+
```
AndyChappell: Excess whitespace

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+TrackValidationTool::TrackValidationTool() :
+    m_edgeBuffer(5.f),
+    m_slidingFitWindow(20),
+    m_endRegion(5.f)
+{
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode TrackValidationTool::Run(const Algorithm *const pAlgorithm, const MCParticle *const /*pMCNu*/,
```
AndyChappell: `[[maybe_unused]]`

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+        bool fitSuccess(pBestMatch && (this->FitTrack(pBestMatch, recoVertex, recoVertexDir, recoEndpoint, recoEndpointDir)));
+
+        if (fitSuccess)
+        {
+            this->GetRecoVertexAndEndpointVars(pMC, recoVertex, recoVertexDir, recoEndpoint, recoEndpointDir, trackTreeVars);
+        }
+        else
+        {
+            this->FillForFailedPfo(trackTreeVars);
+        }
```
AndyChappell: Don't need braces

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+
+    return STATUS_CODE_SUCCESS;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackValidationTool::GetTrueVertexAndEndpointVars(const MCParticle *const pMCParticle, TrackTreeVars &trackTreeVars)
+{
+    int isLeaving(0);
+    const LArMCParticle *const pLArMC(dynamic_cast<const LArMCParticle *>(pMCParticle));
```
AndyChappell, line 78: Check the dynamic cast worked

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+
+    return true;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackValidationTool::GetRecoVertexAndEndpointVars(const MCParticle *const pMCParticle, const CartesianVector &recoVertex, const CartesianVector &recoVertexDir,
+    const CartesianVector &recoEndpoint, const CartesianVector &recoEndpointDir, TrackTreeVars &trackTreeVars)
+{
+    const LArMCParticle *const pLArMC(dynamic_cast<const LArMCParticle *>(pMCParticle));
```
AndyChappell, line 156: Check dynamic cast

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+    endpointAcc *= sign;
+
+    // Start direction
+    float startDirAcc(doesMCHaveStartDir ? pMCParticle->GetMomentum().GetOpeningAngle(recoVertexDir) : -1.f);
+
+    // End direction
+    float endDirAcc(doesMCHaveEndDir ? pLArMC->GetEndDirection().GetOpeningAngle(recoEndpointDir) : -1.f);
+
+    // Is flipped?
+    int isOrientationCorrect((trueEndpoint - recoEndpoint).GetMagnitudeSquared() < (trueEndpoint - recoVertex).GetMagnitudeSquared() ? 1 : 0);
```
AndyChappell: Is this the right check? In principle, if you had poor purity(?), this might claim a flip where there is none? It's a bit of an edge case I think, but would it not be better to check the opening angle between the (end - start) vectors for reco and true and check for more than 90 degrees?

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+    trackTreeVars.m_recoEndDirZ.push_back(recoEndpointDir.GetZ());
+    trackTreeVars.m_endDirAcc.push_back(endDirAcc);
+    trackTreeVars.m_isCorrectOrientation.push_back(isOrientationCorrect);
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+void TrackValidationTool::FillForFailedPfo(TrackTreeVars &trackTreeVars)
+{
+    trackTreeVars.m_recoEndpointX.push_back(-9999.f);
```
AndyChappell: Usual comment.

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+        trackTreeVars.m_endpointPurityW.push_back(-1.f);
+        return;
+    }
+
+    // Need to loop over matches to find this match :(
+    CaloHitList mcHits;
+    for (const LArHierarchyHelper::MCMatches &mcMatches : mcMatchesVec)
+    {
+        if (mcMatches.GetMC()->GetMCParticles().front() == pMCParticle)
+            mcHits = mcMatches.GetMC()->GetCaloHits();
```
AndyChappell, line 256: Presumably you can break here though?

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+
+        // Find end region MC hits
+        float totalEnergy(0.f); // need for function call
+        CaloHitVector viewMCHits, endMCHits;
+        this->GetHitsOfType(mcHits, hitType, viewMCHits, totalEnergy);
+        for (const CaloHit *const pCaloHit : viewMCHits)
+        {
+            const float l(endDirection2D.GetDotProduct(pCaloHit->GetPositionVector() - endRegionStart2D));
+
+            if ((l > 0.f) && (l < endRegionL))
```
AndyChappell: Stricly less than and greater than?

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+        {
+            // Find end region pfo hits
+            CaloHitList viewPfoHitList;
+            LArPfoHelper::GetCaloHits(pPfo, hitType, viewPfoHitList);
+            CaloHitVector viewPfoHits(viewPfoHitList.begin(), viewPfoHitList.end()), endPfoHits;
+            for (const CaloHit *const pCaloHit : viewPfoHits)
+            {                
+                const float l(endDirection2D.GetDotProduct(pCaloHit->GetPositionVector() - endRegionStart2D));
+
+                if ((l > 0.f) & (l < endRegionL))
```
AndyChappell: As above

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+        {
+            for (const MCParticle *const pMCChild : pMCParent->GetDaughterList())
+            {
+                if (std::abs(pMCChild->GetParticleId()) != 11)
+                    continue;
+
+                if ((pMCParent->GetEndpoint() - pMCChild->GetVertex()).GetMagnitude() > 3.f)
+                    continue;
+
+                this_michelFromMuon = 1;
```
AndyChappell, line 370: As elsewhere, may want to check for a decay process to avoid collecting unrolled deltas.

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+                    // look for electron!
+                    for (const MCParticle *const pMCGrandchild : pMCChild->GetDaughterList())
+                    {
+                        if (std::abs(pMCGrandchild->GetParticleId()) != 11)
+                            continue;
+
+                        if ((pMCChild->GetEndpoint() - pMCGrandchild->GetVertex()).GetMagnitude() > 3.f)
+                            continue;
+
+                        pMCMichel = pMCGrandchild;
```
AndyChappell, line 398: As above, check for decay process?

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -0,0 +1,525 @@
...
+
+            if (this_hasTargetMichel)
+            {
+                this_michelIndex = michelIter - targetMC.begin();
+
+                // Has target michel been reco'd
+                if (bestRecoMatch.at(this_michelIndex))
+                {
+                    this_hasRecoMichel = 1;
+                    this_michelIsShower = LArPfoHelper::IsShower(bestRecoMatch.at(this_michelIndex));
```
AndyChappell, line 424: Presumably this is for the analyst trap of "it's a michel, it must be a track"?

AndyChappell (changes requested):
Thanks Isobel. This is looking pretty close. Mostly fairly minor comments, but a couple of more substantive ones. Let me know if you have any questions.

larpandoracontent/LArControlFlow/MasterAlgorithm.cc
imawby: good point, i've moved it after. In reality the Reset function (in the SDK) does nothing to modify the event/subrun/run numbers.

larpandoracontent/LArMetrics/EventValidationTool.cc
imawby: done, added a boolean to the MCParticle (only neutrino is set, non-neutrino particles always receive false). I have also added a function to LArMCParticleHelper that can be used to retrieve isCC for non-neutrino particles.

larpandoracontent/LArMetrics/ValidationAlgorithm.cc
imawby: The reco criteria is set as this: 

`const LArHierarchyHelper::MCHierarchy::ReconstructabilityCriteria recoCriteria(1, 1, 1, m_removeRecoNeutrons);`

So then I pick up all reconstructed particles even if they're not deemed to be reconstructable. I need to do this to make the hierarchy metrics correct i.e. I rebuild the true hierarchy out of 'visible' and 'not visible but has been reco'd' particles . But it does mean I have to externally add an 'is reconstructable' check as I cant use this because it uses the member variables of the reco criteria. If I understood that correctly?

AndyChappell: I'm not sure I follow. Are you using different reconstructability criteria to those used when constructing the Hierarchy in HierarchyHelper? In case we're talking about two different things, my main point was that I think you could just call pMCNode->IsReconstructable() - unless your thresholds are different, but then i'm not sure I understand why they are different.

larpandoracontent/LArMetrics/PFPValidationTool.cc
imawby: ah, its an impossible angle.. well, its not in the -pi->pi range

AndyChappell: Ah, I see. I wonder if it's preferable, where possible, to just have a consistent INVALID_VALUE set at -999, or -9999, or whatever works?

larpandoracontent/LArMetrics/PFPValidationTool.cc
imawby: that is a fair point

larpandoracontent/LArMetrics/ShowerValidationTool.cc
imawby: this is a killer spot, my local version of larpandora altered the vertex for photons. I got rid of that for this PR and modified larpandoracontent to use the endpoint instead, but looks like I forgot this one. Thank you!

larpandoracontent/LArMetrics/TrackValidationTool.cc
imawby: I think it was just for my personal interest, turns out michel track shower classification is about 50/50

larpandoracontent/LArMetrics/EventValidationTool.cc
imawby: done, i _think_ i've found all of the uses of -9999 etc..

larpandoracontent/LArMetrics/ValidationAlgorithm.cc
imawby: Nope, so in the ValidationAlg [here](https://github.com/imawby/LArContent/blob/feature/Metrics/larpandoracontent/LArMetrics/ValidationAlgorithm.cc#L75) I set the reco criteria to be 'does the MCParticle had one hit' so this means that I perform matching to all MCParticles (with one hit). Then this [line](https://github.com/imawby/LArContent/blob/feature/Metrics/larpandoracontent/LArMetrics/ValidationAlgorithm.cc#L108) means that my list of target MCParticles are those that match some higher reco criteria (which is what this function applies) OR are those that have been reco'd anyway despite not being deemed reconstructable. 

I think I cant use the function that you point me too because i think it would apply the reconstructable criteria that was set [here](https://github.com/imawby/LArContent/blob/feature/Metrics/larpandoracontent/LArMetrics/ValidationAlgorithm.cc#L75) i.e. of having 1 hit.

So then why do i care about picking up particles that have been reco'd even if they werent 'reconstructable'? it's because i need them to rebuild the 'true hierarchy'. In the hierarchy tool i rebuild the MCParticle hierarchy by considering the 'reconstructable' and 'has been reco'd anyway' particles instead of all generator particles. 

does that make sense?

larpandoracontent/LArMetrics/PFPValidationTool.cc
imawby: according to my thesis it is ' the angle from the z-axis to the projection of the momentum vector onto the x − z plane'. I chose it to match the experiments def of thetaxz at the time

larpandoracontent/LArMetrics/ShowerValidationTool.cc
imawby: yeah, so its meant to be 'what length covers most of the showers energy deposition' to get a decent estimate of the shower length which reco should hopefully match.. i.e. the length from the true hits would end up being super long if you count those hits very downstream of the main shower - and reco is unlikely to pick those up

larpandoracontent/LArMetrics/ShowerValidationTool.cc
imawby: for photons we defo should start at the point of first energy deposition, i've changed that now

larpandoracontent/LArMetrics/TrackValidationTool.cc
imawby: not sure i understand what the question is here.. l could potentially be negative if the hit is upstream of the true vertex.. so here i am checking whether the hit is between the true vertex and the initial region endpoint

larpandoracontent/LArMetrics/BaseValidationTool.h
imawby: i'm not sure if i've done this correctly..

imawby: i.e. with the use of [out] as opposed to [in, out]

AndyChappell: Yes, I would say that's right, here (this function purely writing to this variable, rather than also reading from it).

larpandoracontent/LArMetrics/TrackValidationTool.cc
AndyChappell: Yes, so this is essentially a question about the behaviour you want in the case where the projection was on the vertex or endRegionL.

larpandoracontent/LArMetrics/ValidationAlgorithm.cc
AndyChappell: Yes, thanks. And I agree, you can't call the function directly in that case.

larpandoracontent/LArMetrics/BaseValidationTool.h
imawby: ah, okay I see the distinction. Thanks!

larpandoracontent/LArHelpers/LArMCParticleHelper.h
```diff
@@ -129,6 +129,15 @@ class LArMCParticleHelper
      */
     static bool IsNeutrino(const pandora::MCParticle *const pMCParticle);
 
+    /**
+     *  @brief  Whether a mc particle is part of a CC interaction
+     *
+     *  @param[in]  pMCParticle the input mc particle
```
AndyChappell, line 135: No need to change this time around, but in general, if it's just an input parameter, I wouldn't bother with the annotation, I think it's only "interesting" for in/out and out.

larpandoracontent/LArMetrics/BaseValidationTool.h
```diff
@@ -22,23 +22,35 @@ namespace lar_content
...
+     *  @param[out]  totalEnergy the total charge of the filtered hits
      */
     void GetHitsOfType(const pandora::CaloHitList &inputList, const pandora::HitType hitType, pandora::CaloHitVector &outputVector, 
         float &totalEnergy);
+
+    float m_maxMichelSep;      ///< the maximum separation between a michel vertex and parent muon endpoint    
+    float m_invalidSmallFloat; ///< small float value used for failure case
+    float m_invalidLargeFloat; ///< large float value used for failure case    
+    int m_invalidInt;          ///< int value used for failure case    
+    float m_invalidAngle;      ///< angle value used for failure case
```
AndyChappell, line 53: No need to fix, but my broader point/query here was a) non-magic numbers, which is now fine, but also b) do we really need/want different numbers (other than for different types?), e.g. could the invalid float not simply be represented by -999, and rather than having a configurable just have a const float INVALID=-999.f to represent all floating point cases.

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -151,29 +160,33 @@ class LArMCParticle : public object_creation::MCParticle::Object
...
 
     /**
-     *  @brief  Get the x positions
+     *  @brief  Get the x-coordinates of the MC trajectory
+     *          note: trajectories are sparsified in the current sim workflow
      *
-     *  @return the x positions
+     *  @return the x-coordinates of the MC trajectory
      */
     pandora::FloatVector GetTrajPointsX() const;
```
AndyChappell: I wonder if we should package this up such that the traj points are represented as a CartesianPointVector. Maybe more natural? It can still be stored in files etc as separate float vectors, but for actual usage in algorithms, we should maybe map those vectors to/from a the CartesianPointVector.

larpandoracontent/LArObjects/LArMCParticle.h
AndyChappell: As above, maybe one CartesianPointVector is cleaner here?

AndyChappell (changes requested):
Thanks Isobel. The latest updates look good to me. I think I'd request one more, I think fairly minor change. Which is about the actual packaging of the traj points. Otherwise, my comments are just observations for future reference - whether you choose to make changes alongside the requested change is up to you.

larpandoracontent/LArHelpers/LArMCParticleHelper.h
imawby: gotcha

larpandoracontent/LArMetrics/BaseValidationTool.h
imawby: made a note of this

larpandoracontent/LArObjects/LArMCParticle.h
imawby: done

larpandoracontent/LArObjects/LArMCParticle.h
```diff
@@ -160,39 +158,21 @@ class LArMCParticle : public object_creation::MCParticle::Object
...
-    pandora::FloatVector GetTrajPointsY() const;
-
-    /**
-     *  @brief  Get the z-coordinates of the MC trajectory
-     *          note: trajectories are sparsified in the current sim workflow
-     *
-     *  @return the z-coordinates of the MC trajectory
-     */    
-    pandora::FloatVector GetTrajPointsZ() const;
+    pandora::CartesianPointVector GetTrajPoints() const;
```
AndyChappell: Final request - can we get this to return a const reference?

i.e. `const pandora::CartesianPointVector &GetTrajPoints() const;`

In principle there could be a lot of trajectory points, so I think we want to avoid copying behaviour here (the compiler will probably optimise this away, but we should enforce it).

AndyChappell (changes requested):
Hi Isobel, I think this should be the last change, and then this can be approved.

AndyChappell (approved):

AndyChappell (approved):

PR #280, opened 2026-04-28
3D cluster association algorithm

Mandri97:
An algorithm to associate clusters in 3D. It is a simple approach inspired by the (2D) longitudinal association algorithm, but it removes the reliance and constraints on pseudo-layer to better support 3D reconstruction


CrossR:
Not checked over yet in depth...but this looks very nicely self contained. Could it be split into a LArContent `ClusterHelper` update and LArRecoND new algorithm?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.h
```diff
@@ -0,0 +1,74 @@
...
+        float length;
+        pandora::CartesianVector innerCoordinate = pandora::CartesianVector(0.f, 0.f, 0.f);
+        pandora::CartesianVector outerCoordinate = pandora::CartesianVector(0.f, 0.f, 0.f);
+        pandora::CartesianVector centroid = pandora::CartesianVector(0.f, 0.f, 0.f);
+        LArPcaHelper::EigenValues eigenValues = LArPcaHelper::EigenValues(0.f, 0.f, 0.f);
+        LArPcaHelper::EigenVectors eigenVectors;
+    };
+
+    typedef std::unordered_map<const pandora::Cluster*, ClusterAttr> ClusterAttrMap;
+    mutable ClusterAttrMap clusterAttrMap;
```
lhwhitehead: Add `m_` prefix for class member variables

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+{
+    if (!pClusterList->empty())
+        m_view = LArClusterHelper::GetClusterHitType(pClusterList->front());
+
+    clusterAttrMap.clear();
+
+    for (ClusterList::const_iterator iter = pClusterList->begin(), iterEnd = pClusterList->end(); iter != iterEnd; ++iter)
+    {
+        const Cluster *const pCluster = *iter;
+        float clusterLength = LArClusterHelper::GetLength(pCluster);
```
lhwhitehead: Can be `const`?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool ThreeDAssociationAlgorithm::IsExtremalCluster(const bool isForward, const Cluster *const pCurrentCluster, const Cluster *const pTestCluster) const
+{
+    CartesianVector currentClusterInnerCoordinate{ clusterAttrMap.at(pCurrentCluster).innerCoordinate };
+    CartesianVector currentClusterOuterCoordinate{ clusterAttrMap.at(pCurrentCluster).outerCoordinate };
+
+    CartesianVector testClusterInnerCoordinate{ clusterAttrMap.at(pTestCluster).innerCoordinate };
+    CartesianVector testClusterOuterCoordinate{ clusterAttrMap.at(pTestCluster).outerCoordinate };
```
lhwhitehead: All `const`?

lhwhitehead (commented):
Only minor things from my side. Once you are done with everything, run clang format to polish up things into the style that we use.

larpandoracontent/LArHelpers/LArClusterHelper.cc
```diff
@@ -774,13 +774,29 @@ bool LArClusterHelper::SortByPosition(const Cluster *const pLhs, const Cluster *
...
+//------------------------------------------------------------------------------------------------------------------------------------------
+
 bool LArClusterHelper::SortByPulseHeight(const Cluster *const pLhs, const Cluster *const pRhs)
 {
     return (pLhs->GetHadronicEnergy() > pRhs->GetHadronicEnergy());
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+
```
AndyChappell: Remove excess whitespace.

larpandoracontent/LArHelpers/LArClusterHelper.cc
```diff
@@ -774,13 +774,29 @@ bool LArClusterHelper::SortByPosition(const Cluster *const pLhs, const Cluster *
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
+bool LArClusterHelper::SortByInnerPosition(const Cluster *const pLhs, const Cluster *const pRhs)
+{
+    CartesianVector lhsInnerCoordinate(0.f, 0.f, 0.f); 
+    CartesianVector rhsInnerCoordinate(0.f, 0.f, 0.f);
+    CartesianVector lhsOuterCoordinate(0.f, 0.f, 0.f); 
+    CartesianVector rhsOuterCoordinate(0.f, 0.f, 0.f);
```
AndyChappell: To avoid the extra instantiation, given you don't actually care about these values you could just define

`CartesianVector unused(0.f, 0.f, 0.f);`

and pass that as the second parameter to both of the extremal coordinate functions below.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.h
```diff
@@ -0,0 +1,74 @@
...
+    mutable pandora::HitType m_view;     ///< The view to which the hits under consideration belong
+
+    // struct to cache PCA attributes
+    struct ClusterAttr {
+        float length;
+        pandora::CartesianVector innerCoordinate = pandora::CartesianVector(0.f, 0.f, 0.f);
+        pandora::CartesianVector outerCoordinate = pandora::CartesianVector(0.f, 0.f, 0.f);
+        pandora::CartesianVector centroid = pandora::CartesianVector(0.f, 0.f, 0.f);
+        LArPcaHelper::EigenValues eigenValues = LArPcaHelper::EigenValues(0.f, 0.f, 0.f);
+        LArPcaHelper::EigenVectors eigenVectors;
```
AndyChappell: Coding style in Pandora is to add a prefix m_ to member variables.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
AndyChappell: In general (here and elsewhere), when initialising variables, prefer

`type varname{value};` over `type varname = value;`

unless there's a specific reason not to. This has some benefits in terms of avoiding unintended type narrowing, etc. The main exception here would be if you are explicitly calling an object constructor that might otherwise be confused for an initialisation list in the case of {}. The classic here would be a std::vector

`std::vector<int> v{10, 20};`   Gets you a vector containing the elements 10 and 20
`std::vector<int> v(10, 20);`   Gets you a vector containing 10 elements, all of which are 20

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+
+    for (ClusterList::const_iterator iter = pClusterList->begin(), iterEnd = pClusterList->end(); iter != iterEnd; ++iter)
+    {
+        const Cluster *const pCluster = *iter;
+        float clusterLength = LArClusterHelper::GetLength(pCluster);
+
+        if( clusterLength < m_minClusterLength )
+            continue;
+
+        this->PCAFit(pCluster);
```
AndyChappell: I wonder if this would be better named PopulatePCAAttributes, or PopulateFitAttributes?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
AndyChappell: Probably want to check that these clusters exist in the attribute map before attempting to access them - presumably simply returning false in the case they are not in the map would be the desirable behaviour here?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+
+    const CartesianVector currentPosition{ isForward ? currentClusterOuterCoordinate : currentClusterInnerCoordinate };
+    const CartesianVector testPosition{ isForward ? testClusterOuterCoordinate : testClusterInnerCoordinate };
+
+    if( isForward && ((LArClusterHelper::SortCoordinatesByPosition(currentPosition, testPosition) || 
+                      ((currentPosition == testPosition) && LArClusterHelper::SortByNHits(pTestCluster, pCurrentCluster)))))
+        return true;
+
+    if( !isForward && ((LArClusterHelper::SortCoordinatesByPosition(testPosition, currentPosition) || 
+                      ((currentPosition == testPosition) && LArClusterHelper::SortByNHits(pTestCluster, pCurrentCluster)))))
```
AndyChappell, line 115: Is this final expression correct?

Would you not want `LArClusterHelper::SortByNHits(pCurrentCluster, pTestCluster)`?

AndyChappell, line 115: I can see why you've used the SortBy functions here, but I think it would at least be worthwhile to add a comment about the intended behaviour and what these functions are doing, as it's not immediately obvious (to me at least) without looking at the code, what the behaviour is.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+        return true;
+
+    return false;
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+bool ThreeDAssociationAlgorithm::AreClustersAssociated(const Cluster *const pInnerCluster, const Cluster *const pOuterCluster) const
+{
+
```
AndyChappell, line 127: Remove excess whitespace

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+    const float closestDistance = LArClusterHelper::GetClosestDistance(pInnerCluster, pOuterCluster);
+
+    if( closestDistance > m_maxGapDistanceSquared )
+        return false;
+
+    const CartesianVector innerClusterInnerCoordinate{ clusterAttrMap.at(pInnerCluster).innerCoordinate };
+    const CartesianVector innerClusterOuterCoordinate{ clusterAttrMap.at(pInnerCluster).outerCoordinate };
+    const CartesianVector outerClusterInnerCoordinate{ clusterAttrMap.at(pOuterCluster).innerCoordinate };
+    const CartesianVector outerClusterOuterCoordinate{ clusterAttrMap.at(pOuterCluster).outerCoordinate };
```
AndyChappell: As above - worth checking the clusters are in the attribute map before trying to access them. Also suggest making these reference variables.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
AndyChappell: And make them reference variables

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+
+    if( !LArClusterHelper::SortCoordinatesByPosition(innerClusterInnerCoordinate, outerClusterInnerCoordinate) )
+        throw pandora::StatusCodeException(STATUS_CODE_NOT_ALLOWED);
```
AndyChappell, line 138: Open question - is this a failure case where we definitely want an exception because something has gone fundamentally wrong, or does it represent a case where you could simply return false?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+    const CartesianVector outerPrimaryAxis  { outerVectors.at(0).GetZ() > 0.f ? outerVectors.at(0) : outerVectors.at(0) * -1.f };
+    const CartesianVector outerSecondaryAxis{ outerVectors.at(1).GetZ() > 0.f ? outerVectors.at(1) : outerVectors.at(1) * -1.f };
+    const CartesianVector outerTertiaryAxis { outerVectors.at(2).GetZ() > 0.f ? outerVectors.at(2) : outerVectors.at(2) * -1.f };
+
+    const CartesianVector innerCentroid{ clusterAttrMap.at(pInnerCluster).centroid };
+    const CartesianVector outerCentroid{ clusterAttrMap.at(pOuterCluster).centroid };
+
+    const float openingAngle{ innerPrimaryAxis.GetCosOpeningAngle(outerPrimaryAxis) };
+
+    constexpr float shortClusterLength = 2.50f;
```
AndyChappell: Does this really need to be a constexpr? Would we not want this to be a configurable, if readonly, parameter read in from XML?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+
+    // Bypass opening angle cut if a short cluster is the extension of a longer one
+    if( (closestDistance > 0.5f) ||
```
AndyChappell: Should 0.5f be a configurable parameter?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+    const CartesianVector outerTertiaryAxis { outerVectors.at(2).GetZ() > 0.f ? outerVectors.at(2) : outerVectors.at(2) * -1.f };
+
+    const CartesianVector innerCentroid{ clusterAttrMap.at(pInnerCluster).centroid };
+    const CartesianVector outerCentroid{ clusterAttrMap.at(pOuterCluster).centroid };
+
+    const float openingAngle{ innerPrimaryAxis.GetCosOpeningAngle(outerPrimaryAxis) };
+
+    constexpr float shortClusterLength = 2.50f;
+
+    // Bypass opening angle cut if a short cluster is the extension of a longer one
```
AndyChappell: Would this not also bypass if the long cluster is an extension of the short one? Is that what you want?

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+        (clusterAttrMap.at(pInnerCluster).length > clusterAttrMap.at(pOuterCluster).length) };
+    
+    const CartesianVector centroid{ isInnerLongest ? innerCentroid : outerCentroid };
+
+    const CartesianVector primaryAxis  { isInnerLongest ? innerPrimaryAxis   : outerPrimaryAxis   };
+    const CartesianVector secondaryAxis{ isInnerLongest ? innerSecondaryAxis : outerSecondaryAxis };
+    const CartesianVector tertiaryAxis { isInnerLongest ? innerTertiaryAxis  : outerTertiaryAxis  };
+
+    const CartesianVector innerEndPrimaryAxis  { centroid + primaryAxis * primaryAxis.GetDotProduct(innerClusterOuterCoordinate - centroid) };
+    const CartesianVector outerStartPrimaryAxis{ centroid + primaryAxis * primaryAxis.GetDotProduct(outerClusterInnerCoordinate - centroid) };
```
AndyChappell, line 174: I can't quite get my head around what this is trying to do

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -0,0 +1,221 @@
...
+        ((clusterAttrMap.at(pInnerCluster).length <= shortClusterLength ) == (clusterAttrMap.at(pOuterCluster).length <= shortClusterLength))){
+
+        if( openingAngle < m_minCosRelativeAngle )
+            return false;
+    }
+
+    bool isInnerLongest{ LArClusterHelper::SortByNHits(pInnerCluster, pOuterCluster) && 
+        (clusterAttrMap.at(pInnerCluster).length > clusterAttrMap.at(pOuterCluster).length) };
+    
+    const CartesianVector centroid{ isInnerLongest ? innerCentroid : outerCentroid };
```
AndyChappell: I think all of these CartesianVectors can be reference variables.

larpandoracontent/LArContent.cc
```diff
@@ -289,6 +290,7 @@
     d("LArUnattachedDeltaRays",                 UnattachedDeltaRaysAlgorithm)                                                   \
     d("LArThreeDHitCreation",                   ThreeDHitCreationAlgorithm)                                                     \
     d("LArThreeDLongitudinalTracks",            ThreeViewLongitudinalTracksAlgorithm)                                           \
+    d("LArThreeDAssociation",                   ThreeDAssociationAlgorithm)                                                     \
```
AndyChappell: The header files will be reordered by clang format, so could you swap these two here? (clang format is not allowed to alter these preprocessor blocks).

AndyChappell (changes requested):
Hi Manoa, Thanks for the PR. I think this is generally looking pretty good. I have a few, mostly minor requests and a couple of queries.

On some more general points, as Leigh noted, we'll ultimately pass this through clang format to get the Pandora style. You can skip that step if you want, as for LArContent PRs, I will apply the clang format is a normal matter of constructing release branches. However if you do choose to apply it yourself, can you ensure that it's isolated to a single commit and make it the last one in response to reviews (this makes it easier to distinguish substantive changes from formatting changes when reviewing updates. Thanks!

A secondary comment - and not anything that I think needs addressing for this PR, but I wonder if some of the sliding fit algorithms might provide some more flexibility versus PCA fits for 3D association?

AndyChappell:
> Could it be split into a LArContent `ClusterHelper` update and LArRecoND new algorithm?

Just to comment on this latter point - a clusterhelper style refactoring may have some value here - I'll let @CrossR comment on that, but to the more general point of LArRecoND versus LArContent, modulo a couple of points where some hard-coded values could do with becoming configurable, I think this is a case where the algorithm is sufficiently generic that it belongs in LArContent. It seems to be a general purpose association algorithm based on 3D hits, which, those hard-coded values aside, doesn't appear to have anything that is specific to the ND, and could equally apply to the built 3D clusters in wire/strip-based TPCS. Happy to hear thoughts on this.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.h
Mandri97: Fixed

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Fixed

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Fixed both

larpandoracontent/LArHelpers/LArClusterHelper.cc
Mandri97: Good idea. It's fixed

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Renamed `PCAFit` to `PopulateFitAttributes`

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Good point. I modified it to retrieve the extremal coordinates with `LArClusterHelper::GetExtremalCoordinates` when the cluster is not available in the attribute map

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: When the current and test clusters are at the same position, we want to choose the one with more hits. `LArClusterHelper::SortByNHits` is a descending comparator, so the ordering is correct.

https://github.com/PandoraPFA/LArContent/blob/bd0903d3253508cb86b9a6899eb651e516ee0c6b/larpandoracontent/LArHelpers/LArClusterHelper.cc#L708-L717

Mandri97: Fair point. I've added this comment to clarity

```c++
// Find the extremal cluster by comparing the outermost coordinates (forward) or innermost coordinates (backward)
// Pick the cluster with more hits if coordinates are the same
```

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Good point. `AreClustersAssociated` is intended to be called after `GetListOfCleanClusters` so that the cluster attribute map is populated. I've added a guard that throws `StatusCodeException(STATUS_CODE_NOT_INITIALIZED)` when a cluster is missing in the attribute map.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: `GetListOfCleanClusters` sorts clusters via `SortByInnerPosition` (calling `SortCoordinatesByPosition`), so if `innerClusterInnerCoordinate` > `outerClusterInnerCoordinate` then either `GetListOfCleanClusters` hasn't been called or something has gone wrong.

```c++
bool LArClusterHelper::SortByInnerPosition(const Cluster *const pLhs, const Cluster *const pRhs)
{
    CartesianVector lhsInnerCoordinate(0.f, 0.f, 0.f); 
    CartesianVector rhsInnerCoordinate(0.f, 0.f, 0.f);
    CartesianVector lhsOuterCoordinate(0.f, 0.f, 0.f); 
    CartesianVector rhsOuterCoordinate(0.f, 0.f, 0.f);

    LArClusterHelper::GetExtremalCoordinates(pLhs, lhsInnerCoordinate, lhsOuterCoordinate);
    LArClusterHelper::GetExtremalCoordinates(pRhs, rhsInnerCoordinate, rhsOuterCoordinate);

    return LArClusterHelper::SortCoordinatesByPosition(lhsInnerCoordinate, rhsInnerCoordinate);
}
```

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Changed it to tunable parameter.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Fair. Updated the comment for clarity
```c++
// Bypass opening angle check if a short cluster extends a long one (or vice versa)
```

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: In ND-LAr, 0.5f corresponds to a 4 mm pixel pitch with a 1 mm tolerance. Updated to
```c++
const double channelPitch{ 1.3 * LArGeometryHelper::GetWirePitch(this->GetPandora(), m_view) };
```
to make it detector agnostic.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Fixed

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: These lines determine the coordinates of the inner cluster's end and the outer cluster's start along the longest cluster's primary axis, relative to the longest cluster's centroid.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.h
Mandri97: Fixed

larpandoracontent/LArContent.cc
Mandri97: Fixed

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Fixed

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -88,15 +89,30 @@ void ThreeDAssociationAlgorithm::PopulateClusterAssociationMap(const ClusterVect
...
-
-    CartesianVector testClusterInnerCoordinate{ clusterAttrMap.at(pTestCluster).innerCoordinate };
-    CartesianVector testClusterOuterCoordinate{ clusterAttrMap.at(pTestCluster).outerCoordinate };
+    CartesianVector currentClusterInnerCoordinate(0.f, 0.f, 0.f);
+    CartesianVector currentClusterOuterCoordinate(0.f, 0.f, 0.f);
+    CartesianVector testClusterInnerCoordinate(0.f, 0.f, 0.f);
+    CartesianVector testClusterOuterCoordinate(0.f, 0.f, 0.f);
+
+    if( m_clusterAttrMap.find(pCurrentCluster) == m_clusterAttrMap.end() )
+        LArClusterHelper::GetExtremalCoordinates(pCurrentCluster, currentClusterInnerCoordinate, currentClusterOuterCoordinate);
```
AndyChappell, line 87: If any part of the conditional is multiline, enclose each branch in braces, even if that branch is only a single line

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -88,15 +89,30 @@ void ThreeDAssociationAlgorithm::PopulateClusterAssociationMap(const ClusterVect
...
+    else{
```
AndyChappell: No need to address this here, as the pre-release clang format run will resolve it, but for future reference, Pandora style is to have the opening brace on its own line.

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -88,15 +89,30 @@ void ThreeDAssociationAlgorithm::PopulateClusterAssociationMap(const ClusterVect
...
+        currentClusterInnerCoordinate = m_clusterAttrMap.at(pCurrentCluster).m_innerCoordinate;
+        currentClusterOuterCoordinate = m_clusterAttrMap.at(pCurrentCluster).m_outerCoordinate;
+    }
+    
+    if ( m_clusterAttrMap.find(pTestCluster) == m_clusterAttrMap.end() )
+        LArClusterHelper::GetExtremalCoordinates(pTestCluster, testClusterInnerCoordinate, testClusterOuterCoordinate);
```
AndyChappell, line 97: As above re brace enclosure

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
```diff
@@ -137,29 +154,25 @@ bool ThreeDAssociationAlgorithm::AreClustersAssociated(const Cluster *const pInn
     const CartesianVector outerSecondaryAxis{ outerVectors.at(1).GetZ() > 0.f ? outerVectors.at(1) : outerVectors.at(1) * -1.f };
     const CartesianVector outerTertiaryAxis { outerVectors.at(2).GetZ() > 0.f ? outerVectors.at(2) : outerVectors.at(2) * -1.f };
 
-    const CartesianVector innerCentroid{ clusterAttrMap.at(pInnerCluster).centroid };
-    const CartesianVector outerCentroid{ clusterAttrMap.at(pOuterCluster).centroid };
-
     const float openingAngle{ innerPrimaryAxis.GetCosOpeningAngle(outerPrimaryAxis) };
+    const double channelPitch{ 1.3 * LArGeometryHelper::GetWirePitch(this->GetPandora(), m_view) };
```
AndyChappell: Should 1.3 be a configurable parameter here?

AndyChappell (changes requested):
A couple of minor additional requests, and then I think this should be good to go (though looks like there will be merge conflicts to resolve).

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Fixed them all

larpandoracontent/LArThreeDReco/LArLongitudinalTrackMatching/ThreeDAssociationAlgorithm.cc
Mandri97: Yes! Fixed.

PR #281, opened 2026-05-06
Make minimum rL for the CR stitching configurable

AlexWilkinsonnn:
Simple change to expose the minimum longitudinal impact parameter in the YZ plane to the setting . In sim, negative rL should only happen due to diffusion and 2D wire response so hardcoded -1 was plenty... in data things can and do get weird though :grin: 

Im not sure where this PR should go, the parent branch was develop of the larsoft fork which has commits not yet in the v5 branch here.


AndyChappell (changes requested):
Hi @AlexWilkinsonnn, the logic of the change looks fine, but it doesn't look like anything has been added to ReadSettings to actually allow the configuration override.

AlexWilkinsonnn:
Woops, thanks for spotting, I update the ReadSettongs

AndyChappell (approved):

PR #286, opened 2026-06-30
Now impossible to throw exception if MC start/end momentum is zero

imawby:
Hello,

When running on 6K files, there was one event where the momentum magnitude was > epsilon, but the calculation of the opening angle failed because of a (momentum magnitude * momentum magnitude) denominator in which both momentum magnitudes were very small :'( 

I've now protected all instances of this via try/catch blocks. It is very rare that the code flow should enter the catch block.

This should be the end of this issue.

Thanks!

### DIFF
--- a/larpandoracontent/LArMetrics/PFPValidationTool.cc
+++ b/larpandoracontent/LArMetrics/PFPValidationTool.cc
@@ -174,13 +174,20 @@ void PFPValidationTool::GetRecoParticleInfo(const MCParticle *const pMCTarget, c
         pfpTreeVars.m_recoVertexZ.push_back(pRecoVertex->GetPosition().GetZ());
 
         // Signed vertexAcc
-        const CartesianVector &trueVertex((pMCTarget->GetParticleId() == 22) ? pMCTarget->GetEndpoint() : pMCTarget->GetVertex());
-        const float vertexAcc((pRecoVertex->GetPosition() - trueVertex).GetMagnitude());
-        const float sign(
-            (vertexAcc < std::numeric_limits<float>::epsilon() || pMCTarget->GetMomentum().GetMagnitude() < std::numeric_limits<float>::epsilon()) ? 1.f
+        float vertexAcc(m_invalidLargeFloat);
+        try
+        {
+            const CartesianVector &trueVertex((pMCTarget->GetParticleId() == PHOTON) ? pMCTarget->GetEndpoint() : pMCTarget->GetVertex());
+            vertexAcc = (pRecoVertex->GetPosition() - trueVertex).GetMagnitude();
+            const float sign(
+                (vertexAcc < std::numeric_limits<float>::epsilon() || pMCTarget->GetMomentum().GetMagnitude() < std::numeric_limits<float>::epsilon()) ? 1.f
                 : (pRecoVertex->GetPosition() - trueVertex).GetOpeningAngle(pMCTarget->GetMomentum()) < (M_PI * 0.5) ? 1.f
-                                                                                                                     : -1.f);
-        pfpTreeVars.m_vertexAcc.push_back(vertexAcc * sign);
+                : -1.f);
+            vertexAcc *= sign;
+        }
+        catch (StatusCodeException &) { vertexAcc = m_invalidLargeFloat; }
+
+        pfpTreeVars.m_vertexAcc.push_back(vertexAcc);
 
         try
         {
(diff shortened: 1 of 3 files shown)


larpandoracontent/LArMetrics/PFPValidationTool.cc
```diff
@@ -174,13 +174,20 @@ void PFPValidationTool::GetRecoParticleInfo(const MCParticle *const pMCTarget, c
...
+                (vertexAcc < std::numeric_limits<float>::epsilon() || pMCTarget->GetMomentum().GetMagnitude() < std::numeric_limits<float>::epsilon()) ? 1.f
                 : (pRecoVertex->GetPosition() - trueVertex).GetOpeningAngle(pMCTarget->GetMomentum()) < (M_PI * 0.5) ? 1.f
-                                                                                                                     : -1.f);
-        pfpTreeVars.m_vertexAcc.push_back(vertexAcc * sign);
+                : -1.f);
+            vertexAcc *= sign;
+        }
+        catch (StatusCodeException &) { vertexAcc = m_invalidLargeFloat; }
+
+        pfpTreeVars.m_vertexAcc.push_back(vertexAcc);
```
AndyChappell: I don't think we should use try/catch blocks here, given we know the value that leads to an exception, if we're going to check a value, we should veto it here (the control flow is not really exceptional). I think I would prefer something like

```
const CartesianVector &trueVertex((pMCTarget->GetParticleId() == 22) ? pMCTarget->GetEndpoint() : pMCTarget->GetVertex());
const float vertexAcc((pRecoVertex->GetPosition() - trueVertex).GetMagnitude());
const float sqrtEpsilon(std::sqrt(std::numeric_limits<float>::epsilon()));
const float sign(
    (vertexAcc < sqrtEpsilon || pMCTarget->GetMomentum().GetMagnitude() < sqrtEpsilon) ? 1.f
        : (pRecoVertex->GetPosition() - trueVertex).GetOpeningAngle(pMCTarget->GetMomentum()) < (M_PI * 0.5) ? 1.f
        : -1.f);
pfpTreeVars.m_vertexAcc.push_back(vertexAcc * sign);
```

I think the code here could do with some more tidying for clarity, but I think conceptually this is what we want.

larpandoracontent/LArMetrics/ShowerValidationTool.cc
```diff
@@ -314,7 +314,10 @@ void ShowerValidationTool::GetRecoVertexInfo(const CartesianVector &recoShrVtx,
     if (pMC)
     {
         const CartesianVector trueDir(pMC->GetMomentum().GetUnitVector());
-        showerTreeVars.m_recoShrDirAcc.push_back(trueDir.GetOpeningAngle(recoShrDir));
+        float recoShrDirAcc(m_invalidLargeFloat);
+        try { recoShrDirAcc = trueDir.GetOpeningAngle(recoShrDir); } 
+        catch (StatusCodeException &) { recoShrDirAcc = m_invalidLargeFloat; }
+        showerTreeVars.m_recoShrDirAcc.push_back(recoShrDirAcc);
```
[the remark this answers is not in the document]
AndyChappell: As above, I think we want a check against sqrt(epsilon), rather than a try/catch block..

larpandoracontent/LArMetrics/TrackValidationTool.cc
```diff
@@ -165,28 +165,41 @@ void TrackValidationTool::GetRecoVertexAndEndpointVars(const MCParticle *const p
     const CartesianVector trueEndpoint(pLArMC->GetEndpoint());
 
     // Endpoint accuracy
-    float endpointAcc((recoEndpoint - trueEndpoint).GetMagnitude());
-    const float sign((endpointAcc < std::numeric_limits<float>::epsilon() || !doesMCHaveEndDir)       ? 1.f
-            : (recoEndpoint - trueEndpoint).GetOpeningAngle(pLArMC->GetEndDirection()) < (M_PI * 0.5) ? 1.f
-                                                                                                      : -1.f);
-    endpointAcc *= sign;
+    float endpointAcc(m_invalidLargeFloat);
```
AndyChappell, line 169: And again throughout this file. I would provisionally revert to the original implementation, adjust to use the sqrt of epsilon, and ideally also try to tidy the code to make it a bit more readable.

AndyChappell (changes requested):
Sorry, I thought I'd already submitted this one. I think here, the overall comment is that I think this should be handled conditionally, rather than exceptionally. We know the conditions under which an exception would be thrown going in, so given how heavy exception handling is, we should prefer to pre-empt it, rather than catch it.

PR #287, opened 2026-07-01
Making sure file version actually written and read

imawby:
Hello!

This PR makes sure that the 'global header' is written to the event file and is subsequently read. At the moment, the global header contains file version information. For example, the xml looks like:

```
<Header>
    <Version>
        <MajorVersion>2</MajorVersion>
        <MinorVersion>0</MinorVersion>
    </Version>
</Header>
<Event>
    <CaloHit>
        <LArTPCVolumeId>0</LArTPCVolumeId>
        <DaughterVolumeId>5</DaughterVolumeId>
        <CellGeometry>0</CellGeometry>
        <PositionVector>-114.319160461 0 251.014801025</PositionVector>
...
```

If the file major version == 2 the event information i.e. run:subrun:event will be written to the output pndrs/xmls. A flag (isEnhancedFile) is added to the event readers to inform the reader that this new block exists and that it should be read.

### DIFF
--- a/larpandoracontent/LArPersistency/EventReadingAlgorithm.cc
+++ b/larpandoracontent/LArPersistency/EventReadingAlgorithm.cc
@@ -110,6 +110,9 @@ void EventReadingAlgorithm::MoveToNextEventFile()
     m_eventFileNameVector.pop_back();
     PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->ReplaceEventFileReader(m_eventFileName));
 
+    if (m_isEnhancedEventFile)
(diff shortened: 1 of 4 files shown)


larpandoracontent/LArPersistency/EventReadingAlgorithm.cc
```diff
@@ -6,6 +6,8 @@
  *  $Log: $
  */
 
+#include "Api/PandoraApi.h"
```
AndyChappell: I think this is redundant now, right? This was for when you were considering the direct route for SetEventInformation?

AndyChappell (changes requested):
One minor change and then I think this looks good. Thanks, Isobel!

AndyChappell (approved):

PR #291, opened 2026-07-20
CR stitching in the worker instances

AlexWilkinsonnn:
Decoupling the stitching tools from the Master alg. The stitching tool took the master alg as input, generalised this is any algorithm that implements the 2 methods in `StitchingPfoOperations.h` interface. So any algorithm that implements this can invoke any stitching tool that implements `StitchingBaseTool.h`.

To run stitching tool from a worker instance, there is the `PfoStitchingAlgorithm`. This does some not great things to "infer" the LArTPC objects that the master just has from the enclosing LArTPC the worker has + the drift line gaps. The LArTPC a PFO is in is then inferred from the pfo 3D hit coordinates... this can all be refactored once we have detailed LArTPC info in the worker instances.

I also made some small changes to `StitchingCosmicRayMergingTool` so it can accept particles that may be in a neutrino hierarchy and to not expect to be called from the Master alg. 

All of this changes nothing by default.

### DIFF
--- a/cmake/LArContent_sources.cmake
+++ b/cmake/LArContent_sources.cmake
@@ -203,7 +203,7 @@ set(LAR_CONTENT_SRCS
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
     larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.cc
     larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
-    larpandoracontent/LArThreeDReco/LArPfoStitching/TrackPfoStitchingAlgorithm.cc
+    larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
     larpandoracontent/LArTrackShowerId/BranchGrowingAlgorithm.cc
     larpandoracontent/LArTrackShowerId/ClusterCharacterisationBaseAlgorithm.cc
     larpandoracontent/LArTrackShowerId/CutClusterCharacterisationAlgorithm.cc
--- a/cmake/LArContent_sources.cmake
+++ b/cmake/LArContent_sources.cmake
@@ -203,6 +203,7 @@ set(LAR_CONTENT_SRCS
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
     larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.cc
     larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
+    larpandoracontent/LArThreeDReco/LArPfoStitching/TrackPfoStitchingAlgorithm.cc
     larpandoracontent/LArTrackShowerId/BranchGrowingAlgorithm.cc
     larpandoracontent/LArTrackShowerId/ClusterCharacterisationBaseAlgorithm.cc
     larpandoracontent/LArTrackShowerId/CutClusterCharacterisationAlgorithm.cc
--- a/cmake/LArContent_sources.cmake
+++ b/cmake/LArContent_sources.cmake
@@ -32,7 +32,6 @@ set(LAR_CONTENT_SRCS
     larpandoracontent/LArControlFlow/PreProcessingAlgorithm.cc
     larpandoracontent/LArControlFlow/SimpleNeutrinoIdTool.cc
     larpandoracontent/LArControlFlow/SlicingAlgorithm.cc
-    larpandoracontent/LArControlFlow/StitchingCosmicRayMergingTool.cc
     larpandoracontent/LArControlFlow/StreamingAlgorithm.cc
     larpandoracontent/LArControlFlow/TestBeamCosmicRayTaggingTool.cc
     larpandoracontent/LArCustomParticles/CustomParticleCreationAlgorithm.cc
@@ -202,6 +201,7 @@ set(LAR_CONTENT_SRCS
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
+    larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
     larpandoracontent/LArTrackShowerId/BranchGrowingAlgorithm.cc
     larpandoracontent/LArTrackShowerId/ClusterCharacterisationBaseAlgorithm.cc
     larpandoracontent/LArTrackShowerId/CutClusterCharacterisationAlgorithm.cc
--- a/cmake/LArContent_sources.cmake
+++ b/cmake/LArContent_sources.cmake
@@ -201,6 +201,7 @@ set(LAR_CONTENT_SRCS
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
+    larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.cc
     larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
     larpandoracontent/LArTrackShowerId/BranchGrowingAlgorithm.cc
     larpandoracontent/LArTrackShowerId/ClusterCharacterisationBaseAlgorithm.cc
--- a/larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
+++ b/larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
@@ -41,8 +41,15 @@ StatusCode PfoStitchingAlgorithm::Run()
 {
     PANDORA_RETURN_IF(STATUS_CODE_SUCCESS, m_stitchingToolVector.empty());
 
-    const PfoList *pPfoList{nullptr};
-    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, PandoraContentApi::GetList(*this, m_inputPfoListName, pPfoList));
+    PfoList sortedPfos;
+    for (const std::string &listName : m_inputPfoListNames)
+    {
+        const PfoList *pPfos{nullptr};
+        if (PandoraContentApi::GetList(*this, listName, pPfos) == STATUS_CODE_SUCCESS)
+            sortedPfos.insert(sortedPfos.end(), pPfos->begin(), pPfos->end());
+    }
+    sortedPfos.sort(LArPfoHelper::SortByNHits);
+    const PfoList *const pSortedPfoList{&sortedPfos};
 
     InferredLArTPCVector inferredTPCs;
     PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->InferLArTPCs(inferredTPCs));
@@ -50,13 +57,13 @@ StatusCode PfoStitchingAlgorithm::Run()
         return STATUS_CODE_SUCCESS;
 
     PfoToLArTPCMap pfoToLArTPCMap;
-    this->GetPfoToLArTPCMap(pPfoList, inferredTPCs, pfoToLArTPCMap);
+    this->GetPfoToLArTPCMap(pSortedPfoList, inferredTPCs, pfoToLArTPCMap);
     if (pfoToLArTPCMap.empty())
         return STATUS_CODE_SUCCESS;
 
     PfoToFloatMap stitchedPfosToX0Map;
     for (StitchingBaseTool *const pStitchingTool : m_stitchingToolVector)
-        pStitchingTool->Run(this, pPfoList, pfoToLArTPCMap, stitchedPfosToX0Map);
+        pStitchingTool->Run(this, pSortedPfoList, pfoToLArTPCMap, stitchedPfosToX0Map);
 
     return STATUS_CODE_SUCCESS;
 }
@@ -183,7 +190,6 @@ const std::string PfoStitchingAlgorithm::GetListName(const Cluster *const pClust
             return listName;
         }
     }
-
     PANDORA_THROW(STATUS_CODE_NOT_FOUND);
 }
 
@@ -200,7 +206,22 @@ const std::string PfoStitchingAlgorithm::GetListName(const Vertex *const pVertex
             return listName;
         }
     }
+    PANDORA_THROW(STATUS_CODE_NOT_FOUND);
+}
 
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+const std::string PfoStitchingAlgorithm::GetInputListName(const ParticleFlowObject *const pPfo) const
+{
+    for (const std::string &listName : m_inputPfoListNames)
+    {
+        const PfoList *pPfoList{nullptr};
+        if ((PandoraContentApi::GetList(*this, listName, pPfoList) == STATUS_CODE_SUCCESS) &&
+            (pPfoList->end() != std::find(pPfoList->begin(), pPfoList->end(), pPfo)))
+        {
+            return listName;
+        }
+    }
     PANDORA_THROW(STATUS_CODE_NOT_FOUND);
 }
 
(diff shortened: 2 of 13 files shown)


larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.h
```diff
@@ -0,0 +1,55 @@
...
+#include "Pandora/AlgorithmTool.h"
+#include "Pandora/Algorithm.h"
+
+#include "larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingPfoOperations.h"
+
+namespace lar_content
+{
+
+typedef std::unordered_map<const pandora::ParticleFlowObject *, const pandora::LArTPC *> PfoToLArTPCMap;
+typedef std::unordered_map<const pandora::ParticleFlowObject *, float> PfoToFloatMap;
```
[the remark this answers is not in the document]
AlexWilkinsonnn: I dont think duplicated typedefs is an error.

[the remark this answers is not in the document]
AndyChappell: I agree it's not an error per se, but, given you include StitchingPfoOperations.h in the base tool, would it not still be preferable to only have the definition in PfoOps, that way if the definition ever changes, it only changes in one place?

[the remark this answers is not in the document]
AlexWilkinsonnn: I think for lightweight typedefs like this its nice to see them in the same header you're reading, especially since they are used in methods it overrides from the interface and in other master private methods. How about just moving them into the class namespace so its obvious which one you're using?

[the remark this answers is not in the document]
AndyChappell: Yes, on that basis, I think inside the class makes sense, to keep the definition local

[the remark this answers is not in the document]
AlexWilkinsonnn: I went with having the typedefs in the interface classes, pfo ops and stitching base.

cmake/LArContent_sources.cmake
```diff
@@ -202,6 +201,9 @@ set(LAR_CONTENT_SRCS
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewSimpleTracksTool.cc
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewThreeDKinkTool.cc
     larpandoracontent/LArThreeDReco/LArTwoViewMatching/TwoViewTransverseTracksAlgorithm.cc
+    larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.cc
```
AndyChappell, line 204: Unfortunately, cetmodules needs to know about folders, so it has its own special handling - if you only add new files, this is sufficient, but if you add a new folder, it needs to be included in the [cetmodules_build.cmake](https://github.com/PandoraPFA/LArContent/blob/267496c23eb7f05871c0625af08bd177b2af4ac5/cmake/cetmodules_build.cmake#L97) as well. I hope this will change in the future.

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.h
```diff
@@ -0,0 +1,126 @@
...
+    const std::string GetListName(const pandora::Vertex *const pVertex) const;
+
+    /**
+     *  @brief  Find the named list containing the input pfo
+     *
+     *  @param[in]  pPfo the input pfo whose list is required
+     *
+     *  @return the list name
+     */
+    const std::string GetInputListName(const pandora::ParticleFlowObject *const pPfo) const;
```
AndyChappell: Appreciate it's used for an input list, but presumably, in principle it doesn't matter and this could just be called GetListName for consistency with the cluster and vertex cases?

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
```diff
@@ -0,0 +1,322 @@
...
+
+        PANDORA_THROW_IF(STATUS_CODE_INVALID_PARAMETER, !pStitchingTool);
+
+        m_stitchingToolVector.push_back(pStitchingTool);
+    }
+
+    return STATUS_CODE_SUCCESS;
+}
+
+} // namespace lar_content
```
AndyChappell, line 313: Newline at end of file

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.cc
```diff
@@ -0,0 +1,30 @@
...
+    if (!pStitchingOperations)
+    {
+        std::cout << "StitchingBaseTool: Calling algorithm must implement StitchingPfoOperations" << std::endl;
+        PANDORA_THROW(STATUS_CODE_INVALID_PARAMETER);
+    }
+
+    this->RunStitching(pAlgorithm, pStitchingOperations, pMultiPfoList, pfoToLArTPCMap, stitchedPfosToX0Map);
+}
+
+} // namespace lar_content
```
AndyChappell, line 30: New line at end of file.

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingBaseTool.h
```diff
@@ -0,0 +1,56 @@
...
+     *  @param  stitchedPfosToX0Map a map of cosmic-ray pfos that have been stitched between lar tpcs to the X0 shift
+     */
+    virtual void RunStitching(const pandora::Algorithm *const pAlgorithm, const StitchingPfoOperations *const pStitchingOperations,
+        const pandora::PfoList *const pMultiPfoList, PfoToLArTPCMap &pfoToLArTPCMap, PfoToFloatMap &stitchedPfosToX0Map) = 0;
+
+};
+
+} // namespace lar_content
+
+#endif // #ifndef LAR_STITCHING_BASE_TOOL_H
```
AndyChappell, line 56: New line at end of file.

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
```diff
@@ -36,30 +37,34 @@ StitchingCosmicRayMergingTool::StitchingCosmicRayMergingTool() :
...
 
-    if (this->GetPandora().GetGeometry()->GetLArTPCMap().size() < 2)
-        return;
-
     if (pfoToLArTPCMap.empty())
         throw StatusCodeException(STATUS_CODE_NOT_FOUND);
 
-    PfoList primaryPfos;
-    this->SelectPrimaryPfos(pMultiPfoList, pfoToLArTPCMap, primaryPfos);
+    const unsigned int firstVolId{pfoToLArTPCMap.begin()->second->GetLArTPCVolumeId()};
```
AndyChappell, line 58: Obviously if there are no TPCs then something has gone really wrong, but technically, with the size check now omitted, this code could in principle dereference an invalid pointer, so we should probably still ensure that the TPC map is non-empty before calling this - it'll keep Coverity happy.

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
```diff
@@ -36,30 +37,34 @@ StitchingCosmicRayMergingTool::StitchingCosmicRayMergingTool() :
...
+    if (std::all_of(std::next(pfoToLArTPCMap.begin()), pfoToLArTPCMap.end(),
+        [&](const auto &iter){ return iter.second->GetLArTPCVolumeId() == firstVolId; }))
+        return;
```
AndyChappell, line 61: Am I missing some edge case here, or is this not effectively equivalent to the size() < 2 code that was deleted above?

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingPfoOperations.h
```diff
@@ -0,0 +1,52 @@
...
+     *  @param  pPfoToDelete the address of the pfo to delete (will become a dangling pointer)
+     *  @param  pfoToLArTPCMap the pfo to lar tpc map
+     */
+    virtual void StitchPfos(const pandora::ParticleFlowObject *const pPfoToEnlarge, const pandora::ParticleFlowObject *const pPfoToDelete,
+        PfoToLArTPCMap &pfoToLArTPCMap) const = 0;
+};
+
+} // namespace lar_content
+
+#endif // #ifndef LAR_STITCHING_PFO_OPERATIONS_H
```
AndyChappell, line 52: New line at end of file.

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
```diff
@@ -0,0 +1,322 @@
...
+    float inferredTPCMinX{parentMinX};
+    for (const std::pair<float, float> &interval : driftGapXIntervals)
+    {
+        inferredTPCXIntervals.emplace_back(inferredTPCMinX, interval.first);
+        inferredTPCMinX = interval.second;
+    }
+    inferredTPCXIntervals.emplace_back(inferredTPCMinX, parentMaxX);
+
+    unsigned int volumeId{0};
+    bool isDriftInPositiveX{true};
```
AndyChappell, line 108: Is this ok for all configurations? If you're CPA-APA-CPA this is fine, but what if you're APA-CPA-APA-CPA-APA, as in the full 10kt? Would that not give you the wrong initial drift direction?

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
```diff
@@ -0,0 +1,322 @@
...
+                }
+            }
+            if (!pPosLArTPC) // 3D hit ended up in a gap or outside detector, ignoring it
+                continue;
+
+            if (pLArTPC && pLArTPC != pPosLArTPC)
+            {
+                pLArTPC = nullptr;
+                break;
+            }
```
AndyChappell: This is identifying PFOs that already span the inferred volumes, and therefore can't need stitching? Might be worth a comment to clarify the intent here?

Related, if that's true, what happens in the case that a PFO spans > 2 inferred volumes (or rather, is reco'd across 2, but maybe not into a valid third volume) in the full 10kt? Seems like we'd give up here? (which is fine, but figured I'd raise it in case that wasn't the intent)

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
```diff
@@ -0,0 +1,322 @@
...
+    {
+        const ClusterList *pClusterList{nullptr};
+        if ((PandoraContentApi::GetList(*this, listName, pClusterList) == STATUS_CODE_SUCCESS) &&
+            (pClusterList->end() != std::find(pClusterList->begin(), pClusterList->end(), pCluster)))
+        {
+            return listName;
+        }
+    }
+    PANDORA_THROW(STATUS_CODE_NOT_FOUND);
+}
```
AndyChappell, line 189: Seems like you could template this for all three types.

AndyChappell (changes requested):
A few clarifications and some minor changes needed. This looks close to ready to me, pending discussions with the ND team.

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
```diff
@@ -0,0 +1,322 @@
...
+    std::vector<std::pair<float, float>> driftGapXIntervals;
+    for (const DetectorGap *const pDetectorGap : this->GetPandora().GetGeometry()->GetDetectorGapList())
+    {
+        const LineGap *const pLineGap(dynamic_cast<const LineGap *>(pDetectorGap));
+
+        if (!pLineGap || (pLineGap->GetLineGapType() != TPC_DRIFT_GAP))
+            continue;
+
+        PANDORA_RETURN_IF(STATUS_CODE_INVALID_PARAMETER, pLineGap->GetLineStartX() > pLineGap->GetLineEndX())
+
```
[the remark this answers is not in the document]
AlexWilkinsonnn: No compilation error but yes style-wise

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
```diff
@@ -0,0 +1,322 @@
...
+}
+
+//------------------------------------------------------------------------------------------------------------------------------------------
+
+StatusCode PfoStitchingAlgorithm::ReadSettings(const TiXmlHandle xmlHandle)
+{
+    PANDORA_RETURN_RESULT_IF(STATUS_CODE_SUCCESS, !=, XmlHelper::ReadVectorOfValues(xmlHandle, "InputPfoListNames", m_inputPfoListNames));
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=,
+        XmlHelper::ReadVectorOfValues(xmlHandle, "DaughterListNames", m_daughterListNames));
+
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Made daughter lists mandatory

cmake/LArContent_sources.cmake
AlexWilkinsonnn: Ah I didnt know that, added. Thanks.

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
AlexWilkinsonnn: Yeah agreed, I templated the the part that does work
```
template <typename TObject, typename TList>
const std::string PfoStitchingAlgorithm::GetListName(
    const TObject *const pObject,
    const std::vector<std::string> &listNames) const
{
    for (const std::string &listName : listNames)
    {
        const TList *pObjectList{nullptr};

        if ((PandoraContentApi::GetList(*this, listName, pObjectList) == STATUS_CODE_SUCCESS) &&
            (pObjectList->end() != std::find(pObjectList->begin(), pObjectList->end(), pObject)))
        {
            return listName;
        }
    }

    PANDORA_THROW(STATUS_CODE_NOT_FOUND);
}
```
Then the GetListNames become like
```
const std::string PfoStitchingAlgorithm::GetListName(const Vertex *const pVertex) const
{
    return this->GetListName<Vertex, VertexList>(pVertex, m_daughterListNames);
}
```
I think this is best way given some look in the daughterlists and some in the inputpfolists

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
AlexWilkinsonnn: Ok, I do check if this pfoToLArTPCMap is empty, I also added a check to return if the pandora instance geometry is empty.

larpandoracontent/LArThreeDReco/LArPfoStitching/StitchingCosmicRayMergingTool.cc
AlexWilkinsonnn: So the problem is that the deleted lartpc map check always fails for worker nodes where I beleive `this->GetPandora().GetGeometry()->GetLArTPCMap()` return a map of size one containing the super enclosing LArTPC as the individual drift volume LArTPCs the master alg has does not get copied to workers, instead a single enclosing LArTPC volume is constructed and only that is given to the workers (along with linegaps). So I replaced this check with this condition which is supposed to be a quick check of "is there anything that could possibly need stitching".

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
AlexWilkinsonnn: Yes that is correct. Although it doesnt matter for the inferred lartpcs, so long as they alternate the stitching helper methods will identify them as "stitchable" LArTPCs. I dont actually need these to encode the correct drift direction, just have them alternating. It's ugly and I will refactor all this once we have more LArTPC info in the workers.

larpandoracontent/LArThreeDReco/LArPfoStitching/PfoStitchingAlgorithm.cc
AlexWilkinsonnn: Yes I am looking for PFOs contained within a single LArTPC, this is what I consider a valid target for stitching. I left a comment.

Yes the case where a PFO spans > 2 TPCs buy may still require stitching is not covered. This should be, but I think that would require even more dev on the stitching helpers which assume each PFO was reconstructed in just one TPC. For now, I am going to ignore these cases, once we have updated LArTPC info, I will look into updating the stitching to handle particles spanning > 2 TPCs.

AndyChappell (approved):
Pending tests, this now looks good to me.

PR #292, opened 2026-08-05
CrossGapAlgorithm update for angle dependency crossing the gap

WemarkW:
Changes made to help muon tracks get reconstructed across the APA gap when propagating at angles close to parallel with the gap.

### DIFF
--- a/larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
+++ b/larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
@@ -13,6 +13,10 @@
 
 #include "larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.h"
 
+#include "larpandoracontent/LArObjects/LArCaloHit.h"
+
+#include <cmath>
+
 using namespace pandora;
 
 namespace lar_content
@@ -28,7 +32,11 @@ CrossGapsAssociationAlgorithm::CrossGapsAssociationAlgorithm() :
     m_maxOnClusterDistance(1.5f),
     m_minMatchedSamplingPoints(10),
     m_minMatchedSamplingFraction(0.5f),
-    m_gapTolerance(0.f)
+    m_crossAPAStepModifier(6.5f),
+    m_crossAPANearClusterModifier(1.0f),
+    m_boost_start_step(12.0f),
+    m_gapTolerance(0.f),
+    m_visualize(false)
 {
 }
 
@@ -43,9 +51,11 @@ void CrossGapsAssociationAlgorithm::GetListOfCleanClusters(const ClusterList *co
     for (const Cluster *const pCluster : *pClusterList)
     {
         if (pCluster->GetNCaloHits() < m_minClusterHits)
+            
             continue;
 
         if (1 + pCluster->GetOuterPseudoLayer() - pCluster->GetInnerPseudoLayer() < m_minClusterLayers)
+            
             continue;
 
         clusterVector.push_back(pCluster);
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
 
     if (outerFitResult.GetCluster()->GetInnerPseudoLayer() < innerFitResult.GetCluster()->GetOuterPseudoLayer())
         return false;
+    
+    
+    const Cluster *const pCluster{innerFitResult.GetCluster()};
+    CaloHitList caloHits;
+    LArClusterHelper::GetAllHits(pCluster, caloHits);
+    unsigned int innervolID = 2;
+    std::set<int> innervolIDs;
+    for (const CaloHit *const pCaloHit : caloHits)
+    {
+        const LArCaloHit *const pLArCaloHit{dynamic_cast<const LArCaloHit *>(pCaloHit)};
+        
+        innervolID = pLArCaloHit->GetLArTPCVolumeId();
+        innervolIDs.insert(innervolID);
+    }
+    caloHits.clear();
+    
+    const Cluster *const pCluster2{outerFitResult.GetCluster()};
+    CaloHitList caloHits2;
+    LArClusterHelper::GetAllHits(pCluster2, caloHits2);
+    unsigned int outervolID = 2;
+    std::set<int> outervolIDs;
+    for (const CaloHit *const pCaloHit2 : caloHits2)
+    {
+        const LArCaloHit *const pLArCaloHit2{dynamic_cast<const LArCaloHit *>(pCaloHit2)};
+        
+        outervolID = pLArCaloHit2->GetLArTPCVolumeId();
+        outervolIDs.insert(outervolID);
+    }
+    caloHits2.clear();
+    
+    bool isCrossingAPA = false;
+    
+    
+    std::cout << "InnervolID = " << innervolID << " / OutervolID = " << outervolID << std::endl;
+    
+    std::vector<int> sharedClusterVolIDs;
+    
+    std::set_intersection(innervolIDs.begin(), innervolIDs.end(),
+                          outervolIDs.begin(), outervolIDs.end(),
+                          std::back_inserter(sharedClusterVolIDs));
+    
+    if (sharedClusterVolIDs.empty()) {
+    
+        isCrossingAPA = true; //set this to false to turn off the corrections for crossing the apa gap
+        
+    }
+    
+    
+    innervolIDs.clear();
+    outervolIDs.clear();
 
-    return (this->IsAssociated(innerFitResult.GetGlobalMaxLayerPosition(), innerFitResult.GetGlobalMaxLayerDirection(), outerFitResult) &&
-        this->IsAssociated(outerFitResult.GetGlobalMinLayerPosition(), outerFitResult.GetGlobalMinLayerDirection() * -1.f, innerFitResult));
+    return (this->IsAssociated(innerFitResult.GetGlobalMaxLayerPosition(), innerFitResult.GetGlobalMaxLayerDirection(), outerFitResult, isCrossingAPA) &&
+        this->IsAssociated(outerFitResult.GetGlobalMinLayerPosition(), outerFitResult.GetGlobalMinLayerDirection() * -1.f, innerFitResult, isCrossingAPA));
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 bool CrossGapsAssociationAlgorithm::IsAssociated(
-    const CartesianVector &startPosition, const CartesianVector &startDirection, const TwoDSlidingFitResult &targetFitResult) const
+    const CartesianVector &startPosition, const CartesianVector &startDirection, const TwoDSlidingFitResult &targetFitResult, bool &isCrossingAPA) const
 {
     const HitType hitType(LArClusterHelper::GetClusterHitType(targetFitResult.GetCluster()));
     const float ratio{LArGeometryHelper::GetWirePitchRatio(this->GetPandora(), hitType)};
     const float sampleStepSizeAdjusted{ratio * m_sampleStepSize};
     unsigned int nMatchedSamplingPoints(0), nUnmatchedSampleRun(0);
 
+    bool shouldVisualise{false};
+    if (m_visualize)
+    {
+        for (unsigned int iSample = 0; iSample < m_maxSamplingPoints; ++iSample)
+        {
+            const CartesianVector samplingPoint(startPosition + startDirection * static_cast<float>(iSample) * sampleStepSizeAdjusted);
+
+            if (LArGeometryHelper::IsInGap(this->GetPandora(), samplingPoint, hitType, m_gapTolerance))
+            {
+                shouldVisualise = true;
+                break;
+            }
+        }
+    }
+    
+    unsigned int CrossAPAGapAdditionalSteps = 0;
+    
+    if ( startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
+        
+    if ( startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(-1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
+    std::cout << "CrossAPAGapAdditionalSteps = " << CrossAPAGapAdditionalSteps << std::endl;
+    std::cout << "Angle1 = " << startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) << std::endl;
+    std::cout << "Angle2 = " << startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) << std::endl;
+    float num_apa_steps = 0.0;
     for (unsigned int iSample = 0; iSample < m_maxSamplingPoints; ++iSample)
     {
+    
         const CartesianVector samplingPoint(startPosition + startDirection * static_cast<float>(iSample) * sampleStepSizeAdjusted);
 
         if (LArGeometryHelper::IsInGap(this->GetPandora(), samplingPoint, hitType, m_gapTolerance))
         {
+            if (shouldVisualise)
+            {
+                PANDORA_MONITORING_API(AddMarkerToVisualization(this->GetPandora(), &samplingPoint, "", BLUE, 1));
+            }
+            num_apa_steps = num_apa_steps + 1.0;
             nUnmatchedSampleRun = 0; // ATTN Choose to also reset run when entering gap region
             continue;
         }
 
-        if (this->IsNearCluster(samplingPoint, targetFitResult))
+        if (this->IsNearCluster(samplingPoint, targetFitResult, num_apa_steps, isCrossingAPA))
         {
             ++nMatchedSamplingPoints;
             nUnmatchedSampleRun = 0;
+            if (shouldVisualise)
+            {
+                PANDORA_MONITORING_API(AddMarkerToVisualization(this->GetPandora(), &samplingPoint, "", GREEN, 1));
+            }
         }
-        else if (++nUnmatchedSampleRun > m_maxUnmatchedSampleRun)
+        else if (++nUnmatchedSampleRun > m_maxUnmatchedSampleRun + CrossAPAGapAdditionalSteps)
         {
             break;
         }
+        else if (shouldVisualise)
+        {
+            PANDORA_MONITORING_API(AddMarkerToVisualization(this->GetPandora(), &samplingPoint, "", RED, 1));
+        }
     }
-
+    std::cout << "Number of steps in APA gap = " << num_apa_steps << std::endl;
+    if (shouldVisualise)
+    {
+        PANDORA_MONITORING_API(ViewEvent(this->GetPandora()));
+    }
+        
     const float expectation(
         (targetFitResult.GetGlobalMaxLayerPosition() - targetFitResult.GetGlobalMinLayerPosition()).GetMagnitude() / sampleStepSizeAdjusted);
     const float matchedSamplingFraction(expectation > 0.f ? static_cast<float>(nMatchedSamplingPoints) / expectation : 0.f);
@@ -176,7 +283,7 @@ bool CrossGapsAssociationAlgorithm::IsAssociated(
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplingPoint, const TwoDSlidingFitResult &targetFitResult) const
+bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplingPoint, const TwoDSlidingFitResult &targetFitResult, float &num_apa_steps, bool &isCrossingAPA) const
 {
     const HitType hitType(LArClusterHelper::GetClusterHitType(targetFitResult.GetCluster()));
     const float ratio{LArGeometryHelper::GetWirePitchRatio(this->GetPandora(), hitType)};
@@ -186,10 +293,13 @@ bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplin
     targetFitResult.GetLocalPosition(samplingPoint, rL, rT);
 
     CartesianVector fitPosition(0.f, 0.f, 0.f);
+    
+    float crossingAPAcheck = static_cast<float>(isCrossingAPA);
+    
 
     if (STATUS_CODE_SUCCESS == targetFitResult.GetGlobalFitPosition(rL, fitPosition))
     {
-        if ((fitPosition - samplingPoint).GetMagnitudeSquared() < maxOnClusterDistanceAdjusted * maxOnClusterDistanceAdjusted)
+        if ((fitPosition - samplingPoint).GetMagnitudeSquared() < (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck) * (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck))
             return true;
     }
 
@@ -197,7 +307,7 @@ bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplin
 
     if (STATUS_CODE_SUCCESS == targetFitResult.GetGlobalFitPositionAtX(samplingPoint.GetX(), fitPositionAtX))
     {
-        if ((fitPositionAtX - samplingPoint).GetMagnitudeSquared() < maxOnClusterDistanceAdjusted * maxOnClusterDistanceAdjusted)
+        if ((fitPositionAtX - samplingPoint).GetMagnitudeSquared() < (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck) * (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck))
             return true;
     }
 
@@ -238,8 +348,19 @@ StatusCode CrossGapsAssociationAlgorithm::ReadSettings(const TiXmlHandle xmlHand
 
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=,
         XmlHelper::ReadValue(xmlHandle, "MinMatchedSamplingFraction", m_minMatchedSamplingFraction));
+        
+    PANDORA_RETURN_RESULT_IF_AND_IF(
+        STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "CrossAPAStepModifier", m_crossAPAStepModifier));
+        
+    PANDORA_RETURN_RESULT_IF_AND_IF(
+        STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "CrossAPANearClusterModifier", m_crossAPANearClusterModifier));
+        
+    PANDORA_RETURN_RESULT_IF_AND_IF(
+        STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "BoostStartStep", m_boost_start_step));
 
     PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "GapTolerance", m_gapTolerance));
+    
+    PANDORA_RETURN_RESULT_IF_AND_IF(STATUS_CODE_SUCCESS, STATUS_CODE_NOT_FOUND, !=, XmlHelper::ReadValue(xmlHandle, "Visualize", m_visualize));
 
     return ClusterAssociationAlgorithm::ReadSettings(xmlHandle);
 }
--- a/larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
+++ b/larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
@@ -32,9 +32,10 @@ CrossGapsAssociationAlgorithm::CrossGapsAssociationAlgorithm() :
     m_maxOnClusterDistance(1.5f),
     m_minMatchedSamplingPoints(10),
     m_minMatchedSamplingFraction(0.5f),
-    m_crossAPAStepModifier(6.5f),
-    m_crossAPANearClusterModifier(1.0f),
-    m_boost_start_step(12.0f),
+    m_crossTPCStepModifier(0.0f),
+    m_crossTPCOnClusterDistanceModifier(0.0f),
+    m_boostStartStep(0.0f),
+    m_crossTPCMaxAdditionalSteps(1000.0f),
     m_gapTolerance(0.f),
     m_visualize(false)
 {
@@ -51,11 +52,9 @@ void CrossGapsAssociationAlgorithm::GetListOfCleanClusters(const ClusterList *co
     for (const Cluster *const pCluster : *pClusterList)
     {
         if (pCluster->GetNCaloHits() < m_minClusterHits)
-            
             continue;
 
-        if (1 + pCluster->GetOuterPseudoLayer() - pCluster->GetInnerPseudoLayer() < m_minClusterLayers)
-            
+        if (1 + pCluster->GetOuterPseudoLayer() - pCluster->GetInnerPseudoLayer() < m_minClusterLayers)   
             continue;
 
         clusterVector.push_back(pCluster);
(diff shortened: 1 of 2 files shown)


larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -28,7 +32,11 @@ CrossGapsAssociationAlgorithm::CrossGapsAssociationAlgorithm() :
     m_maxOnClusterDistance(1.5f),
     m_minMatchedSamplingPoints(10),
     m_minMatchedSamplingFraction(0.5f),
-    m_gapTolerance(0.f)
+    m_crossAPAStepModifier(6.5f),
```
AlexWilkinsonnn, line 35: The drift gap being an APA is not true for all geometries (full 10kt HD for example). I think "crossTPC" would be more appropriate for these variables

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -28,7 +32,11 @@ CrossGapsAssociationAlgorithm::CrossGapsAssociationAlgorithm() :
...
+    m_crossAPANearClusterModifier(1.0f),
```
AlexWilkinsonnn, line 36: NearCluster ->OnClusterDistance so it is clear what threshold this modifies

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -28,7 +32,11 @@ CrossGapsAssociationAlgorithm::CrossGapsAssociationAlgorithm() :
...
+    m_boost_start_step(12.0f),
```
AlexWilkinsonnn, line 37: use camel case for consistency. Also I think using "sample" instead of "step" is more consistent with the other variable names? That would be clearer to me but its subjective

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -43,9 +51,11 @@ void CrossGapsAssociationAlgorithm::GetListOfCleanClusters(const ClusterList *co
     for (const Cluster *const pCluster : *pClusterList)
     {
         if (pCluster->GetNCaloHits() < m_minClusterHits)
+
```
AlexWilkinsonnn, line 54: remove

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -43,9 +51,11 @@ void CrossGapsAssociationAlgorithm::GetListOfCleanClusters(const ClusterList *co
     for (const Cluster *const pCluster : *pClusterList)
     {
         if (pCluster->GetNCaloHits() < m_minClusterHits)
+            
             continue;
 
         if (1 + pCluster->GetOuterPseudoLayer() - pCluster->GetInnerPseudoLayer() < m_minClusterLayers)
+
```
AlexWilkinsonnn, line 58: remove

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+        
+        outervolID = pLArCaloHit2->GetLArTPCVolumeId();
+        outervolIDs.insert(outervolID);
+    }
+    caloHits2.clear();
+    
+    bool isCrossingAPA = false;
+    
+    
+    std::cout << "InnervolID = " << innervolID << " / OutervolID = " << outervolID << std::endl;
```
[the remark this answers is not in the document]
AlexWilkinsonnn: This cout should just be deleted

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
AlexWilkinsonnn, line 37: The algorithm default values should be set to do nothing, the new crossing gap logic should be opt-in from the xml on the experiment side

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    
+    std::set_intersection(innervolIDs.begin(), innervolIDs.end(),
+                          outervolIDs.begin(), outervolIDs.end(),
+                          std::back_inserter(sharedClusterVolIDs));
+    
+    if (sharedClusterVolIDs.empty()) {
+    
+        isCrossingAPA = true; //set this to false to turn off the corrections for crossing the apa gap
+        
+    }
```
AlexWilkinsonnn, line 186: We only need to check if the clusters are from distinct TPC volumes if we have the new cross gap logic activated. So only do this calculation if the values of the new member variables mean we might apply the angle dependency and cluster distance boost.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    for (const CaloHit *const pCaloHit2 : caloHits2)
+    {
+        const LArCaloHit *const pLArCaloHit2{dynamic_cast<const LArCaloHit *>(pCaloHit2)};
+        
+        outervolID = pLArCaloHit2->GetLArTPCVolumeId();
+        outervolIDs.insert(outervolID);
+    }
+    caloHits2.clear();
+    
+    bool isCrossingAPA = false;
```
AlexWilkinsonnn, line 171: I think `isCrossingTPCCandidate` is clearer, we don't know yet if this pair is actually from a TPC crossing particle

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    CaloHitList caloHits;
+    LArClusterHelper::GetAllHits(pCluster, caloHits);
+    unsigned int innervolID = 2;
+    std::set<int> innervolIDs;
+    for (const CaloHit *const pCaloHit : caloHits)
+    {
+        const LArCaloHit *const pLArCaloHit{dynamic_cast<const LArCaloHit *>(pCaloHit)};
+        
+        innervolID = pLArCaloHit->GetLArTPCVolumeId();
+        innervolIDs.insert(innervolID);
```
AlexWilkinsonnn, line 153: Insert directly into the set

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
AlexWilkinsonnn, line 186: remove emtpy lines and curly braces

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    
+    if (sharedClusterVolIDs.empty()) {
+    
+        isCrossingAPA = true; //set this to false to turn off the corrections for crossing the apa gap
+        
+    }
+    
+    
+    innervolIDs.clear();
+    outervolIDs.clear();
```
AlexWilkinsonnn, line 190: dont need to clear

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    LArClusterHelper::GetAllHits(pCluster, caloHits);
+    unsigned int innervolID = 2;
+    std::set<int> innervolIDs;
+    for (const CaloHit *const pCaloHit : caloHits)
+    {
+        const LArCaloHit *const pLArCaloHit{dynamic_cast<const LArCaloHit *>(pCaloHit)};
+        
+        innervolID = pLArCaloHit->GetLArTPCVolumeId();
+        innervolIDs.insert(innervolID);
+    }
```
AlexWilkinsonnn, line 154: Could you put this in a member function that can be called for the inner and outer fit resullt to return the volume ID set? something like
```
StatusCode FindLArTPCVolumeIds(const TwoDSlidingFitResult &fitResult, std::set<unsigned int> &volIds)
```
could then be called as
```
std::set<unsigned int> innerVolIDs;
PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->FindLArTPCVolumeIds(innerFitResult, innerVolIDs);
std::set<unsigned int> outerVolIDs;
PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->FindLArTPCVolumeIds(outerFitResult, innerVolIDs);
```
then you return STATUS_CODE_NOT_INITIALIZED if pLArCaloHit is null and return STATUS_CODE_SUCCESS at the end.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
-        this->IsAssociated(outerFitResult.GetGlobalMinLayerPosition(), outerFitResult.GetGlobalMinLayerDirection() * -1.f, innerFitResult));
+    return (this->IsAssociated(innerFitResult.GetGlobalMaxLayerPosition(), innerFitResult.GetGlobalMaxLayerDirection(), outerFitResult, isCrossingAPA) &&
+        this->IsAssociated(outerFitResult.GetGlobalMinLayerPosition(), outerFitResult.GetGlobalMinLayerDirection() * -1.f, innerFitResult, isCrossingAPA));
 }
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
 bool CrossGapsAssociationAlgorithm::IsAssociated(
-    const CartesianVector &startPosition, const CartesianVector &startDirection, const TwoDSlidingFitResult &targetFitResult) const
+    const CartesianVector &startPosition, const CartesianVector &startDirection, const TwoDSlidingFitResult &targetFitResult, bool &isCrossingAPA) const
```
AlexWilkinsonnn, line 199: `bool &` -> `const bool`. it is better to pass small types (bool, int, float, ...) by const value when you dont need them to be changed in the function

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
 {
     const HitType hitType(LArClusterHelper::GetClusterHitType(targetFitResult.GetCluster()));
     const float ratio{LArGeometryHelper::GetWirePitchRatio(this->GetPandora(), hitType)};
     const float sampleStepSizeAdjusted{ratio * m_sampleStepSize};
     unsigned int nMatchedSamplingPoints(0), nUnmatchedSampleRun(0);
 
+    bool shouldVisualise{false};
```
AlexWilkinsonnn, line 206: `Visualise` -> `Visualize` for consistency (that was me sorry :))

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+        }
+    }
+    
+    unsigned int CrossAPAGapAdditionalSteps = 0;
+    
+    if ( startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
+        
+    if ( startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) ) {
```
AlexWilkinsonnn, line 227: Is this not the same as `else` ?

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+                shouldVisualise = true;
+                break;
+            }
+        }
+    }
+    
+    unsigned int CrossAPAGapAdditionalSteps = 0;
+    
+    if ( startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
```
AlexWilkinsonnn, line 224: better to just only calculate this if `isCrossingAPA` is true rather than use it as a 0 in the product

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+
+            if (LArGeometryHelper::IsInGap(this->GetPandora(), samplingPoint, hitType, m_gapTolerance))
+            {
+                shouldVisualise = true;
+                break;
+            }
+        }
+    }
+    
+    unsigned int CrossAPAGapAdditionalSteps = 0;
```
AlexWilkinsonnn, line 221: camelcase, APA -> TPC, Steps -> SamplingPoints

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    if ( startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
+        
+    if ( startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(-1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
+    std::cout << "CrossAPAGapAdditionalSteps = " << CrossAPAGapAdditionalSteps << std::endl;
+    std::cout << "Angle1 = " << startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) << std::endl;
+    std::cout << "Angle2 = " << startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) << std::endl;
```
AlexWilkinsonnn, line 232: remove

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    float num_apa_steps = 0.0;
```
AlexWilkinsonnn, line 233: camel case and I think i makes more sense to use an int, increment it with `++` then you can static cast it to float when you need to later

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
     for (unsigned int iSample = 0; iSample < m_maxSamplingPoints; ++iSample)
     {
+
```
AlexWilkinsonnn, line 236: remove

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    
+    unsigned int CrossAPAGapAdditionalSteps = 0;
+    
+    if ( startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
+        
+    if ( startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) ) {
+        CrossAPAGapAdditionalSteps = static_cast<int>(std::round( m_crossAPAStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(-1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingAPA);
+    }
```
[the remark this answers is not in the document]
AlexWilkinsonnn: I agree that we need some way to ensure this isnt a huge factor, perhaps another config like "m_crossTPCMaxAdditionalSteps" could be used to clamp this?

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +138,139 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
         {
             break;
         }
+        else if (shouldVisualise)
+        {
+            PANDORA_MONITORING_API(AddMarkerToVisualization(this->GetPandora(), &samplingPoint, "", RED, 1));
+        }
     }
-
+    std::cout << "Number of steps in APA gap = " << num_apa_steps << std::endl;
```
AlexWilkinsonnn, line 268: remove or move into visualise block

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -176,7 +283,7 @@ bool CrossGapsAssociationAlgorithm::IsAssociated(
 
 //------------------------------------------------------------------------------------------------------------------------------------------
 
-bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplingPoint, const TwoDSlidingFitResult &targetFitResult) const
+bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplingPoint, const TwoDSlidingFitResult &targetFitResult, float &num_apa_steps, bool &isCrossingAPA) const
```
AlexWilkinsonnn, line 286: const and pass by value the new arguments

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -186,18 +293,21 @@ bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplin
...
+        if ((fitPosition - samplingPoint).GetMagnitudeSquared() < (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck) * (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck))
             return true;
     }
 
     CartesianVector fitPositionAtX(0.f, 0.f, 0.f);
 
     if (STATUS_CODE_SUCCESS == targetFitResult.GetGlobalFitPositionAtX(samplingPoint.GetX(), fitPositionAtX))
     {
-        if ((fitPositionAtX - samplingPoint).GetMagnitudeSquared() < maxOnClusterDistanceAdjusted * maxOnClusterDistanceAdjusted)
+        if ((fitPositionAtX - samplingPoint).GetMagnitudeSquared() < (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck) * (maxOnClusterDistanceAdjusted + m_crossAPANearClusterModifier * sqrt(std::max(num_apa_steps - m_boost_start_step, 0.f)) * crossingAPAcheck))
```
AlexWilkinsonnn, line 310: The if statement is too busy here. I suggest using the crossingTPC flag as a flag rather than converting it to 0 or 1, and then calculating the extra factors before the if statement.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -186,18 +293,21 @@ bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplin
...
             return true;
     }
```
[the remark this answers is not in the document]
AlexWilkinsonnn: Agreed, I think this part needs breaking down into components, more lines is good if it makes it clearer

AlexWilkinsonnn (changes requested):
Thank you! The underlying logic looks good, and we have seen it is helpful. There are few aspects of the style & structure that need to be changed before merging. We can discuss them in the review or one Slack if you're not sure about anything.

Thanks - Alex

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -28,7 +32,12 @@ CrossGapsAssociationAlgorithm::CrossGapsAssociationAlgorithm() :
     m_maxOnClusterDistance(1.5f),
     m_minMatchedSamplingPoints(10),
     m_minMatchedSamplingFraction(0.5f),
-    m_gapTolerance(0.f)
+    m_crossTPCStepModifier(0.0f),
+    m_crossTPCOnClusterDistanceModifier(0.0f),
+    m_boostStartStep(0.0f),
```
AlexWilkinsonnn: maybe this should be `m_crossTPCBoostStartStep` so it's clear these 4 parameters are all controlling the same feature?

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +137,109 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
 
     if (outerFitResult.GetCluster()->GetInnerPseudoLayer() < innerFitResult.GetCluster()->GetOuterPseudoLayer())
         return false;
+        
+    bool isCrossingTPCCandidate = false;
+    if (m_crossTPCStepModifier != 0.0) {
```
AlexWilkinsonnn: use `0.f` on the rhs (float v double)

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.h
```diff
@@ -63,7 +63,9 @@ class CrossGapsAssociationAlgorithm : public ClusterAssociationAlgorithm
      *
      *  @return boolean
      */
-    bool IsNearCluster(const pandora::CartesianVector &samplingPoint, const TwoDSlidingFitResult &targetFitResult) const;
+    bool IsNearCluster(const pandora::CartesianVector &samplingPoint, const TwoDSlidingFitResult &targetFitResult, const float numGapSteps, const bool isCrossingTPCCandidate) const;
+    
+    pandora::StatusCode FindLArTPCVolumeIds(const TwoDSlidingFitResult &fitResult, std::set<unsigned int> &volIds) const;
```
AlexWilkinsonnn: add doxygen comment above

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +137,109 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+        PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->FindLArTPCVolumeIds(outerFitResult, outerVolIDs));
+
+        std::vector<int> sharedClusterVolIDs;
+    
+        std::set_intersection(innerVolIDs.begin(), innerVolIDs.end(),
+                          outerVolIDs.begin(), outerVolIDs.end(),
+                          std::back_inserter(sharedClusterVolIDs));
+    
+        if (sharedClusterVolIDs.empty()) 
+            isCrossingTPCCandidate = true; //set this to false to turn off the corrections for crossing the gap
```
AlexWilkinsonnn: remove comment

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +137,109 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+                break;
+            }
+        }
+    }
+    
+    unsigned int crossTPCGapAdditionalSteps = 0;
+    if(isCrossingTPCCandidate) {
+        if ( startDirection.GetOpeningAngle(CartesianVector(1.0, 0.0, 0.0)) < startDirection.GetOpeningAngle(CartesianVector(-1.0, 0.0, 0.0)) ) {
+            crossTPCGapAdditionalSteps = static_cast<int>(std::round( m_crossTPCStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(1.0, 0.0, 0.0))))  *      static_cast<int>(isCrossingTPCCandidate);
+        }
```
AlexWilkinsonnn: I think the intent here can be made clearer. The condition is just checking the direction, that is clearer if you do `if (startDirection.GetX() > 0.f)` instead of the original condition.

Also, you no longer need the multiplication of `isCrossingTPCCandidate`.

Also, the cast should be to unsigned int.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +137,109 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+        
+        else {
+            crossTPCGapAdditionalSteps = static_cast<int>(std::round( m_crossTPCStepModifier / startDirection.GetCosOpeningAngle(CartesianVector(-1.0, 0.0, 0.0))))  * static_cast<int>(isCrossingTPCCandidate);
+        }
+        if ( crossTPCGapAdditionalSteps > m_crossTPCMaxAdditionalSteps ) {
+            crossTPCGapAdditionalSteps = m_crossTPCMaxAdditionalSteps;
```
AlexWilkinsonnn: The capping of the extra steps needs to happen in the denominator to avoid a potential division by zero. I think it is fine to hardcode this limit to, say, `std::max(startDirection.GetCosOpeningAngle(CartesianVector(1.f, 0.f, 0.f)), 0.001f)`. This would mean you can get rid of the `m_crossTPCMaxAdditionalSteps` configurable.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +137,109 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+    
         const CartesianVector samplingPoint(startPosition + startDirection * static_cast<float>(iSample) * sampleStepSizeAdjusted);
 
         if (LArGeometryHelper::IsInGap(this->GetPandora(), samplingPoint, hitType, m_gapTolerance))
         {
+            if (shouldVisualize)
+            {
+                PANDORA_MONITORING_API(AddMarkerToVisualization(this->GetPandora(), &samplingPoint, "", BLUE, 1));
+            }
+            numGapSteps = numGapSteps + 1.0;
```
AlexWilkinsonnn: `numGapSteps` should be be declared as an int if we are going to increment it in this loop, we can cast it to a float later to use as a scaling factor

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
AlexWilkinsonnn: You might want to use the CrossTPCOnClusterDistanceModifier and not the step modifier, so this condition should check if both of these are turned off.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +137,109 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+            if (LArGeometryHelper::IsInGap(this->GetPandora(), samplingPoint, hitType, m_gapTolerance))
+            {
+                shouldVisualize = true;
+                break;
+            }
+        }
+    }
+    
+    unsigned int crossTPCGapAdditionalSteps = 0;
+    if(isCrossingTPCCandidate) {
```
AlexWilkinsonnn: Also check if `m_crossTPCStepModifier` is not zero

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -186,18 +262,23 @@ bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplin
     targetFitResult.GetLocalPosition(samplingPoint, rL, rT);
 
     CartesianVector fitPosition(0.f, 0.f, 0.f);
+    
+    float additionalMaxOnClusterDistance = 0.0;
+    if (isCrossingTPCCandidate) {
+        additionalMaxOnClusterDistance = m_crossTPCOnClusterDistanceModifier * sqrt(std::max(numGapSteps - m_boostStartStep, 0.f));
+    }
```
AlexWilkinsonnn: Use a ternary `isCrossingTPCCandidate ? ... : 0.f` and then add this to `maxOnClusterDistanceAdjusted` directly

AlexWilkinsonnn (changes requested):
It looks a lot better now! I left a few more comments to neaten it up. Also, in general, try to stick to the curly brackets style of 
```
if (cond)
{
    ...
}
```

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +136,106 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+                shouldVisualize = true;
+                break;
+            }
+        }
+    }
+    
+    unsigned int crossTPCGapAdditionalSteps = 0;
+    if(isCrossingTPCCandidate and m_crossTPCStepModifier != 0.0f) {
+        if (startDirection.GetX() > 0.f) {
+            crossTPCGapAdditionalSteps = static_cast<unsigned int>(std::round( m_crossTPCStepModifier / std::max(startDirection.GetCosOpeningAngle(CartesianVector(1.f, 0.f, 0.f)), 0.001f)));
```
AlexWilkinsonnn: This looks good, could you leave a comment just to note what this hardcoded number does? Maybe `// ATTN hard limit on factor from angle, prevents very large number of steps from angles close to 90 degrees`

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +136,106 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
 
     if (outerFitResult.GetCluster()->GetInnerPseudoLayer() < innerFitResult.GetCluster()->GetOuterPseudoLayer())
         return false;
+        
+    bool isCrossingTPCCandidate = false;
+    if (m_crossTPCOnClusterDistanceModifier != 0.0f) {
```
AlexWilkinsonnn: I think this should also have `... || m_crossTPCStepModifier != 0.f)`. Otherwise  you cant use the number of steps boosting functionality without also using the cluster distance increasing functionality..

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +136,106 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+            if (LArGeometryHelper::IsInGap(this->GetPandora(), samplingPoint, hitType, m_gapTolerance))
+            {
+                shouldVisualize = true;
+                break;
+            }
+        }
+    }
+    
+    unsigned int crossTPCGapAdditionalSteps = 0;
+    if(isCrossingTPCCandidate and m_crossTPCStepModifier != 0.0f) {
```
AlexWilkinsonnn: `and` -> ` && `

AlexWilkinsonnn (changes requested):
These are my last comments. the L141 comment is the most important.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -128,42 +136,106 @@ bool CrossGapsAssociationAlgorithm::AreClustersAssociated(const TwoDSlidingFitRe
...
+        
+    bool isCrossingTPCCandidate = false;
+    if (m_crossTPCOnClusterDistanceModifier != 0.0f) {
+        
+        std::set<unsigned int> innerVolIDs;
+        PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->FindLArTPCVolumeIds(innerFitResult, innerVolIDs));
+        std::set<unsigned int> outerVolIDs;
+        PANDORA_THROW_RESULT_IF(STATUS_CODE_SUCCESS, !=, this->FindLArTPCVolumeIds(outerFitResult, outerVolIDs));
+
+        std::vector<int> sharedClusterVolIDs;
```
AndyChappell: make this an unsigned int vector for type consistency with the input sets

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
AndyChappell: It would also be useful to note how that number was chosen - my initial reaction is that at ~90 degrees, this turns into a very large number of steps (which may be completely fine, but naively, a factor of 1000 seems like a lot).

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
```diff
@@ -186,18 +258,20 @@ bool CrossGapsAssociationAlgorithm::IsNearCluster(const CartesianVector &samplin
     targetFitResult.GetLocalPosition(samplingPoint, rL, rT);
 
     CartesianVector fitPosition(0.f, 0.f, 0.f);
-
+    
+    float additionalMaxOnClusterDistance = isCrossingTPCCandidate ? m_crossTPCOnClusterDistanceModifier * sqrt(std::max(static_cast<float>(numGapSteps) - m_crossTPCBoostStartStep, 0.f)) : 0.f;
```
AndyChappell: std::sqrt

AndyChappell (changes requested):
Thanks @WemarkW . A couple of minor changes requested and one clarification in addition to Alex's comments, but otherwise this looks close to ready.

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
AlexWilkinsonnn: I suggested the sqrt just to have something that scales sub-linear because that feels right. I think there is some handwavey argument that the sqrt is a reasonable paramtrisation of the transverse uncertainty of a track that can undergo small scatters

AndyChappell: ok, thanks Alex - not a geometric adjustment then

larpandoracontent/LArTwoDReco/LArClusterAssociation/CrossGapsAssociationAlgorithm.cc
WemarkW: Alex proposed this change to me with this number. However, when using the gap width and the length of the full geometry detector, I get a cos of appx .001. I will agree though, it is definitely a lot

AndyChappell: Ok, knowing that 1e-3 puts it on the scale of the detector (which geometry by the way?) helps. So, at this point I'll just ask the question (to @AlexWilkinsonnn as well), are we generally happy that, that's a reasonable hard limit? I think I'm fairly convinced that we wouldn't want to try to span huge gaps, even in a full DUNE 10kt geometry (but does that motivate making this a configurable value?), and I imagine there are some fairly large true gaps that we could reasonably expect to span - how does this value compare to the largest true gap we've seen?

WemarkW: I get something on the order of 0.001 when using the full 5800 cm length of a 10kt module of the detector. Which I agree is a huge gap to span over. The geometry that I have been working with is the 1x2x6 and the largest gap I found was 249 steps in the APA gap. This is about a 125 cm gap, and the cos I get in this case is about 0.048, which is a lot higher than 0.001. I could change the limit to 0.01 if that makes more sense

